PR #7, opened 2022-04-08
Basic client server #3

gijsvl:


src/keymgmt/server.rs
```diff
@@ -0,0 +1,132 @@
...
+
+                    // For each request, dispatch to the appropriate method, defined elsewhere
+                    let interact = move |session_key, (), chan: Chan<KeyMgmt>| {
+                        // Clone `Arc`s for the various resources we need in this request
+                        let client = client.clone();
+                        let service = service.clone();
+                        let config = config.clone();
+
+                        // TODO: permit configuration option to make this deterministic for testing
+                        let rng = StdRng::from_entropy();
```
marsella, line 75: Issue: I think this will create a new RNG every time the server gets a request. Is there a way to declare this outside of this function and pass a mutable reference or something instead? I am not sure how the ownership rules will work since this is a `move` block, but I am a bit concerned about repeated instantiation here.

src/lib.rs
```diff
@@ -1,8 +1,39 @@
 //! This crate is an implementation of a key management system
-
-#![warn(missing_docs)]
+// #![warn(missing_docs)]
```
marsella, line 2: todo: If we're not using this tag, can you delete it?

src/protocol.rs
```diff
@@ -0,0 +1,169 @@
...
+/// The two parties in the protocol.
+#[derive(Debug, Clone, Copy, Eq, Ord, PartialEq, PartialOrd, Hash, Serialize, Deserialize)]
+pub enum Party {
+    /// The client.
+    Client,
+    /// The server.
+    Server,
+}
+
+impl Display for Party {
```
marsella, line 77: suggestion: Since the display names are the same as the enum variants, you can probably use this shorthand: https://stackoverflow.com/questions/32710187/how-do-i-get-an-enum-as-a-string#32712140

src/protocol.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// Get the other party.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use keymgmt::protocol::Party::*;
+    ///
+    /// assert_eq!(Client.opposite(), Server);
+    /// assert_eq!(Server.opposite(), Client);
+    /// ```
```
marsella, line 101: Nice doctest!

src/protocol.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub type Retrieve = RetrieveSecret;
+
+    /// The internals of the Retrieve sessionType
+    pub type RetrieveSecret = Session! {
+        send SecretRetrieveRequest;
+        recv SecretInfo;
+    };
+}
+
+pub mod daemon {
```
marsella, line 159: Issue: I don't think we need the daemon protocol yet. Let's remove it until it's necessary.

src/config/client.rs
```diff
@@ -0,0 +1,49 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+#[non_exhaustive]
+pub struct Config {
+    #[serde(default = "defaults::backoff")]
+    pub backoff: Backoff,
+    #[serde(with = "humantime_serde", default = "defaults::connection_timeout")]
+    pub connection_timeout: Option<Duration>,
+    #[serde(default = "defaults::daemon_port")]
+    pub daemon_port: u16,
```
marsella, line 19: todo: Please remove daemon, I don't think we need it yet.

src/defaults.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    pub const fn max_note_length() -> u64 {
+        1024 * 8
+    }
+
+    pub const fn daemon_port() -> u16 {
+        // KMD :3
+        11134
+    }
```
marsella, line 81: todo: Please remove daemon, I don't think we need it yet. Even though it's a cute port #.

src/cli/server.rs
```diff
@@ -0,0 +1,31 @@
...
+    pub config: Option<PathBuf>,
+
+    /// Run server commands.
+    #[structopt(subcommand)]
+    pub server: Server,
+}
+
+#[derive(Debug, StructOpt)]
+pub enum Server {
+    Placeholder(Placeholder),
```
marsella, line 19: question: Is this so you can have the exhaustive "not implemented" case in the tests? I wouldn't mind deleting that case for now, since you _have_ implemented the tests for everything that exists to be tested. If we add new options here and don't put them in the test, we'll get a compile error and can add a default (or more tests) at that point.

src/keymgmt/server.rs
```diff
@@ -0,0 +1,132 @@
...
+use create::Create;
+use retrieve::Retrieve;
+
+/// A single server-side command, parameterized by the currently loaded configuration.
+///
+/// All subcommands of [`cli::Server`](crate::server::cli::Server) should implement this.
+#[async_trait]
+pub trait Command {
+    /// Output type of the server
+    type Output;
```
marsella, line 23: suggestion: You could remove this type entirely for now and simplify the `run` function to return `Result<(), anyhow::Error>`. I'm not sure yet what other server operations we'll have. Since this will behave as more of an actual server than zkchannels, where the merchant server could also initiate protocol executions, we might not need any non-`()` return types.

tests/common.rs
```diff
@@ -0,0 +1,259 @@
...
+
+use keymgmt::{server::keymgmt::Command as _, timeout::WithTimeout};
+
+pub const CLIENT_CONFIG: &str = "tests/gen/TestClient.toml";
+pub const SERVER_CONFIG: &str = "tests/gen/TestServer.toml";
+pub const ERROR_FILENAME: &str = "tests/gen/errors.log";
+
+/// The default server services we will set up for tests (all run on localhost)
+#[derive(Debug, Clone, Copy, EnumIter)]
+enum ServerServices {
```
marsella, line 29: suggestion: I over-engineered this in zeekoe because I thought we might want to test multiple server services (e.g. localhosts, other IP addresses, who knows?!), but it turns out we only use the IPv4 version. If you want, you can probably remove all of the `ServerServices` stuff and just make a `const` to hold the server address.

marsella (changes requested):
Thanks for this PR! Nice to see a bunch of infrastructure in one place. I have some suggestions for simplifying this implementation, since it is currently less complicated than zeekoe and we're not sure exactly how it will add complications. I also have some questions about design decisions, all inline.

src/keymgmt/client/retrieve.rs
```diff
@@ -0,0 +1,41 @@
...
+use async_trait::async_trait;
+
+use crate::client::{
+    cli::Retrieve,
+    keymgmt::{connect, Config, SecretRetrieveRequest},
+};
+
+#[async_trait]
+impl Command for Retrieve {
+    type Output = ();
```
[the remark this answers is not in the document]
gijsvl: Good point, this was something I wanted to talk to you about. It breaks the e2e tests as they are written now. This is also an issue in zeekoe, but up until now we haven't had any command that also is an action in the e2e test. I was hoping Vidya would solve it in one of the tests she's writing for zeekoe 😅. Let's discuss.

src/bin/client.rs
```diff
@@ -1,11 +1,28 @@
...
+        result
+            .with_context(|| format!("Could not load client configuration from {:?}", config_path))
+    });
 
-    #[test]
-    fn test_client_runs() {
-        main()
+    match cli.client {
+        Create(_run) => Ok(()),
+        Retrieve(_run) => Ok(()),
```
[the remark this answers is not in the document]
gijsvl: Nope, good catch, totally forgot

Cargo.toml
```diff
@@ -9,10 +9,49 @@ description = "implementation of a secure key mangement system"
...
 [[bin]]
 name = "key-mgmt-client"
 path = "src/bin/client.rs"
 
 [[bin]]
 name = "key-mgmt-server"
 path = "src/bin/server.rs"
+
+[features]
+allow_explicit_certificate_trust = []
```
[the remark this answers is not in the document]
gijsvl: I think the only obvious misuse is using the `--all-features` I guess, but I'm not sure we have to optimize for that case. I think the feature already makes it very clear what someone is doing, and I believe we shouldn't worry too much about the edge case. It's gonna give us more headaches than actually preventing anything.

marsella (approved):
Looks great, thanks for fixing!

src/keymgmt/server.rs
marsella: Added #20

PR #36, opened 2022-06-22
Add remote client API

### LINKED ISSUE #35: Define remote client API
Write out external facing API for the remote client component. Write out documentation, focusing on requirements for callers and assumptions.

marsella:
This adds the remote client API and tests. Not much else to say here, it follows a similar formula to the local client API, but with updated documentation.

Closes #35


src/remote_client.rs
```diff
@@ -0,0 +1,101 @@
...
+    todo!()
+}
+
+/// Retrieve the public key info for all keys associated with the specified
+/// user from the key server.
+///
+/// The [`KeyId`] must correspond to a [`Passive`](crate::keys::Passive) or
+/// [`Delegated`](crate::keys::Delegated) key that
+/// have delegated signing authority to the key fiduciary authenticated in the
+/// session. The [`KeyId`] must correspond to a key owned by the [`UserId`].
```
indomitableSwan, line 59: On the one hand, I like keeping this restrictive. On the other hand, it is not clear to me that this will meet all of the Service Provider's legitimate use cases, e.g., isn't it likely that the remote A&A server would want to retrieve this information for all keys for a given user, for say, a web browser version of an asset dashboard? That said, I am OK with leaving this functionality out for now.

src/remote_client.rs
```diff
@@ -0,0 +1,101 @@
...
+//! application running on the servers of a Service Provider.
+
+use crate::{
+    keys::{KeyId, KeyInfo, UseRestriction, UserId},
+    transaction::{TransactionApprovalRequest, TransactionSignature},
+};
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum Error {}
```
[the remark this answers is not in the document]
indomitableSwan: Ignore this comment: we can handle sessions in a separate epic/issue.

indomitableSwan (commented):
Reviewed the last commit, which adds import and export functionality for passive keys. Pushed a commit that tweaks the documentation, but otherwise this commit looks good to me.

src/remote_client.rs
marsella: That's a good point. I guess that's a difference between a key fiduciary (who should only get access to keys delegated to them) vs the service provider (who should get all the keys for a user). So I guess in my head this behavior changes based on who's authenticated in the session. I can add that distinction in the docs or just ignore it for now (we won't have to support that difference until we have a non-SP key fiduciary, which is probably a while in the future).

src/remote_client.rs
```diff
@@ -0,0 +1,101 @@
...
+/// or a key fiduciary. This request will fail if the calling party
+/// is not from one of those entities.
+///
+/// TODO #30 (design, implementation): Pass a session
+#[allow(unused)]
+pub fn request_transaction_signature(
+    transaction_approval_request: TransactionApprovalRequest,
+) -> Result<TransactionSignature, Error> {
+    todo!()
+}
```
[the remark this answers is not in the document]
marsella: Made #47 to address the TAR ID problem.

The `UserId` is a field in the `TransactionApprovalRequest`. Should it also be passed separately to the function?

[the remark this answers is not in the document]
indomitableSwan: > Made #47 to address the TAR ID problem.
> 
> The `UserId` is a field in the `TransactionApprovalRequest`. Should it also be passed separately to the function?

I think this specific UserId question needs to be addressed in #47, not here.

The documentation should be adjusted here to include known session requirements with respect to entity authentication.

src/remote_client.rs
indomitableSwan: I think we leave additional functionality for fetching public info on other types of keys out for now.

src/remote_client.rs
```diff
@@ -0,0 +1,101 @@
...
+#[derive(Debug, Error)]
+pub enum Error {}
+
+/// Generate a new, distributed, [`Passive`][crate::keys::Passive]
+/// [`DigitalAssetKey`](crate::keys::DigitalAssetKey) with the given use
+/// restrictions for the [`UserId`], and compatible with the specified
+/// blockchain.
+///
+/// TODO #25 (implementation): Pass the appropriate blockchain as a parameter.
+/// TODO #30 (design, implementation): Pass a session
```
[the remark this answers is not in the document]
vidyaap: I think it makes sense to add a `register_pass_user` function since we need a way to know whether a key fiduciary is allowed to take actions on behalf of a given user in other functions. I made a function to do that, and tried to write the previous sentence in as an assumption in the docs, but please feel free to edit!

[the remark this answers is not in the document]
vidyaap: On the outputs, do you mean to document the output on every function or just the `create_passive_digital_asset_key` function specifically?

src/remote_client.rs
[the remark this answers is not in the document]
vidyaap: It looks like there are 2 issues here: what is the scope of the `TransactionApprovalRequest` type (i.e. when should we treat it as a single object vs. separate inputs), and what are its contents. I think #47 is part of the TAR scope issue, so I’ve edited the title and left a comment there to reflect that. Made #48 to address the TAR contents issue.

src/remote_client.rs
```diff
@@ -0,0 +1,134 @@
...
+///
+/// Assumption: The [`import_asset_key`] functionality is called by the service provider. This is cryptographically enforced with
+/// an authenticated session between the key server and the service provider. This request will fail otherwise.
+///
+/// TODO #25 (implementation): Pass the appropriate blockchain as a parameter.
+/// TODO #30 (design, implementation): Pass a session
+#[allow(unused)]
+pub fn import_asset_key(
+    user_id: UserId,
+    restriction: impl UseRestriction,
```
marsella, line 67: Doesn't this need to pass the actual key material, too? The `UseRestriction` describes key material but does not contain it. I would expect this to take either a complete digital asset key, like:
```
pub fn import_asset_key<R: UseRestriction>(
    user_id: UserId,
    key: DigitalAssetKey<Passive, R>) -> Result<...>
```

or some key material of another type - maybe just bytes for now? I'd also suggest editing the 2nd line of the docs to reflect that we're literally passing the key material, not an ID or link to it:
> The key material will be imported as a [`Passive`] [`DigitalAssetKey`].

Also re: docs: links to `DigitalAssetKey` will fail because that type is private right now. We'll have to make it public if it becomes a parameter, or leave it private and remove the doc link.

src/remote_client.rs
```diff
@@ -0,0 +1,134 @@
...
+///
+/// The [`KeyId`] must correspond to a [`Passive`][crate::keys::Passive]
+/// [`DigitalAssetKey`](crate::keys::DigitalAssetKey).
+///
+/// Assumption: The [`export_asset_key`] functionality is called by the service provider. This is cryptographically enforced with
+/// an authenticated session between the key server and the service provider. This request will fail otherwise.
+///
+/// TODO #30 (design, implementation): Pass a session
+#[allow(unused)]
+pub fn export_asset_key(user_id: UserId, key_id: &KeyId) -> Result<Passive, Error> {
```
marsella, line 82: See comment above - the successful result type should be actual key material, like a `DigitalAssetKey`, not the description of the use permission.

src/remote_client.rs
vidyaap: Oh right... I guess I'm not sure about passing the DigitalAssetKey type since it has a `KeyId` field which we have to assign. Accepting bytes for now makes sense to me, but maybe there's another issue to create here?

Re: bytes: do we take this type from this [crate](https://docs.rs/bytes/latest/bytes/) or is there another, and what size should we default to?

Re: docs: It looks like `DigitalAssetKey` is marked public?

src/remote_client.rs
vidyaap: Oversight on my part - I keep treating that type as a type of key. But I'm having similar hangups about returning the `DigitalAssetKey` type as I was for passing it as a parameter. There are fields in it that aren't meaningful outside of the system. What would be a good type to represent just the key material?

src/remote_client.rs
marsella: That is a good point, I concur with avoiding the `KeyId` origination problem.

The one in the crate is good! 256 bits = 32 bytes is an appropriate default. I would also suggest using a little wrapper around those 32 `Bytes` called `KeyMaterial` or something, and then making an issue like, "define key material properly".

I made `DigitalAssetKey` private at some point but if it's now public, that's fine! Just wanted to head off the potential CI failure.

src/remote_client.rs
marsella: No worries. If you make a `KeyMaterial` wrapper around `Bytes` as described above, I think it's fine to return that here.

src/remote_client.rs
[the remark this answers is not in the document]
marsella: Thanks for adding all those, Vidya. To be honest, I don't find the explanations particularly useful -- they don't give much context beyond what we already had from the return type. But maybe we should leave them and flesh them out into something more specific when we build out the function implementations.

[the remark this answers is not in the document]
vidyaap: Yeah I think that's what I was running into earlier. At first, it felt more complete to have an explicit statement about what is returned (and also more consistent with other Rust docs) but there is enough detail in the function descriptions and assumptions already that I'm not sure what else to add in the return description.

indomitableSwan (approved):
Lgtm.

Issue #23, opened 2022-06-13
Update module / crate structure


marsella:
Separate the existing modules into crates:

1. dams-crypto (lib)
2. dams-local-client (lib)
3. dams-remote-client (lib)
4. dams-key-server (bin)
5. dams (lib)

`dams` will have all the shared utility stuff that is used in multiple crates.

(Check crate name standards first - should the -s be _s?)

indomitableSwan:
Closed by PR #57.

Issue #42, opened 2022-06-28
Add unstable rustfmt options


marsella:
I think it would be a good best practice to enable [the nightly `rustfmt` option to wrap comments](https://rust-lang.github.io/rustfmt/?version=v1.5.0&search=#wrap_comments). 

We might not want to enable it in the CI, because we might occasionally have exceptions that need to be maintained manually (e.g. if the wrapping breaks a link or a table). But maybe we should enable until there's a specific thing that breaks.

----
Another unstable option we've discussed is to set [the import granularity](https://rust-lang.github.io/rustfmt/?version=v1.5.0&search=#imports_granularity) to `Crate`.

marsella:
Is this closed by #106?

picklenerd:
Yep!

PR #57, opened 2022-07-05
Move files into appropriate crates

vidyaap:
Made this draft PR to make seeing the file changes easier.  After resolving the key-mgmt sub-directory concern, this should close issue #23.

The existing code has been split into 3 libs and a binary (utility, local client, remote client, and key server as listed in #23). The problem I’m having is what to do with the key-mgmt sub-directory itself. @marsella and I decided to split this directory up and put the server-relevant parts into the key-server binary and the client-relevant parts into the local-client library. 

But the following files are causing a circular dependency issue: `key-mgmt/src/key-mgmt/server.rs` (now key-mgmt/dams/src/key-mgmt/server.rs) and `key-mgmt/src/key-mgmt/server/create.rs` (now key-mgmt/dams/src/key-mgmt/server/create.rs).

`create.rs` has a `use` statement for the client’s `SecretInfo` type which would create a dependency to the local-client lib. `server.rs` has a `use` statement for the server’s cli module which would create a dependency to the key-server bin. The local-client lib already has the dams utility lib as a dependency.
- If I split the directory as described above, the key-server would have dependencies to the local-client lib and the dams utility lib, and the local-client already has the dams utility as a dependency
- If I don’t split and leave this whole key-mgmt directory in the dams utility lib, then we will have the dams utility depending on the local-client which depends on the dams utility, creating a circle

If there is no issue with the first bullet, then I will just do that, but I was unsure whether that would also create a circular dependency.  Another look from someone would be appreciated!


vidyaap:
I needed to leave the package tag in the top level Cargo.toml so I could have dependencies in the test suite, but that doesn't seem to be the way to handle that. Anyone know what you're supposed to do in this case?

dams-key-server/src/bin/main.rs
```diff
@@ -2,25 +2,21 @@ use structopt::StructOpt;
...
+    Server(key_server::cli::Cli),
 }
 
 #[tokio::main]
 pub async fn main() {
     let filter = EnvFilter::try_new("info,sqlx::query=warn").unwrap();
     tracing_subscriber::fmt().with_env_filter(filter).init();
 
-    use Cli::{Client, Server};
+    use Cli::Server;
```
gijsvl, line 17: Can we move use statements to the top? I know it was already here, but I like them grouped at the top.

dams-key-server/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "dams-key-server"
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+name = "key_server"
```
gijsvl, line 7: why did we specifically rename and expose a library here? We don't seem to do that for the local client?

dams-local-client/src/api.rs
```diff
@@ -30,7 +30,7 @@ pub struct SessionConfig;
 /// communication session between an asset owner and a key server
 /// that provides mutual authentication, confidentiality, and integrity.
 /// An open `Session` is
-/// required to interact with the [`crate::local_client`] API.
+/// required to interact with the [`crate::api`] API.
```
gijsvl, line 33: I know I made this change, but maybe we have to change this into something more intelligible.

tests/common.rs
```diff
@@ -13,9 +13,8 @@ use tokio::{task::JoinHandle, time::Duration};
 use tracing::info_span;
 use tracing_futures::Instrument;
 
-use da_mgmt::TestLogs;
-
-use da_mgmt::{server::key_mgmt::Command as _, timeout::WithTimeout};
+use dams::{timeout::WithTimeout, TestLogs};
+use key_server::command::Command as _;
```
gijsvl, line 17: What does `use ... as _` do?

tests/local_client.rs
```diff
@@ -1,9 +1,9 @@
-use da_mgmt::{
+use dams::{
     blockchain::Blockchain,
     keys::{KeyId, SelfCustodial, SharedControl, UserId, UserPolicySpecification},
-    local_client::*,
     transaction::TransactionApprovalRequest,
 };
+use dams_local_client::api::*;
```
gijsvl, line 6: Avoid wildcard import

tests/remote_client.rs
```diff
@@ -1,9 +1,7 @@
-use da_mgmt::blockchain::Blockchain;
-use da_mgmt::{
-    keys::{KeyId, KeyMaterial, SharedControl, UserId},
-    remote_client::*,
-    transaction::TransactionApprovalRequest,
-};
+use dams::blockchain::Blockchain;
+use dams::keys::{KeyId, KeyMaterial, SharedControl, UserId};
+use dams::transaction::TransactionApprovalRequest;
+use dams_remote_client::api::*;
```
gijsvl, line 4: avoid wildcard import

dams-key-server/Cargo.toml
vidyaap: Syntax knowledge gap on my part - removed this and adjusted references

dams-local-client/src/api.rs
vidyaap: It might be fine to just say "... with this API" since it's now its own crate and not just the "local_client" mod of a larger crate. Does that work?

tests/common.rs
vidyaap: This was already here, I think it was to avoid the duplicate `Command` name from `process`. Also it's probably left blank since we never use this `Command` by name, only to use the trait functions.

tests/local_client.rs
vidyaap: The wildcard imports in the tests were also here already, so for this and the next comment, if the fmt checks pass, can these stay? Otherwise we would have to import every function and type from each of these files. Or give the import an alias and then prepend each function and type usage with that alias. (If that is preferred, though, happy to make that change!)

dams-key-server/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "dams-key-server"
+version = "0.1.0"
+edition = "2021"
```
marsella, line 4: question: Does this need to be a package if the top-level crate already defines a package? I thought this would just have a binary called `dams-key-server`, but I can't find an example crate with a similar structure to compare.

Cargo.toml
```diff
@@ -1,56 +1,28 @@
...
-path = "src/bin/client.rs"
-
-[[bin]]
-name = "key-mgmt-server"
-path = "src/bin/server.rs"
-
-[features]
-allow_explicit_certificate_trust = []
-
 [dependencies]
```
marsella, line 16: thought: My (possibly incorrect) assumption is that these dependencies are things used for the tests, and are not "passed down" or inherited into the workspace members. If so, consider putting them into `dev-dependencies` instead. Dev dependencies aren't compiled into the release version of a crate, but that might be irrelevant here since the "release version" will only be compiled for the child crates. I guess the primary value of moving them would be (a) to ingrain the habit or (b) to indicate to readers that they didn't miss some top-level source code.

dams-local-client/src/bin/main.rs
```diff
@@ -1,10 +1,9 @@
...
-use da_mgmt::client::key_mgmt::Command;
-use da_mgmt::client::{
-    cli::Client::{Create, Retrieve},
-    defaults::config_path,
-    Cli, Config,
-};
+use dams::config::client::Config;
+use dams::defaults::client::config_path;
+use dams_local_client::cli::Cli;
+use dams_local_client::cli::Client::{Create, Retrieve};
```
marsella, line 5: note: we had discussed using the unstable rustfmt option that would group these together. I updated #42 to reflect that plan, but I think you can skip for now because it will probably change stuff across the repo.

tests/main.rs
```diff
@@ -1,19 +1,19 @@
...
-use da_mgmt::{client, client::key_mgmt::Command};
+use dams_local_client::command::Command;
 use std::fs::OpenOptions;
 use structopt::StructOpt;
 use thiserror::Error;
 
-// Form a client CLI request. These cannot be constructed directly because the
-// CLI types are non-exhaustive.
+/// Form a client CLI request. These cannot be constructed directly because the
+/// CLI types are non-exhaustive.
```
marsella, line 10: Good catch!

tests/remote_client.rs
marsella: This is the same case as above, so I vote to keep it here as well.

dams-local-client/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "dams-local-client"
+version = "0.1.0"
+edition = "2021"
+
+[[bin]]
+name = "local-client-cli"
+path = "src/bin/main.rs"
```
marsella, line 8: thought: eventually we will want to remove this CLI entirely, but I think this PR is not the right place to do it. We will also want to export the library here (the current `api` module, but not `cli` or `command`). Can you make an issue to document the intended external face for this crate?

dams/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
```
marsella, line 1: question: Why don't we need to specify the [lib] separately from the package?

marsella (commented):
This looks great! Everything is where I expected it to be! Thank you for taking this big PR on.

I've left a small handful of questions but most of them are because I don't understand the details of `Cargo.toml`s.

tests/common.rs
marsella: That's exactly it. I think you would only do this for traits, where the trait name is mirrored somewhere, you're not writing a type that includes the trait name, but you are using the trait functions.

dams-local-client/Cargo.toml
vidyaap: Made #79 for this. It might need more details about what to do with the code in `cli` and `command`.

dams/Cargo.toml
vidyaap: It looks like if there's only one thing being exposed in a crate, it is just inferred based on the presence of a lib.rs file vs. a main.rs file. I'd like for someone to confirm that though.

Cargo.toml
vidyaap: Yep, these are all for the tests. Changed to dev-dependencies!

Cargo.toml
```diff
@@ -1,56 +1,28 @@
...
+    "dams-remote-client",
+    "dams-key-server",
+]
+
 [package]
 name = "key-mgmt"
 version = "0.1.0"
 edition = "2018"
 license = "MIT"
 description = "implementation of a secure key mangement system"
```
leboiko, line 14: On the `crate-structure` branch we dont have a target for `key-mgm` on the root of the repository, right? 
image

image

dams-key-server/Cargo.toml
gijsvl: I believe in the current structure we need it to be a package, not sure what else it is going to be? A possibility is to just have sub-folders, but I thought the whole idea of having the different crates is that you don't necessarily have to import all code (both from the client's as well as the server's perspective).

marsella (approved):
This looks good to me, thank you for handling this monster!

Vidya, you mentioned you had a pending comment on remaining open comment thread. If you go to the file structure and hit "review changes -> submit review", that should appear for the rest of us. I'm pretty sure you resolved it, but for historical posterity, I think it's useful to publish it.

Cargo.toml
vidyaap: Right, the package tag in the top level Cargo.toml was left there so we could have dev-dependencies for the tests. On your suggestion, I moved all the integration tests to their own crate which has the other crates as dependencies instead.

PR #58, opened 2022-07-06
Opaque authenticate

gijsvl:
To review this opaque authentication I've created a PR into `opaque-register`, once that branch has been merged, we will change this PR to merge with `main` branch. But in this way the PR can already be reviewed


tests/local_client.rs
```diff
@@ -12,19 +12,39 @@ use std::fs::File;
...
-fn open_session_not_implemented() {
-    let _result = Session::open(
-        UserId::default(),
+#[tokio::test]
+async fn test_session() {
+    let server_future = common::setup().await;
+    test_register_session().await;
+    test_open_session().await;
+    common::teardown(server_future).await;
+}
```
marsella, line 21: question: It looks like the point of this is to bundle multiple tests together under the same server setup and teardown. This raises the same questions I had in #55 about what the different goals are for this test suite vs the `main+common` test suite. The point of that suite is to do exactly this -- make common setup and teardown functions for all of our tests.

src/local_client.rs
```diff
@@ -81,6 +83,7 @@ impl Default for SessionConfig {
 #[allow(unused)]
 pub struct Session {
     pub config: SessionConfig,
+    pub session_key: [u8; 64],
```
marsella, line 86: todo: Make this a private field, then add a constructor and a public accessor method.

src/cli/client.rs
```diff
@@ -44,6 +45,20 @@ pub struct Register {
...
+pub struct Authenticate {
+    /// The username to register
+    #[structopt(long)]
+    pub username: String,
+    /// The password to register
+    #[structopt(long)]
+    pub password: String,
+    /// The `keymgmt://` address for the server.
+    pub server: KeyMgmtAddress,
+}
```
marsella, line 60: todo: I have the same feedback as in #55:
- Replace these `Strings` with the actual types we want to use (`UserId`, `Password`).
- Rename `username` to `user_id` -- we don't have plans to add a meaningful / human-readable username.
- Consider deprecating the CLI and just defining an `authenticate` function that takes these as parameters.

src/key_mgmt/client.rs
```diff
@@ -24,6 +28,17 @@ pub type RegisterStartReceived = RegistrationResponse<OpaqueCipherSuite>;
 /// The object that the client sends to the server to finish registration using OPAQUE
 pub type RegisterFinish = RegistrationUpload<OpaqueCipherSuite>;
 
+/// The object that the client sends to the server when registering using OPAQUE
+#[derive(Debug, Serialize, Deserialize)]
+pub struct AuthStart {
+    pub request: CredentialRequest<OpaqueCipherSuite>,
+    pub username: String,
```
marsella, line 35: todo: replace with `user_id: UserId`.

src/key_mgmt/server/authenticate.rs
```diff
@@ -0,0 +1,72 @@
...
+    protocol::authenticate,
+    server::{config::Service, Config},
+    timeout::WithTimeout,
+};
+use anyhow::Context;
+use opaque_ke::{ServerLogin, ServerLoginStartParameters, ServerRegistration};
+use rand::rngs::{OsRng, StdRng};
+use transport::server::{Chan, SessionKey};
+
+pub struct Authenticate;
```
marsella, line 14: suggestion: I think this structure is an artifact of a non-ideal design decision in zeekoe that assumed all commands coming into the server would have the same set of parameters, so we defined a `Command` trait and these unit structs to implement their corresponding `run` functions.

In fact, that wasn't a helpful structure and I think we dropped the trait (in any case, it doesn't appear here). Maybe @leboiko can weigh in on whether it's better to have a unit struct with a `run` function vs just having an unattached `authenticate` or `run_authenticate` function?

In either case, the `run` function here takes a bunch of ignored parameters that I would prefer to delete. Is there a benefit to keeping them around?

src/key_mgmt/server/authenticate.rs
```diff
@@ -0,0 +1,72 @@
...
+        let server_setup = create_or_retrieve_server_key_opaque(rng, service)
+            .context("could not find/create server key")?;
+
+        let (auth_start, chan) = chan
+            .recv()
+            .with_timeout(service.message_timeout)
+            .await
+            .context("Did not receive RegisterStart")??;
+
+        let password_file_serialized = retrieve_opaque(service, auth_start.username.clone());
```
marsella, line 35: suggestion: when you replace `username: String` with `user_id: UserId`, you should be able to drop this `clone` because user ids are `Copy` (if not, you should make them `Copy`).

src/key_mgmt/server/authenticate.rs
```diff
@@ -0,0 +1,72 @@
...
+        if password_file_serialized.is_err() {
+            abort!(in chan return authenticate::Error::UsernameDoesNotExist)
+        }
+        let password_file = ServerRegistration::<OpaqueCipherSuite>::deserialize(
```
marsella, line 39: nitpick: Can we rename this to something about `authentication` or `opaque` or `client_auth` instead of `password`? I don't want to confuse future developers.

src/key_mgmt/server/authenticate.rs
```diff
@@ -0,0 +1,72 @@
...
+            &password_file_serialized.unwrap(),
+        )
+        .unwrap();
```
marsella, line 42: todo: please remove the `unwrap`s throughout.

src/key_mgmt/server/authenticate.rs
```diff
@@ -0,0 +1,72 @@
...
+        let mut rng = OsRng;
```
marsella, line 43: question: Is there a reason you can't use the `StdRng` passed in as a parameter? I would prefer to keep our rng behavior a bit more centralized.

src/local_client.rs
```diff
@@ -113,24 +135,21 @@ impl Session {
         config: &SessionConfig,
     ) -> Result<Self, SessionError> {
         let result = Register {
-            username: user_id.0,
-            password: password.0,
+            username: user_id.0.clone(),
```
marsella, line 138: suggestion: I think it is also fine to pass the `UserId` by reference. 

This comes back to my other comment -- I don't think it makes sense to retcon a CLI tool to call a function. I'd rather see just a plain `register` function that also takes a `user_id` by reference so we don't have to clone all this stuff here (we're not creating anything that will persist outside of this function, so it doesn't make sense that we'd have to be cloning these objects).

tests/main.rs
```diff
@@ -180,6 +199,7 @@ enum Operation {
     Create,
     Register,
     Retrieve,
+    Authenticate,
```
marsella, line 202: todo: I noted this in #55, too, but I think this should take a username and password instead of hard-coding them.

src/key_mgmt/client/authenticate.rs
```diff
@@ -0,0 +1,75 @@
...
+impl Command for Authenticate {
+    type Output = [u8; 64];
+    async fn run(self, config: self::Config) -> Result<Self::Output, anyhow::Error> {
+        let Self {
+            username,
+            password,
+            server: address,
+        } = self;
+
+        let mut rng = OsRng;
```
marsella, line 25: todo: I don't think `OsRng` is the best choice here, and I don't like distributing the RNG choice to each client operation. Our randomness best practices recommend `StdRng` or `ThreadRng`. Perhaps you can make an `rng` function in the main module file (`client.rs`) that returns one of the recommended versions with the type `impl CryptoRng + RngCore`and call it from each function that needs a RNG. That way we have the flexibility to switch it out across the whole client more easily.

src/key_mgmt/client/authenticate.rs
```diff
@@ -0,0 +1,75 @@
...
+
+        // Connect with the server...
+        let (_session_key, chan) = connect(&config, &address)
+            .await
+            .context("Failed to connect to server")?;
+
+        // ...and select the Create session
```
marsella, line 32: todo: replace this comment and the context with `authenticate session`, not create.

src/key_mgmt/client/authenticate.rs
```diff
@@ -0,0 +1,75 @@
...
+        let chan = chan
+            .choose::<3>()
+            .await
+            .context("Failed to select create secret session")?;
+
+        let client_login_start_result =
+            ClientLogin::<OpaqueCipherSuite>::start(&mut rng, password.as_bytes()).unwrap();
```
marsella, line 39: todo: remove this unwrap.

tests/main.rs
```diff
@@ -87,6 +87,10 @@ fn tests() -> Vec<Test> {
             name: "Register as a client to the server".to_string(),
             operations: vec![(Operation::Register, Outcome { error: None })],
         },
+        Test {
+            name: "Authenticate as a client to the server".to_string(),
+            operations: vec![(Operation::Authenticate, Outcome { error: None })],
```
marsella, line 92: todo: I expect this test to fail without a registration step first. I'm guessing it doesn't because the `test_user` username is hard-coded across every register and authenticate step. Can you update this test to throw an error and make another passing test that registers and authenticates?

marsella (changes requested):
Thanks for taking this on -- I appreciate the work you put in to figure out the opaque library and start integrating it into the APIs.

My primary issues here are similar to those in #55:
- I think we should stop using/expanding the client CLI. I think it will improve code quality to have a simpler, more appropriate interface for calling client functions.
- There's a lot of `unwrap`s around this PR. I think it's important to set the foundation for non-panicking error handling from the beginning, even in pieces of code that we don't intend to use for the long run. I commented some of them, but I don't want to approve this until all the unwraps are either gone or have documentation explaining why they won't panic.
- The testing in this repo is evolving into two dueling systems with unclear (to me) distinctions. I think we should converge on one -- probably the custom test harness.

There are also a few tests I would have expected to see here that didn't appear, like
- registering with the same user twice fails
- a registered user can open multiple sessions
- opening a session with an unregistered user id fails

I recognize that this codebase comes with some technical baggage that we imported from zeekoe. I'm hoping to reduce this as much as possible in the early stages of this project. I've made a couple comments to that effect, pointing out places that I think are historical artifacts that no longer serve us.

Apologies for some duplicated comments within this PR and across this and #55.

src/local_client.rs
```diff
@@ -108,29 +234,88 @@ impl Session {
...
+            .await
+            .context("Failed to connect to server")?;
+
+        // ...and select the Register session
+        let chan = chan
+            .choose::<1>()
+            .await
+            .context("Failed to select Register session")?;
+
+        let client_registration_start_result = match ClientRegistration::<OpaqueCipherSuite>::start(
```
marsella, line 278: todo: Replace `match` with a `map_err`.

src/local_client.rs
```diff
@@ -108,29 +234,88 @@ impl Session {
...
             }
+        }
+    }
+
+    async fn do_register(
+        user_id: &UserId,
+        password: &Password,
+        server: KeyMgmtAddress,
+        config: self::Config,
+    ) -> Result<(), anyhow::Error> {
```
marsella, line 264: todo: Replace the error type with our `SessionError` and replace all the hard-coded `anyhow` errors with new variants in that enum. @vidyaap

src/local_client.rs
```diff
@@ -108,29 +234,88 @@ impl Session {
...
+            .context("Failed to send RegisterStart")?;
+
+        offer_abort!(in chan as Client);
+
+        let (register_start_received, chan) = chan
+            .recv()
+            .await
+            .context("Failed to recv RegisterStartReceived from server")?;
+
+        let client_finish_registration_result = match client_registration_start_result.state.finish(
```
marsella, line 303: todo: Replace `match` with `map_err`. @vidyaap

src/opaque_storage.rs
```diff
@@ -1,47 +1,104 @@
...
+    }
+}
+
+//TODO: replace with decent and secure storage for server keys #56
+pub(crate) fn create_or_retrieve_server_key_opaque(
+    mut rng: StdRng,
+    service: &Service,
+) -> Result<ServerSetup<OpaqueCipherSuite, PrivateKey<Ristretto255>>, Error> {
+    let server_key_file = File::open(service.opaque_server_key.clone());
+    match server_key_file {
```
marsella, line 56: todo: Can you add short inline comments describing which branch is "create new" and which is "retrieve existing"? @vidyaap

tests/main.rs
```diff
@@ -104,27 +192,40 @@ impl Test {
...
+                        .expect("Failed to load client config");
+                    let config = SessionConfig::new(
+                        client_config,
+                        KeyMgmtAddress::from_str("keymgmt://localhost").unwrap(),
                     );
-                    est.run(client_config.clone()).await.map(|_| ())
+                    Session::register(user_id, password, &config)
+                        .await
+                        .map(|_| ())
+                        .map_err(|e| anyhow!(e.to_string()))
```
marsella, line 210: todo: Can we replace the `anyhow!` with a plain `into`? I think the `thiserror::Error` trait automatically implements that conversion. If so, there's another instance down on L228. @vidyaap

tests/main.rs
```diff
@@ -104,27 +192,40 @@ impl Test {
...
-                        vec![
-                            "register",
-                            "keymgmt://localhost",
-                            "--username",
-                            "test_user",
-                            "--password",
-                            "test_password"
-                        ]
+                Register(user_id, password) => {
+                    let client_config = client::Config::load(common::CLIENT_CONFIG)
```
marsella, line 200: todo: Use the client config passed in as a parameter. @vidyaap

tests/main.rs
```diff
@@ -104,27 +192,40 @@ impl Test {
...
+                Retrieve => {
+                    let retrieve = client_cli!(Retrieve, vec!["retrieve", "keymgmt://localhost"]);
+                    retrieve.run(client_config.clone()).await.map(|_| ())
                 }
-                Operation::Retrieve => {
-                    let est = client_cli!(Retrieve, vec!["retrieve", "keymgmt://localhost"]);
-                    est.run(client_config.clone()).await.map(|_| ())
+
+                Authenticate(user_id, password) => {
+                    let client_config = client::Config::load(common::CLIENT_CONFIG)
```
marsella, line 218: todo: use the client config passed in as a parameter. @vidyaap

marsella (commented):
Thanks for making those changes, this is looking much better!

There are two main cryptography issues I see here:
1. We're using a version of the opaque library that has been yanked / unpublished from the crate index. I think we should stop using this.
2. The random number generator selection behavior is not very centralized right now - we just create one whenever we need one. This doesn't adhere to our RNG standards and I think we should try as much as possible to make RNGs mutable parameters that are passed around to functions.

There are also a bunch of Rust idiom comments. @vidyaap has agreed to take on most of these and I've tried to tag her in the relevant comments, but it might be instructive for @gijsvl to read through those to see what to do in the future.

Cargo.toml
```diff
@@ -27,6 +27,7 @@ allow_explicit_certificate_trust = []
 [dependencies]
```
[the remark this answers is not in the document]
indomitableSwan: I'd prefer to use the most recent unstable version for now. We want to be using a library that matches the current version of the RFC and may even want to be helping to improve the library, as capacity allows. For deployment, the only alternative I can think of (besides switching to a different aPAKE entirely) is to write our own implementation of OPAQUE that tracks the most recent changes, which seems like a much heavier lift.

There are a quite a few changes in the RFC since the one captured in 1.2.0, most of which we care about.

src/key_mgmt/server/authenticate.rs
```diff
@@ -32,24 +31,22 @@ impl Authenticate {
...
-            abort!(in chan return authenticate::Error::UsernameDoesNotExist)
-        }
-        let password_file = ServerRegistration::<OpaqueCipherSuite>::deserialize(
-            &password_file_serialized.unwrap(),
-        )
-        .unwrap();
+        let server_registration = match retrieve_opaque(service, auth_start.user_id()) {
+            Ok(server_registration) => server_registration,
+            Err(_) => abort!(in chan return authenticate::Error::UserIdDoesNotExist),
+        };
```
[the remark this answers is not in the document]
gijsvl: I don't believe `map_err` achieves what we want here, because it returns a new result, while in this case we either want the unwrapped object or send an abort... Unless I'm missing something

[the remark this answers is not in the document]
marsella: Oh, good catch.

src/local_client.rs
```diff
@@ -108,29 +234,88 @@ impl Session {
...
+        }
+    }
+
+    async fn do_register(
+        user_id: &UserId,
+        password: &Password,
+        server: KeyMgmtAddress,
+        config: self::Config,
+    ) -> Result<(), anyhow::Error> {
+        let mut rng = OsRng;
```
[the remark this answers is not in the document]
marsella: Update: Gijs replaced the `OsRng` with a `StdRng` parameter, but I think it's still a good option to use the `impl CryptoRng` type to describe the behavior we want, rather than restricting ourselves to the specific `StdRng`.

tests/main.rs
marsella: I wasn't clear about this. The execute function takes a `client_config` as a parameter, so we can use that one instead of making it anew here. @vidyaap, can you revert the change that adds a client config as a parameter to the `Operation` enum variants?

src/local_client.rs
vidyaap: I'm not sure what to do with the `context` errors here though. It looks like because they are from `anyhow` errors in the `connect()` function, that the return type still needs to be the `anyhow` one.

vidyaap: I could probably leave the return type on this and the next comment, but still change the hard-coded `anyhow`s to `SessionErrors` though

tests/main.rs
vidyaap: I think the trade-off is having to hard code unique strings into the tests vs. having to go into the DB to delete things. As long as each test is an isolated set of events, which both of these options allow, then it’s up to us what we prefer. Maybe we can leave it as is until we implement the first pass of the DB and see how complicated switching it would be then?

marsella: That sounds like a good plan to me.

tests/main.rs
```diff
@@ -81,15 +88,96 @@ fn tests() -> Vec<Test> {
...
+                    },
+                ),
+                (
+                    Register(
+                        UserId::from_str("sameUser").unwrap(),
+                        Password::from_str("testPassword2").unwrap(),
+                    ),
+                    Outcome {
+                        error: Some(Client),
+                        expected_error: Some(anyhow!("RegistrationFailed")),
```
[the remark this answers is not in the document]
vidyaap: Probably, we can make an enum in the tests to wrap around our custom errors (including `SessionError` in this case) so we can use that enum as the parameter of the `Outcome` type. Made #86 as a placeholder issue for this, but it will need fleshing out.

src/local_client.rs
vidyaap: Left the return types, but changed as many `context`s and hard-coded `anyhow`s as I could. The `offer_abort!` macro is the only thing left that makes us return anyhow errors here, so I made #85 to handle that later.

marsella: Thanks, I forgot about all the `context`s!

src/protocol.rs
```diff
@@ -114,6 +166,7 @@ pub type KeyMgmt = Session! {
         0 => Create,
         1 => Register,
         2 => Retrieve,
+        3 => Authenticate,
```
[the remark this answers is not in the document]
marsella: Added #87 to capture this.

marsella (approved):
Thanks @vidyaap and @gijsvl for addressing all these comments! Looks much improved.

Issue #92, opened 2022-07-20
Update OPAQUE code to work with mongoDB


vidyaap:
Right now, the OPAQUE information is stored in a file-based system. We want to move the user authentication information to mongoDB instead.

Issue #91, opened 2022-07-20
Define a mongoDB User model


vidyaap:
For now, this just needs to be a simple User model that holds:
- [x] a user ID (i.e. the `UserId` type, not the mongo-assigned `_id`)
- [x] a list of secrets
- [x] the OPAQUE information that the server needs to authenticate the user

PR #95, opened 2022-07-22
Opaque with db

### LINKED ISSUE #83: Add infrastructure for MongoDB to repo
This task is to create the initial setup to run a MongoDB database in the key-server crate. This issue can be closed when:
- [x] the mongodb crate has been added as a dependency to the key-server
- [x] there is a module in the key-server for the database
- [x] there is code to setup the initial connection to the database, and it is called when the server starts

vidyaap:
Closes #83 and #92.

Making this a draft first to resolve some Rust/code quality questions. These are the concerns I have right now:
- The [mongodb crate docs](https://mongodb.github.io/mongo-rust-driver/manual/performance.html#lifetime) claim that cloning of the DB handle is fine/preferred. But is there too much cloning of the DB handle right now?
- The mongodb crate functions sometimes return Options wrapped in Results. Is that construction idiomatically correct? Is there a better way to handle it?
- Some `dotenv` docs say to use the package using `extern crate` and some just say `use dotenv`. Is there one that is more "correct" to use?
- Is using .env files standard for Rust?
- I'm getting a clippy warning that says I have too many variables (8 instead of 7) in 2 functions. Is it ok to suppress that warning or should I change the code somehow?


leboiko:
> Is using .env files standard for Rust?
for applications, yes, but not for libraries. For local development we can provide a `.env.example` containing example values that can be used for the local build and ask the users to copy that to `.env` before building (or add that as part of some build script)

leboiko:
> I'm getting a clippy warning that says I have too many variables (8 instead of 7) in 2 functions. Is it ok to suppress that warning or should I change the code somehow?

Clippy is not a perfect tool, but it has been improved a lot since its early days. Its possible to add a `#allow(clippy::too_many_arguments)]` to bypass that, but let me try to understand if thats the case here or if we have another kind of problem

leboiko:
> Some dotenv docs say to use the package using extern crate and some just say use dotenv. Is there one that is more "correct" to use?

Before Rust 2018 we needed to specify the `extern` keyword when we wanted to bring external crate names (not part of the workspace, for instance) into scope, however we dont need to do that anymore. One of the few cases where we still need to use the `extern` keyword is to either to define foreign function interfaces, such as on [this example](https://github.com/dingelish/teaclave-sgx-sdk/blob/4874effb57ae3759fce5b8eb2097d98710888ed4/samplecode/hello-rust/enclave/src/lib.rs#L36), or to access foreign interfaces, and in this case foreign means build in other languages. If you want there is more information available on the following link: **[Calling foreign functions](https://doc.rust-lang.org/nomicon/ffi.html#calling-foreign-functions)**

leboiko:
> The mongodb crate functions sometimes return Options wrapped in Results. Is that construction idiomatically correct? Is there a better way to handle it?

If I'm understanding that correctly you are getting something like `Result<Option<someType>, CustomError>`, is that correct? If so, that is correct. Think about a situation where you have an endpoint that returns a list of keys from a given User. There may be a situation where the user don't have any key, and in that case, it don't make sense to return a custom error, but it may make more sense to return a `None` (every Option can return either Some(value) or None) instead.

leboiko:
> The [mongodb crate docs](https://mongodb.github.io/mongo-rust-driver/manual/performance.html#lifetime) claim that cloning of the DB handle is fine/preferred. But is there too much cloning of the DB handle right now?

Usually "safe to clone" means that the object we are cloning is living inside an Arc, and with that, cloning the object does not dupplicate it, but just increase the reference counter of the object. As you can see [here ](https://github.com/mongodb/mongo-rust-driver/blob/1932e113524e236602678bf5eed2cfcb5bd7c237/src/client/mod.rs#L122-L126), the `Client` indeed lives inside an Arc

dams-key-server/src/command/authenticate.rs
```diff
@@ -33,9 +37,9 @@ impl Authenticate {
             .context("Did not receive RegisterStart")??;
 
         let (credential_request, user_id) = auth_start.into_parts();
-        let server_registration = match retrieve_opaque(service, &user_id) {
-            Ok(server_registration) => server_registration,
-            Err(_) => abort!(in chan return authenticate::Error::UserIdDoesNotExist),
+        let server_registration = match User::find_user(&db, user_id.clone()).await? {
```
FF-agonzalez, line 40: Im thinking of the possibility to create a fn with a descriptive name to clarify the conection between a server registration and the result of find a user. What do you think?

dams/Cargo.toml
```diff
@@ -22,6 +22,7 @@ http = "0"
 http-serde = "1"
 humantime = "2"
 humantime-serde = "1"
+mongodb = "2.1"
```
FF-agonzalez, line 25: Is there any reason for why we are using this version?

dams-tests/Cargo.toml
```diff
@@ -11,7 +11,9 @@ dams = { version = "0.1.0", path = "../dams" }
 dams-key-server = { version = "0.1.0", path = "../dams-key-server" }
 dams-local-client = { version = "0.1.0", path = "../dams-local-client", features=["allow_explicit_certificate_trust"] }
 dams-remote-client = { version = "0.1.0", path = "../dams-remote-client" }
+dotenv = "0.15.0"
 futures = "0"
+mongodb = "2.1"
```
FF-agonzalez, line 16: Same here. Is there any reason for why we are using this version?

dams-key-server/Cargo.toml
```diff
@@ -18,7 +18,9 @@ dialectic = { git = "https://github.com/boltlabs-inc/dialectic.git", branch = "m
 dialectic-tokio-serde = { git = "https://github.com/boltlabs-inc/dialectic.git", branch = "main" }
 dialectic-tokio-serde-bincode = { git = "https://github.com/boltlabs-inc/dialectic.git", branch = "main" }
 dialectic-reconnect = { features = ["serde", "humantime-serde"], git = "https://github.com/boltlabs-inc/dialectic.git", branch = "main" }
+dotenv = "0.15.0"
 futures = "0"
+mongodb = "2.1"
```
FF-agonzalez, line 23: Same here. Is there any reason for why we are using this version?

vidyaap:
> If I'm understanding that correctly you are getting something like `Result<Option<someType>, CustomError>`, is that correct? ...

Yes except the "CustomError" here is just the mongodb Error that the crate functions throw anyway. I was hoping there was a way to turn `Result<Option<someType>, MongoError>` into `Result<someType, MongoError>` besides just matching, but if this is standard Rust practice anyway then I'll leave it as is.

vidyaap:
> Usually "safe to clone" means that the object we are cloning is living inside an Arc, and with that, ...

Yep thanks, I saw that, just wanted to make sure if there was such a thing as "too much" cloning even if it is "safe to clone"?

leboiko:
> > Usually "safe to clone" means that the object we are cloning is living inside an Arc, and with that, ...
> 
> Yep thanks, I saw that, just wanted to make sure if there was such a thing as "too much" cloning even if it is "safe to clone"?

Depending on the situation it may make sense to have a `Config` struct that holds the DB clients, for instance, and share that across the whole application. We can combine Arc and Mutex (if it requires shared **mutable** state) so its safe to share across threads. This way we can limit the amount of clones, but we are going to have threads "fighting" for the lock instead. Its a matter of how we want to design the iterations

dams-key-server/src/bin/main.rs
```diff
@@ -13,13 +16,20 @@ pub enum Cli {
...
     tracing_subscriber::fmt().with_env_filter(filter).init();
 
-    let result = match Cli::from_args() {
-        Server(cli) => server::main_with_cli(cli).await,
-    };
-    if let Err(e) = result {
-        error!("{}, caused by: {}", e, e.root_cause());
+    let db = database::connect_to_mongo().await;
+
+    if db.is_ok() {
```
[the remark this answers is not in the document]
vidyaap: Not sure I understand which part should be moved to a function; what would that function return? For context, I ended up explicitly checking `is_ok()` like this because of the return types for the tokio main function.

dams/Cargo.toml
vidyaap: Updated to 2.2.2; I think I was looking at outdated docs. (same for the other 2 version comments)

dams-key-server/src/command/authenticate.rs
vidyaap: Can you clarify why the connection doesn't make sense? I think it might just be a matter of changing the name `opaque_information` to `server_registration` in the User model. They're the same thing here.

dams/Cargo.toml
FF-agonzalez: Great! 2.3 version is also available. Having the 2.3 Rust driver give us the versatile to choose between MongoDB version. For example, if we want to use MongoDb 6.0 (or all the old versions) we  could only use it with the 2.3. Instead if we use the 2.2 we are limiting the version to 5.0

For more context: https://www.mongodb.com/docs/drivers/rust/

dams-key-server/src/bin/main.rs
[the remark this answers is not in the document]
FF-agonzalez: `    let db = database::connect_to_mongo().await;

    if db.is_ok() {
        let result = match Cli::from_args() {
            Server(cli) => server::main_with_cli(cli, db.unwrap().clone()).await,
        };
        if let Err(e) = result {
            error!("{}, caused by: {}", e, e.root_cause());
        }
    } else {
        error!("Unable to connect to mongo");
    }`
    
   
In this section your are making a decision based on the is_ok() of the database.

It would be nice to encapsulate this portion of code in a function for example that is responsible for this behavior and the main function uses it. With this approach it is avoided to have a very long main function in the future and it is also easier in the future to differentiate what each function does. Something like a "single responsibility"

dams-key-server/src/command/authenticate.rs
FF-agonzalez: I think that the lack of context on my part is not letting me understand the reason for the relationship between a server_registration and the search for a user in the database.

What would the connection be?

vidyaap: It's sort of parallel to when you compare a password someone entered to whatever is stored in the DB, except that when using OPAQUE, what we store as authentication information for the user is not a password but instead, this "server registration" type.  If I change the field in the User struct from `opaque_information` to `server_registration` is that clearer?

FF-agonzalez: Oh I see! Now is much clear for me. Yes, I think that is clearer!

dams/Cargo.toml
vidyaap: Ah, this time my crates.io page needed to be reloaded! Updated again

dams-key-server/src/database/user.rs
```diff
@@ -0,0 +1,35 @@
...
+    bson::{doc, oid::ObjectId},
+    error::Error,
+    Database,
+};
+use opaque_ke::ServerRegistration;
+
+/// Create a new [`User`] with their authentication information and insert it into the MongoDB database.
+pub async fn create_user(
+    db: &Database,
+    user_id: UserId,
```
marsella, line 20: suggestion: consider taking the `user_id` by reference and cloning it inside this function. I think it's a bit nicer to keep the clone as close to the place where it's necessary as possible.

dams-key-server/src/database/user.rs
```diff
@@ -0,0 +1,35 @@
...
+    server_registration: ServerRegistration<OpaqueCipherSuite>,
+) -> Result<Option<ObjectId>, Error> {
+    let collection = db.collection::<User>("users");
+    let new_user = User::new(user_id, server_registration);
+    let insert_one_res = collection.insert_one(new_user, None).await?;
+    Ok(insert_one_res.inserted_id.as_object_id())
+}
+
+/// Find a [`User`] by their `user_id`. This is different from the Mongo-assigned `_id` field.
+pub async fn find_user(db: &Database, user_id: UserId) -> Result<Option<User>, Error> {
```
marsella, line 30: suggestion: I think this can take the `user_id` by reference and not need to clone it at all.

dams-tests/tests/common.rs
```diff
@@ -81,7 +82,7 @@ pub async fn setup() -> ServerFuture {
     #[allow(clippy::infallible_destructuring_match)]
     let run = server_cli!(Run, vec!["run"]);
     let server_handle = tokio::spawn(
-        run.run(server_config)
+        run.run(server_config, db.clone())
```
marsella, line 85: todo: I don't think this clone is necessary.

dams-tests/tests/main.rs
```diff
@@ -83,7 +90,7 @@ pub async fn integration_tests() {
     if !errors.is_empty() {
         panic!("Test failed: {:?}", errors);
     } else {
-        common::teardown(server_future).await;
+        common::teardown(server_future, db.clone()).await;
```
marsella, line 93: todo: I don't think this clone is necessary.

dams-tests/tests/main.rs
```diff
@@ -33,7 +33,14 @@ macro_rules! client_cli {
...
 pub async fn integration_tests() {
-    let server_future = common::setup().await;
+    // Read environment variables from .env file
+    dotenv::dotenv().ok();
+    let db = match database::connect_to_mongo().await {
+        Ok(db) => db,
+        Err(_) => panic!("Unable to connect to Mongo"),
+    };
+    let _ = db.create_collection("users", None).await;
+    let server_future = common::setup(db.clone()).await;
```
marsella, line 43: todo: I don't think this clone is necessary.

dams-tests/tests/main.rs
```diff
@@ -33,7 +33,14 @@ macro_rules! client_cli {
 
 #[tokio::test]
 pub async fn integration_tests() {
-    let server_future = common::setup().await;
+    // Read environment variables from .env file
+    dotenv::dotenv().ok();
+    let db = match database::connect_to_mongo().await {
+        Ok(db) => db,
+        Err(_) => panic!("Unable to connect to Mongo"),
+    };
```
marsella, line 41: suggestion: You can replace the `match` with `expect`, which is like `unwrap` but throws a custom panic message if it's an error.

dams/src/user.rs
```diff
@@ -0,0 +1,67 @@
...
+        server_registration: ServerRegistration<OpaqueCipherSuite>,
+    ) -> Self {
+        User {
+            user_id,
+            secrets: Vec::new(),
+            server_registration,
+        }
+    }
+
+    pub fn server_registration(self) -> ServerRegistration<OpaqueCipherSuite> {
```
marsella, line 64: thought: This function consumes the `User` (e.g. once you access the `server_registration`, you can't use the `User` as a whole anymore). In general, it's nice to make accessor functions take and return references, like
```rust
pub fn server_registration(&self) -> &ServerRegistration { ... }
```
so that you don't destroy the object.

However, with the current use, the only time you use this function, you actually do need to take ownership of the `ServerRegistration`, so this is fine. If we generate use cases later where we don't want to consume the user, we might look into renaming this `into_server_registration`, [based on the conversion naming guidelines](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv).

marsella (commented):
This looks good! A few comments:

Re: your question about cloning. I agree with what @leboiko said, but I also spotted a few places in the code where we're cloning without needing to, e.g. it would be fine to pass ownership of an object to a function (and we don't use that object later, so there's no need to clone). I've tagged those places; I think we should remove clones that are unnecessary (e.g. where the compiler doesn't complain if it's not there), but just not be too stressed about cloning when the compiler asks for it (e.g. in `server/src/command.rs`).

What's the behavior of the database when you insert two users with the same `user_id`? It looks like this constraint is currently enforced at registration by checking whether a user_id already exists before inserting. Do you think this should also fail at `create_user` and be a constraint on the database as a whole ("there can be no more than one record with a given user_id") instead of just on the calling code? I'm not sure what the right/conventional abstraction boundaries are.

dams-key-server/src/bin/main.rs
```diff
@@ -13,13 +14,20 @@ pub enum Cli {
...
```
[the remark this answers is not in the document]
vidyaap: I was getting the error `match arms have incompatible types` when I did the match where the `Ok(db)` arm gave the DB and the `Err(_)` arm threw the tracing `error!`. Is there something else I could put there that would make the match work?

dams-tests/tests/main.rs
vidyaap: Left this one in because I got the compiler error for moving it without the clone. I think it's because I use the same db for teardown later?

dams/src/user.rs
vidyaap: Yeah I didn't love this when I wrote it, but I wasn't sure how else to get the return types to work. Is there an issue or change to make here for now though?

dams-key-server/src/command/authenticate.rs
```diff
@@ -1,21 +1,26 @@
...
 };
+use mongodb::Database;
 use opaque_ke::{ServerLogin, ServerLoginStartParameters};
 use rand::rngs::StdRng;
 use transport::server::{Chan, SessionKey};
 
 pub struct Authenticate;
 
 impl Authenticate {
+    #[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
vidyaap: Just to make sure, are you saying that there is a chance we can remove this “too many arguments” tag if we analyze the arguments and see which ones actually need to stay?   

Also I think this issue may become moot if we are switching from dialectic completely.  Should I leave the code as is for now?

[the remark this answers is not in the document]
marsella: Yes, sorry, this is all to say that maybe we don't need all the arguments and could get rid of the clippy tag.

Given the upcoming server switch, I think ignoring this for now is fine. I think you're right that this may get refactored away in that update.

dams/src/user.rs
marsella: If you also don't like it, I would keep the signature but rename it to `into_server_registration` - destroying the `User` means it's not really an accessor method and shouldn't be named like one.

marsella:
Does this also fix #91?

vidyaap:
> Does this also fix #91?

Oh, yes it does! Probably that task should have just been part of #92 instead of its own task.

marsella:
> > Does this also fix #91?
> 
> Oh, yes it does! Probably that task should have just been part of #92 instead of its own task.

Woo! Can you edit the issue description to use closing keywords for all three of the issues? Github is a little finicky, you have to use one closing keyword per issue to get them to auto-close when the PR is merged:
> closes #83, closes #92, closes #91

(or manually add them in the "development" widget on the sidebar)

dams-key-server/src/bin/main.rs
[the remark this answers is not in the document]
marsella: Oh, I see. I misread the tracing `error!` as a panic. In that case, you could replace the conditional with an `if let` like in L27 to avoid the unwrap, but that's mostly a style preference.

marsella (approved):
Thanks for making these changes! This looks good to me.

PR #96, opened 2022-07-25
fix broken links in README

### LINKED ISSUE #94: link to specification in readme
We should link to the [specification](https://github.com/boltlabs-inc/key-mgmt-spec) in the README.

jakinyele:
closes #94 as well


vidyaap (approved):
Thanks for fixing the links; looks good to me!

jakinyele:
> Thanks for fixing the links; looks good to me! But fyi, it strangely looks like Github took the links changing as a conflict instead of a change.

Oh, right. I think it's probably b/c I branched from `main` instead of `develop`. I used relative path links to avoid any issues between applying these fixes to `main` vs `develop`.

PR #103, opened 2022-08-01
Audit log infra

### LINKED ISSUE #31: Create a structured `log` type
The client components of the system have an option to retrieve the audit log from the system, which will aggregate logs from all the key servers and back-end components. The log should prioritize returning content over returning carefully structured content - that is, if one of the components has a slightly corrupted log file, users should still get access to it.

However, the log should be more structured than the current implementation, which is an arbitrary `String` (although, it should definitely implement `Display`).

Some of the relevant fields in an uncorrupted log entry include
- who generated this log
- when (date and time)
- what key is it for
- what action was requested
- who initiated the action
- what's the outcome of the action + context
I'm imagining the set of log actions is larger than the set of API requests. For example, the `request_transaction_signature` request might have logs like `request received`, `policy engine consulted`, `policy engine response received`, `distributed computation completed`, `output sent`.

This issue is complete when
- [ ] There's an implementation of logs with access to (at least) the above fields in an uncorrupted log, and access to a raw `String` in a corrupted log
- [ ] The API functions that involve audit logs have been updated to handle these
- [ ] The log structure is well documented for external parties

vidyaap:
Closes #31 

Changes:
- Create a basic `LogEntry` type
- Add logs for registration and authentication flows (failures and successes)
- Add DB functions for creating new logs

### DIFF
--- a/dams-key-server/src/command/authenticate.rs
+++ b/dams-key-server/src/command/authenticate.rs
@@ -1,6 +1,8 @@
 use crate::{database::user::find_user, error::DamsServerError, server::Context};
 
+use crate::error::LogExt;
 use dams::{
+    audit_log::Action,
     channel::ServerChannel,
     config::opaque::OpaqueCipherSuite,
     opaque_storage::create_or_retrieve_server_key_opaque,
@@ -8,7 +10,7 @@ use dams::{
         authenticate::{client, server},
         Message, MessageStream,
     },
-    user::UserId,
+    user::{AccountName, UserId},
 };
 use opaque_ke::{ServerLogin, ServerLoginStartParameters, ServerLoginStartResult};
 use tokio_stream::wrappers::ReceiverStream;
@@ -26,9 +28,13 @@ impl Authenticate {
         let (mut channel, rx) = ServerChannel::create(request.into_inner());
 
         let _ = tokio::spawn(async move {
-            let (login_start_result, user_id) = authenticate_start(&mut channel, &context).await?;
+            let (login_start_result, user_id, account_name) =
+                authenticate_start(&mut channel, &context).await?;
             authenticate_finish(&mut channel, login_start_result).await?;
-            send_user_id(&mut channel, user_id).await?;
+            send_user_id(&mut channel, user_id)
+                .await
+                .log(&context.db, &account_name, None, Action::Authenticate)
+                .await?;
 
             Ok::<(), DamsServerError>(())
         });
@@ -42,7 +48,14 @@ impl Authenticate {
 async fn authenticate_start(
     channel: &mut ServerChannel,
     context: &Context,
-) -> Result<(ServerLoginStartResult<OpaqueCipherSuite>, UserId), DamsServerError> {
+) -> Result<
+    (
+        ServerLoginStartResult<OpaqueCipherSuite>,
+        UserId,
+        AccountName,
+    ),
+    DamsServerError,
+> {
     // Receive start message from client
     let start_message: client::AuthenticateStart = channel.receive().await?;
 
@@ -79,7 +92,11 @@ async fn authenticate_start(
     // Send response to client
     channel.send(reply).await?;
 
-    Ok((server_login_start_result, user_id))
+    Ok((
+        server_login_start_result,
+        user_id,
+        start_message.account_name,
+    ))
 }
 
 async fn authenticate_finish(
--- a/dams-key-server/src/command/authenticate.rs
+++ b/dams-key-server/src/command/authenticate.rs
@@ -16,6 +16,12 @@ use opaque_ke::{ServerLogin, ServerLoginStartParameters, ServerLoginStartResult}
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::{Request, Response};
 
+struct AuthenticateStartResult {
+    login_start_result: ServerLoginStartResult<OpaqueCipherSuite>,
+    user_id: UserId,
+    account_name: AccountName,
+}
+
 #[derive(Debug)]
 pub struct Authenticate;
 
@@ -28,8 +34,11 @@ impl Authenticate {
         let (mut channel, rx) = ServerChannel::create(request.into_inner());
 
         let _ = tokio::spawn(async move {
-            let (login_start_result, user_id, account_name) =
-                authenticate_start(&mut channel, &context).await?;
+            let AuthenticateStartResult {
+                login_start_result,
+                user_id,
+                account_name,
+            } = authenticate_start(&mut channel, &context).await?;
             authenticate_finish(&mut channel, login_start_result).await?;
             send_user_id(&mut channel, user_id)
                 .await
@@ -48,14 +57,7 @@ impl Authenticate {
 async fn authenticate_start(
     channel: &mut ServerChannel,
     context: &Context,
-) -> Result<
-    (
-        ServerLoginStartResult<OpaqueCipherSuite>,
-        UserId,
-        AccountName,
-    ),
-    DamsServerError,
-> {
+) -> Result<AuthenticateStartResult, DamsServerError> {
     // Receive start message from client
     let start_message: client::AuthenticateStart = channel.receive().await?;
 
@@ -92,11 +94,11 @@ async fn authenticate_start(
     // Send response to client
     channel.send(reply).await?;
 
-    Ok((
-        server_login_start_result,
+    Ok(AuthenticateStartResult {
+        login_start_result: server_login_start_result,
         user_id,
-        start_message.account_name,
-    ))
+        account_name: start_message.account_name,
+    })
 }
 
 async fn authenticate_finish(
--- a/dams-key-server/src/command/authenticate.rs
+++ b/dams-key-server/src/command/authenticate.rs
@@ -42,12 +42,7 @@ impl Authenticate {
             authenticate_finish(&mut channel, login_start_result).await?;
             send_user_id(&mut channel, user_id)
                 .await
-                .log(
-                    &context.db,
-                    &account_name.into(),
-                    None,
-                    ClientAction::Authenticate,
-                )
+                .log(&context.db, &account_name, None, ClientAction::Authenticate)
                 .await?;
 
             Ok::<(), DamsServerError>(())
--- a/dams-key-server/src/command/authenticate.rs
+++ b/dams-key-server/src/command/authenticate.rs
@@ -2,7 +2,6 @@ use crate::{database::user::find_user, error::DamsServerError, server::Context};
 
 use crate::error::LogExt;
 use dams::{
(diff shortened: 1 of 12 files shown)


dams/src/audit_log.rs
```diff
@@ -0,0 +1,62 @@
...
+    Register,
+    Login,
+}
+
+/// Options for the outcome of a given action in a [`LogEntry`]
+#[derive(Debug, Serialize, Deserialize)]
+pub enum Outcome {
+    Successful,
+    Failed,
+}
```
picklenerd, line 24: I think this is good for now but we may want to tie these to the actions in the future. I can see this type becoming very large if there are a range of different outcomes for different actions. It would be easy to make mistakes if that ends up happening.

Just wanted to make a note of this. As long as we keep a consistent format for the database representation of each entry, this should be easy to refactor down the line. For now, I think what you've got is the best way to do this.

dams-key-server/src/database/log.rs
```diff
@@ -0,0 +1,28 @@
...
+
+/// Create a new [`LogEntry`] for the given actor, action, and outcome
+pub async fn create_log_entry(
+    db: &Database,
+    actor: &UserId,
+    secret_id: Option<ObjectId>,
+    action: Action,
+    outcome: Outcome,
+) -> Result<Option<ObjectId>, Error> {
+    let collection = db.collection::<LogEntry>("log_entries");
```
marsella, line 24: question: I see we use this string in tests as well. Will it be used in other places, too? E.g. in a normal (non-test) run, where is `db.create_collection()` called? It's may be worth making it a named variable that other methods & modules can use, if that's the case.

marsella (changes requested):
This looks good! 

I think it would be easier to read if we separate the protocol logic from the log entries on the server side (in authenticate and register). This would look like making a second function that has the original body of the `run` function, then making the `run` function call that and create the appropriate log entry based on whether it succeeded or failed. 

The infrastructure looks good though, and I don't have concerns beyond readability.

dams-key-server/src/database/log.rs
vidyaap: Yes, I meant to leave a question about this. I'd like to make some sort of config to hold all our table names so we can create them all in one shot in the tests. We shouldn't need to call `create_collection` in the actual code because Mongo will take care of creating collections that don't exist - it's just for some reason there seems to be a timing issue in the test setting.

dams/src/audit_log.rs
vidyaap: Yeah, I think you're right that we will eventually need to refactor the outcome-action setup, once we know what we can report back to the user.  For context, I think we still haven't decided the level of granularity we can return for the outcome of certain actions (e.g. for OPAUQE processes, I believe all types of failures need to appear the same to the user). @marsella, does that sounds right?

dams-key-server/src/database/log.rs
```diff
@@ -0,0 +1,28 @@
...
+};
+
+/// Create a new [`LogEntry`] for the given actor, action, and outcome
+pub async fn create_log_entry(
+    db: &Database,
+    actor: &UserId,
+    secret_id: Option<ObjectId>,
+    action: Action,
+    outcome: Outcome,
+) -> Result<Option<ObjectId>, Error> {
```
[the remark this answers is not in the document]
vidyaap: I suppose there isn't, I think I just wanted consistency across the `create` functions for different models. If that's not worth considering then I'm happy to switch the return type!

vidyaap:
Thanks for the review! I did want to ask if we should include the `FromStr` implementation that we talked about in this PR (or whatever will take care of returning a string even if we can't parse the `LogEntry` type out of whatever the server returns)?

marsella:
I think it's okay to not include the `FromStr` implementation we discussed. I was imagining that being useful in a context where the servers are writing logs to a file, and to retrieve them we have to read strings from the file and convert to log entries. In this version, we're storing them directly as log entries, so it doesn't seem like an immediate issue. We can revisit in a later phase when we receive relevant content from external parties (e.g. the policy engine).

dams/src/audit_log.rs
marsella: Yeah, for authentication we only want to report a single failure type. You're both correct that we will want to reveal more information for future actions (e.g. if the policy engine rejects a signature later, we want to know which asset fiduciaries rejected it, and maybe why). But without those behaviors spec'd out, I don't know whether it would be best to add a context string to the existing `Failure` vs add more variants to express wider range of outcomes vs some other option.

vidyaap:
This branch was really outdated so I just rebased it and changed the logging accordingly. I think it looks more readable now, but I'd like feedback on:
- The `LogExt` trait I added based on [this post](https://stackoverflow.com/questions/63489847/how-do-i-log-to-a-file-when-an-error-occurs-while-using-the-question-mark-op). I wanted to keep the `?` structure of our helpers, but is this actually cleaner or is it more confusing?
- Is it a good idea to include the `DamsServerError` that led to the failure as context in the "failed" log entries?
- What should the actor field in the logs be? It used to be `UserId` and I changed it to be `AccountName` but will we always have access to an account name in flows besides registration and auth?
- Should I maybe move `ClientAction` to the dams crate and use that instead of the `Action` type defined in the audit_log mod of dams?

Thanks!

dams-key-server/src/command/authenticate.rs
```diff
@@ -42,7 +48,14 @@ impl Authenticate {
 async fn authenticate_start(
     channel: &mut ServerChannel,
     context: &Context,
-) -> Result<(ServerLoginStartResult<OpaqueCipherSuite>, UserId), DamsServerError> {
+) -> Result<
+    (
+        ServerLoginStartResult<OpaqueCipherSuite>,
+        UserId,
+        AccountName,
+    ),
```
picklenerd, line 56: I think this needs its own private struct.

dams-key-server/src/error.rs
```diff
@@ -1,3 +1,12 @@
+use crate::database::log as Log;
```
picklenerd, line 1:
```suggestion
use crate::database::log;
```

This is more of a Javascript pattern. We shouldn't use it in Rust code.

dams-key-server/src/command/authenticate.rs
```diff
@@ -26,9 +28,13 @@ impl Authenticate {
...
 
         let _ = tokio::spawn(async move {
-            let (login_start_result, user_id) = authenticate_start(&mut channel, &context).await?;
+            let (login_start_result, user_id, account_name) =
+                authenticate_start(&mut channel, &context).await?;
             authenticate_finish(&mut channel, login_start_result).await?;
-            send_user_id(&mut channel, user_id).await?;
+            send_user_id(&mut channel, user_id)
+                .await
+                .log(&context.db, &account_name, None, Action::Authenticate)
```
picklenerd, line 36: Addressing your general comment, I really like this. It's a little noisy but I think the alternative is much noisier and less ergonomic. We might be able to come up with a clever way to reduce the number of arguments and then this would be perfect IMO.

picklenerd:
> Is it a good idea to include the DamsServerError that led to the failure as context in the "failed" log entries?

I think it depends on the purpose of these logs. If they're just used for auditing user actions, I don't think it's necessary. The error message would still be logged as part of the normal `tracing` based logging.

picklenerd:
> What should the actor field in the logs be? It used to be UserId and I changed it to be AccountName but will we always have access to an account name in flows besides registration and auth?

I think UserId would be better so that people's email addresses aren't floating around.

picklenerd:
> Should I maybe move ClientAction to the dams crate and use that instead of the Action type defined in the audit_log mod of dams?

Yeah I like this idea.

dams-key-server/src/error.rs
vidyaap: Okay! I started this for `user/User` so there wouldn't be a name conflict with variables, and I kept the pattern for this too. Would changing it to just `log` conflict with the `log()` function though?

vidyaap:
> I think UserId would be better so that people's email addresses aren't floating around.

That makes sense, but the only thing with using UserIds is that in the registration function, we don't have the UserId until after register_finish, so I'm not sure how I'd pass it to the log function.

dams-key-server/src/error.rs
picklenerd: I don't think it would. I believe the compiler will consider the context before reporting it as a conflict.

If we do end up with conflicts, we can do something like `use crate::database::log as logging` or just rename the module. I just want to make sure we stick to Rust's naming conventions.

picklenerd:
> That makes sense, but the only thing with using UserIds is that in the registration function, we don't have the UserId until after register_finish, so I'm not sure how I'd pass it to the log function.

That's a good point. Maybe we could have a trait for it and implement it for both `UserId` and `AccountName` so that either one could be passed as a parameter.

marsella (commented):
Re-reviewing just to remove my "request changes" blocker.

dams-key-server/src/database/log.rs
```diff
@@ -0,0 +1,27 @@
...
+    crypto::KeyId,
+    user::LogIdentifier,
+    ClientAction,
+};
+use mongodb::Database;
+
+/// Create a new [`LogEntry`] for the given actor, action, and outcome
+pub async fn create_log_entry(
+    db: &Database,
+    actor: &LogIdentifier,
```
picklenerd, line 18:
```suggestion
    actor: impl Into<LogIdentifier>,
```

This would let us pass user ids and account id's directly.

dams-key-server/src/error.rs
```diff
@@ -48,3 +53,32 @@ impl From<DamsServerError> for Status {
...
+        action: ClientAction,
+    ) -> Self;
+}
+
+#[async_trait]
+impl<T: std::marker::Send> LogExt for Result<T, DamsServerError> {
+    async fn log(
+        self,
+        db: &Database,
+        actor: &LogIdentifier,
```
picklenerd, line 73:
```suggestion
        actor:  impl Into<LogIdentifier>,
```

dams/src/user.rs
```diff
@@ -116,3 +116,20 @@ impl User {
...
+    fn from(user_id: UserId) -> Self {
+        Self(user_id.0)
+    }
+}
+
+impl From<AccountName> for LogIdentifier {
+    fn from(account_name: AccountName) -> Self {
+        Self(account_name.0)
+    }
+}
```
picklenerd, line 135: I know I said we shouldn't do this before, but we could also implement conversions for `&UserId` and `&AccountName` that automatically clone. This combined with the `impl Into<LogIdentifier>` changes I suggested above would let us pass the id or account name directly into the `log` function.

picklenerd (approved):
LGTM

PR #106, opened 2022-08-03
Added wrap_comments and import_granularity to rustfmt

picklenerd:
Issue: #42 

These are unstable features so in order to apply them you have to use `cargo +nightly fmt`.

These new features can be automatically applied by VS Code by adding `+nightly` to the `Rustfmt: Extra Args` setting or directly adding `"rust-analyzer.rustfmt.extraArgs": ["+nightly"]` to your `settings.json`.

One downside to this change is that running `cargo fmt` in your terminal will spam a bunch of the following
```
Warning: can't set `wrap_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `imports_granularity = Crate`, unstable features are only available in nightly channel.
```

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,13 +25,20 @@ jobs:
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
 
+      # Install stable toolchain
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
-          components: rustfmt, clippy
+          components: clippy
+
+      # Install nightly toolchain
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          components: rustfmt
 
       - name: Format check
-        run: cargo fmt --all -- --check
+        run: cargo +nightly fmt --all -- --check
 
       - name: Clippy lint
         run: cargo clippy --all-features --all-targets --verbose --workspace -- -Dwarnings
--- a/dams-key-server/src/database.rs
+++ b/dams-key-server/src/database.rs
@@ -8,7 +8,8 @@ use std::env;
 
 pub(crate) mod user;
 
-/// Connect to the MongoDB instance and database specified by your environment variables
+/// Connect to the MongoDB instance and database specified by your environment
+/// variables
 pub async fn connect_to_mongo() -> Result<Database, anyhow::Error> {
     let mongodb_url = env::var("MONGODB_URI")?;
     let db_name = env::var("DB_NAME")?;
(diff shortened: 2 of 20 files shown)


marsella (commented):
Nice, thanks. Two questions:

1. Do you have any thoughts on whether it would be wise to find a way to avoid those warnings? How often does nightly break `fmt`? Would it be reasonable to alias `cargo fmt` to `cargo +nightly fmt` by default?

2. Will this automatically get pulled into CI? I can't tell from the `rust.yml` whether it looks at the `rustfmt` file, but the CI doesn't use nightly for `fmt`. As an aside, I realized that when we added the develop branch we didn't add it to the CI, which is why tests aren't running for any of these pulls. Added #112 to address this.

picklenerd:
1. I couldn't find a way to prevent the warnings when I looked. I haven't seen nightly break cargo tools in a very long time though so this is probably safe. I made an alias to try it out but it doesn't like the `+nightly` flag so I'll need to investigate a little more.

2. I'm a little wary of adding this to CI just because the features are unstable and the comment wrapping could break explicit formatting like you mentioned in the ticket. I could add it in a separate commit so it's easy to find and revert down the line if we need to. CI should definitely respect the formatting config so this should work as long as we update the CI config to use nightly for `fmt`.

picklenerd:
This might need a separate CI definition that runs on nightly since the Rust action explicitly chooses `stable`.

vidyaap:
> 2. At this point I don't have a strong preference for whether it's in or out of the CI...

As long as the format check doesn't fail because of the difference between the unstable features and the regular version, no strong preference from me!

picklenerd:
> > 2. At this point I don't have a strong preference for whether it's in or out of the CI...
> 
> As long as the format check doesn't fail because of the difference between the unstable features and the regular version, no strong preference from me!

I don't think it will. I do both of these things manually most of the time and they never get reformatted with the default config.

marsella (approved):
This looks good to me! Before closing, please log your decision about whether you update the CI to use nightly or not so it's easy to reference.

picklenerd:
I'm going to add it to CI but I need to play around with the CI config first to figure out the best way to run the step on nightly. It looks like I don't have access to the settings for the repo so I think I'm blocked by #112.

jakinyele:
> I'm going to add it to CI but I need to play around with the CI config first to figure out the best way to run the step on nightly. It looks like I don't have access to the settings for the repo so I think I'm blocked by #112.

Just gave you the privileges - please try again!

picklenerd:
Ok I tried a couple things and pushed an update that works in CI. @marsella could you double check that you're ok with the changes I made to the CI config before I merge?

marsella:
That looks fine to me, but I'd like to see the output. Could you take on #112 (here or in a separate PR) so we can confirm that it runs the way we expect it to?

picklenerd:
@marsella Updated with #112. You can see `Run cargo +nightly fmt --all -- --check` in CI now.

marsella:
Beautiful, thank you!

Issue #109, opened 2022-08-03
Add generated integration test files to gitignore


picklenerd:
When integration tests fail, the `dams-tests/tests/gen` directory is not deleted. There are good reasons to leave these files around but they should be added to `.gitignore` to prevent them from showing up in git or accidentally being comitted.

picklenerd:
Closed by https://github.com/boltlabs-inc/key-mgmt/pull/110

PR #110, opened 2022-08-03
Add integration test files to gitignore #109

picklenerd:
When integration tests fail, the dams-tests/tests/gen directory is not deleted. There are good reasons to leave these files around but they should be added to .gitignore to prevent them from showing up in git or accidentally being comitted.


marsella:
This is probably a good thing to commit anyway, but the teardown function should be called regardless of whether the tests fail. I see that the logic is just not what I wanted in there (`dams_tests/tests/main.rs`), it's currently
```rust
    if !errors.is_empty() {
        panic!("Test failed: {:?}", errors);
    } else {
        common::teardown(server_future, db).await;
    }
```
but it should be
```rust
    common::teardown(server_future, db).await;
    if !errors.is_empty() {
        panic!("Test failed: {:?}", errors);
    }
```

Would you mind adding this fix to the PR?

picklenerd:
This was my original plan but I think it would be useful to leave the database around if the tests fail. There's also a file in this directory called `errors.log` that sounds like it might be useful.

On the flipside, anything you could do to troubleshoot the database could be done in the tests themselves during a re-rerun so maybe it's not that useful.

marsella:
That's fair. I think my design goal for the test suite was that any error would have enough context to debug without needing to manually query the database. For example, if there are any unwanted errors in `errors.log`, they're saved in the `TestError` that gets raised. I was also concerned about "previous-run" database content messing up "next-run" tests (e.g. I think user IDs (?) are derived from the test name, so we'd get "user already exists" errors on an otherwise-correct test). The teardown function also gracefully shuts down the server, which seems important.

And yes, to your point, if the tests currently fail to log useful debugging information and we discover that mid-bug, we should just add it to the suite for our future convenience.

A midway option could add an optional flag to the teardown function that keeps the database around for manual testing (e.g. keep it around if the tests failed, otherwise drop it). I'd prefer to always tear down completely. But I'll let you make the final call.

picklenerd:
I actually looked into adding a flag but it seems like `cargo test` explicitly rejects arbitrary tags. Normally I'd append a `--` and it would work but it fails here, maybe because `test` has its own args that go after `--`. I tried with a second `--` but that rust filters out all of the tests.

I'll update to do the full teardown. It's easy enough to comment out that line if you need to troubleshoot.

Issue #108, opened 2022-08-03
Implement secure handling and deletion for keys revealed client-side


marsella:
In the course of the arbitrary secrets proof of concept, the user will 
1. derive a `MasterKey` from their authentication materials
2. generate an arbitrary `StorageKey`; later, they will retrieve it from the server

Within a given workflow, the user typically uses these objects only once. When the user is done with these key types, they must securely delete them from memory. If possible, these types shouldn't be written to disk at all. 

- [ ] Do some research into secure deletion methods (overwrite with 0s or find a better way). Document selection in the comments for this issue.
- [ ] Implement `Drop` for these two types using the chosen method
- [ ] Do some research into programmatic methods for preventing objects of this type being written to disk (not sure if these exist, but if they do we should consider implementing them)
- [ ] Investigate the `!Send` trait to see if it can prevent types from being sent over the network. [See tracking issue #68318](https://github.com/rust-lang/rust/issues/68318).  Document that we're trying to block network sending, not thread passing, in case this proves to be too restrictive in the future.

marsella:
Review how ZCash does this?

indomitableSwan:
> Review how ZCash does this?

I believe some of the NCC/Trail of Bits audits touch on best practices here as well. I can dig up references for you if that ends up being helpful. I also have some internal notes on this that I can share.

jakinyele:
And [zeroize crate](https://docs.rs/zeroize/1.1.1/zeroize/) is a popular library for secure deletion recommended by auditors like NCC/ToB/Kudelski. 

Check out this short blog post for usage - https://benma.github.io/2020/10/16/rust-zeroize-move.html.

indomitableSwan:
> Within a given workflow, the user typically uses these objects only once. When the user is done with these key types, they must securely delete them from memory. If possible, these types shouldn't be written to disk at all.

This assumption may not be true in the future, particularly if we keep sessions "open" for a period of time and allow the user to make multiple requests over the course of the open session.

indomitableSwan:
@marsella I suggest we remove this from the PoC phase. I think this would be OK to include in a future phase.

PR #151, opened 2022-08-15
Add encryption abstraction

### LINKED ISSUE #113: Define encryption abstraction
The cryptography module in the `dams` crate includes encryption and decryption of a variety of key types under an AEAD encryption scheme (see #107). A good implementation should satisfy several properties:

1. Encryption is only defined in one place
2. Decryption is only defined in one place
3. If a type `T` is encrypted, the resulting blob can't be decrypted into any other type.
4. We can specify what type an encrypted blob decrypts to.

1 and 2 aim to avoid duplicated code and mistakes. 3 and 4 are to increase misuse resistance, so protocols can't accidentally decrypt things to the wrong types or mis-parse messages. For the arbitrary secrets PoC, this should support the following encryptions:
- a `StorageKey` is encrypted under a `MasterKey`
- a `Secret` is encrypted under a `StorageKey`

I think a good option is to use `PhantomData` and some trait bound to make sure that encryptable types can be serialized or turned to bytes.

- [x] Define an abstraction. Write the encryption and decryption signatures
- [x] Update (tagged) function signatures that involve encrypted types
- [x] Add tests that enc/dec aren't implemented

marsella:
Closes #113.

This adds some more (mostly private) infrastructure for implementing encryption. I've started to add some implementation-detail scaffolding to make sure I understand my type signatures correctly (e.g. `AssociatedData`). I updated all the function signatures that are supposed to take or return encrypted things. I also added the underlying encrypt/decrypt functions and the generic AEAD key type (existing keys, like `MasterKey` and `StorageKey`, will be wrappers around this one).

I've decided to punt all the `!Send` security-based decision making to #108. 

I have a few dev questions inline. Would also like a cryptography review from @indomitableSwan.


dams/src/crypto.rs
```diff
@@ -4,14 +4,86 @@
...
+    ///
+    /// Raises a [`CryptoError::DecryptionFailed`] if decryption fails or if the
+    /// associated data fails to authenticate.
+    fn decrypt(self, aead_key: &AeadKey) -> Result<T, CryptoError> {
+        todo!()
+    }
+}
+
+impl Encrypted<Secret> {
+    pub fn decrypt_secret(self, _storage_key: StorageKey) -> Secret {
```
marsella, line 82: Dev question: compare this to the `decrypt_storage_key` function in `client.rs` (you probably can't see it in this PR). Is one of these better? This is a struct method, the other one is just a floating function. Now that I've written this one out, I slightly prefer it, if only because for the other functions I've been trying to write chain-able methods.

E.g. we can already do 
```rust
opaque_export_key
    .derive_master_key()
    .encrypt_storage_key()
```
and it might be nice to maintain that style rather than calling a separate function.
```rust
let encrypted_storage_key = opaque_export_key
    .derive_master_key()
    .encrypt_storage_key();

decrypt_storage_key(export_key, encrypted_storage_key);
```
These things won't be called in line like this outside of tests, though.

dams/src/crypto.rs
```diff
@@ -4,14 +4,86 @@
...
+/// encryption scheme) authenticated encryption with associated data (AEAD)
+/// scheme.
+#[derive(Debug, Clone)]
+struct AeadKey;
+
+#[allow(unused)]
+impl<T> Encrypted<T>
+where
+    T: From<Bytes>,
+    Bytes: From<T>,
```
marsella, line 65: Dev question: These trait bounds might change depending on which encryption scheme we use and what kind of representation they require, so I put them in the impl block instead of at the struct definition. Is that reasonable? Orphan rules would prevent someone outside this crate from doing a separate `impl` block with different/no bounds on `T`, right?

Alternately, is it a better option to require `Serialize` and `Deserialize` here, derive them automatically on all the relevant types, and then do the conversion between the serde output and the AEAD input type within this implementation? My goal is to reduce errors due to inconsistent conversions around encryption and decryption.

dams/src/crypto.rs
```diff
@@ -4,14 +4,86 @@
...
+#[derive(Debug, Clone, Copy, Error)]
+pub enum CryptoError {
+    #[error("Failed to decrypt ciphertext and/or authenticate associated data")]
+    DecryptionFailed,
+}
+
+/// The associated data and additional context used in [`Encrypted`] AEAD
+/// ciphertexts and (TODO #130: HKDF) key derivations.
+#[derive(Debug, Clone)]
+struct AssociatedData(String);
```
marsella, line 31: FYI: I plan to add some methods to expand an existing associated data with additional information. E.g. you can append a string or maybe we can have fixed functions that define certain parameter choices that are used in multiple places in the code.
```rust
let context: AssociatedData = AssociatedData::default()
    .with_aead_parameters()
    .with_string("Lock Keeper storage key. ");
println!("{}", context);
```
would give

> Version 0.1: Arbitrary secrets proof of concept. AEAD parameters: (whatever) scheme with (something) key size. Lock Keeper storage key.

dams/src/crypto.rs
```diff
@@ -106,7 +174,7 @@ impl KeyId {
 ///
 /// This is generated by the client and should never be revealed to the server.
 #[derive(Debug, Deserialize, Serialize)]
-pub(crate) struct Secret {
+pub struct Secret {
```
marsella, line 177: I had to change the publicity on this to allow us to return an `Encrypted<Secret>` for use elsewhere. I think this is fine because there are no public functions to generate arbitrary secrets.

dams/src/crypto.rs
picklenerd: Yeah putting the bounds in the `impl` block is reasonable. This is the recommended pattern for generic bounds in general unless you have a really good reason to put them in the `struct` definition.

Orphan rules wouldn't prevent someone from implementing this with different bounds (or no bounds) in another crate but it looks like they'd have no way to construct an `Encrypted` since it's all private fields, so I think it's still safe.

It's probably worth adding a note to the struct's doc comments that no one should be attempting to `impl` this outside of this crate just to make it clear.

dams/src/crypto.rs
picklenerd: I like this better too. I think it's a good generalization and it centralizes encryption/decryption functionality a bit more.

dams/src/crypto.rs
```diff
@@ -4,14 +4,86 @@
...
 
 pub mod client;
 
+/// Errors that arise in the cryptography module.
+///
+/// Implementation note: this is not necessarily exhaustive yet - more variants
+/// may be added as the module is implemented.
+#[derive(Debug, Clone, Copy, Error)]
+pub enum CryptoError {
+    #[error("Failed to decrypt ciphertext and/or authenticate associated data")]
```
indomitableSwan, line 24: Nit: Given how AEADs work, I would just say "AEAD decryption failed" here, or if you are hiding that we are using an AEAD, just "decryption failed".

dams/src/crypto.rs
```diff
@@ -4,14 +4,86 @@
...
+    DecryptionFailed,
+}
+
+/// The associated data and additional context used in [`Encrypted`] AEAD
```
indomitableSwan, line 28: Nit: Technically everything is associated data for the AEAD here, so I would delete "and additional context".

dams/src/crypto.rs
indomitableSwan: This looks good.

dams/src/crypto.rs
```diff
@@ -4,14 +4,86 @@
...
+/// ciphertexts and (TODO #130: HKDF) key derivations.
+#[derive(Debug, Clone)]
+struct AssociatedData(String);
+
+impl Default for AssociatedData {
+    fn default() -> Self {
+        Self(String::from(
+            "Version 0.1: Arbitrary secrets proof of concept. ",
```
indomitableSwan, line 36: Nit: I might be tempted to just use, e.g., alphanumeric naming or numbering without additional context for versioning.

dams/src/crypto.rs
```diff
@@ -4,14 +4,86 @@
...
+        ))
+    }
+}
+
+/// A ciphertext representing an object of type `T`, encrypted under (TODO #107:
```
indomitableSwan, line 41: Add "an AEAD" before the TODO note.

dams/src/crypto.rs
```diff
@@ -4,14 +4,86 @@
...
+/// encryption scheme).
+///
+/// Implementation note: there may be additional fields in this struct (e.g. an
+/// authentication blob separate from the ciphertext), or the types might
```
indomitableSwan, line 45: What do you mean by "authentication blob" here?

dams/src/crypto.rs
```diff
@@ -4,14 +4,86 @@
...
+/// change.
+#[allow(unused)]
+#[derive(Debug, Clone)]
+pub struct Encrypted<T> {
+    ciphertext: Bytes,
+    associated_data: AssociatedData,
+    original_type: PhantomData<T>,
+}
+
+/// A well-formed symmetric encryption key for the (TODO #107: specify
```
indomitableSwan, line 55: Add "an AEAD" before the TODO note here.

dams/src/crypto.rs
```diff
@@ -4,14 +4,86 @@
...
+/// encryption scheme) authenticated encryption with associated data (AEAD)
+/// scheme.
+#[derive(Debug, Clone)]
+struct AeadKey;
```
indomitableSwan, line 59: Nit: Consider using simpler naming here, like `SymKey` or `EncKey` (a little weird, admittedly).

dams/src/crypto.rs
indomitableSwan: @picklenerd, can I poke you to expand on your above answer and comment on the serde alternative @marsella described?

dams/src/crypto.rs
```diff
@@ -4,14 +4,86 @@
...
+{
+    /// Encrypt the `T` under the [`AeadKey`] with the [`AssociatedData`].
+    fn encrypt(object: T, associated_data: &AssociatedData, aead_key: &AeadKey) -> Encrypted<T> {
+        todo!()
+    }
+
+    /// Decrypt the ciphertext to a `T`.
+    ///
+    /// Raises a [`CryptoError::DecryptionFailed`] if decryption fails or if the
+    /// associated data fails to authenticate.
```
indomitableSwan, line 75: Same nit as above - I don't think it's necessary to include all reasons _why_ decryption might fail, and typically you'd just talk about decryption failing here. (Which is to say, you don't _learn_ the reason that decryption failed here.)

dams/src/crypto.rs
indomitableSwan: I like it better, too.

dams/src/crypto.rs
indomitableSwan: Maybe at some point (with more developers onboarded) we can re-evaluate our structuring for this project, but this seems OK.

dams/src/crypto/client.rs
```diff
@@ -6,6 +6,8 @@ use crate::crypto::{OpaqueExportKey, StorageKey};
 
 use rand::{CryptoRng, RngCore};
 
+use super::{Encrypted, Secret};
+
 /// Create an encrypted storage key. This is part of the registration flow and
 /// is executed after completing the OPAQUE registration session with the
 /// server. This key should be sent to the server for storage.
```
indomitableSwan, line 13: Technically, this is called _during_ an open registration session.

indomitableSwan (changes requested):
Looks pretty good. There are a couple questions/comments inline, with a few nits that I'd like fixed.

Also: L136 in crypto.rs has a typo.

dams/src/crypto.rs
picklenerd: On `serde`, I think I need to know more about how this will be implemented. If `serde` is how you're going to get from bytes to `T` then I think adding `Serialize` and `Deserialize` bounds will be required no matter what. It might be worth breaking this up into two `impl` blocks in that case so the bounds are more appropriate for each function. I might be misinterpreting this question though.

picklenerd: To expand on my answer above:

I just double checked the orphan rule and it doesn't actually apply here since this isn't a trait `impl`. You would be blocked from writing an `impl` block here based on the normal rules for writing `impl` blocks for a `struct`.

The protection I mentioned above is just that `Encrypted` contains a bunch of private fields but no constructor. Since no one outside of this module can see the private fields, there's no way to construct an `Encrypted` from outside of this module. The only way to get one from outside would be to call `Encrypted::encrypt`.

dams/src/crypto.rs
picklenerd, line 177: Do you think we'll ever add more fields to this? Maybe this could be `struct Secret(Bytes)` to emphasize that it's a wrapper.

dams/src/crypto.rs
marsella: That's fair. We might want to come up with a versioning policy or something to determine what minor and major version bumps look like and keep a history of what changed at each version, but I'm happy to put that off at least until we change it for the first time.

dams/src/crypto.rs
marsella: I wasn't sure if the AEAD scheme would keep the ciphertext bit separate from the authenticated bit. I'll edit this comment to remove the speculative parenthetical.

dams/src/crypto.rs
marsella: I switched it from `EncryptionKey` because it felt so vague! But I'll put it back :)

dams/src/crypto.rs
marsella: Re: serde. The options I've considered are:

1. Require the to/from `Bytes`. Convert in the `encrypt` function from Bytes to whatever representation the AEAD scheme uses.
2. Require a to/from the specific representation the AEAD scheme uses (e.g. `Vec[u8]`).
3. Require `Serialize` and `Deserialize`. Convert in the `encrypt` function from whichever serde output we choose to the representation the AEAD scheme uses.

In hindsight, (2) is not a great option because switching schemes would maybe require changing the bounds to match the output of the new scheme (or adding a conversion step in `encrypt`). The benefit of (3) over (1) is serde's automatic derivation, which might help avoid programmer error.

dams/src/crypto.rs
marsella: Yes! It will also have `AssociatedData` -- now that you mention it, I'll just add that field from now.

picklenerd (approved):
This looks good to me from a code perspective.

dams/src/crypto.rs
picklenerd: One nice thing about 1 is that you can do the `serde` stuff in the `From` impl but leave the option open for other types to do something other than `serde`.

dams/src/crypto.rs
marsella: Fixed!

dams/src/crypto.rs
marsella: That's a good point. I think I'll leave it as-is for now and see if we run into anything weird once we implement it.

indomitableSwan (approved):
I am OK with this PR now, but the encryption signature doesn't match what is in the spec currently. That is, the ordering of inputs is different.

Will this discrepancy cause implementation errors? If so, we should resolve the discrepancy, either here or in the spec.

dams/src/crypto.rs
indomitableSwan: Yes, these are all feeling a bit vague. I think my discomfort here is entirely aesthetic, ie., I don't like the way Rust naming conventions play with acronyms.

dams/src/crypto.rs
indomitableSwan: Ahh. I just took a look at RustCrypto's trait, and it does look like they don't keep these separate.

marsella:
Good catch! I'll fix encryption to match. Note that the specific encryption calls (`encrypt_storage_key`, `encrypt_data`) already match the expected format of (key, value). 

Decryption doesn't match because of Rust formatting standards -- the spec says `Dec(key, ciphertext)` but we have decryption as a function on the ciphertext, so the arg order is `Dec(ciphertext, key)`. I don't think this will cause errors, though -- it feels obvious to me which argument is which :)

Issue #152, opened 2022-08-16
Refactor message handling to use generic messages


picklenerd:
The existing pattern for processing messages is cumbersome and error-prone. We went through a design and prototyping phase in #128. We have decided to use a generic message type in the protobuf definitions and define our message types directly in Rust.

This issue is complete when:

- [ ] The generic message system from [the demo project](https://github.com/boltlabs-inc/message-handling-demos) is implemented.
- [ ] Existing endpoints for registration and authentication are refactored to use this syste,

indomitableSwan:
This seems like a pretty large issue: should it perhaps be split into two?

picklenerd:
I'd prefer to keep it as a single issue since it'll be a single PR. The system itself is handled by a relatively small macro so the bulk of the work here is refactoring.

PR #154, opened 2022-08-16
Implement message handling system #152

picklenerd:
The PR implements a message handling system to reduce boilerplate and simplify protocol definitions.

This implementation differs a bit from the original demo. Instead of using a macro to generate the message handling boilerplate, I created a `Channel` type that handles type conversions and channel operations. The macro was convenient but inflexible, especially in situations with persistent state between steps.

I tried to avoid any other refactoring within the protocols to keep this PR simple, but I did move the client-side implementation logic to modules in order to reduce the size of `api.rs`.


dams-local-client/src/api.rs
```diff
@@ -80,21 +70,6 @@ impl Password {
...
 }
 
-/// Deployment details for a [`Session`].
-///
-/// Possible fields include: timeouts, key server IPs, PKI information,
-/// preshared keys.
-#[derive(Debug, Default, Clone)]
-pub struct SessionConfig {
-    client_config: Config,
-}
```
picklenerd, line 90: This was passed around but not actually used anywhere from what I could tell. I removed it to get rid of the warnings but I left the parent type alone since it looks useful.

dams/src/types/authenticate.rs
```diff
@@ -0,0 +1,39 @@
+pub mod client {
```
picklenerd, line 1: This is just a pattern I use to shorten repetitive type names. We can flatten this and rename the types to look like `ClientAuthenticateStart` if people would prefer that.

The idea here is that you import `types::client` and use `client::AuthenticateStart` to refer to the type. In theory, this would end up with shorter type names in situations where you only need stuff from `types::client` but we end up importing both `types::client` and `types::server` on both the client and server handlers.

marsella: I like this. The `::` acts like a bit of a space to me and makes it easier to read.

marsella (commented):
This looks good to me! The overhead is much reduced and I appreciate the documentation throughout. It looks like it'll be much nicer to use. I have a couple questions about how this fits into the bigger picture and a couple about design.

1. After a user authenticates, they will make a request over the same channel. In the server `Authenticate::run` function, it looks like the `channel` only exists for the lifetime of authentication. Where will relevant state (like the session key generated by OPAQUE and used for MACs) persist from authentication to a request on the server side? For the client, they do persist the session key (by saving it into a `Session`) but not the export key, which they'll need for any storing or retrieving. On the client side, do you expect the following request flows to set up a fresh `Channel` and take the existing `Session`, or will you persist the channel somehow?

1. What is your plan for differentiating between `Session`s and `Channel`s moving forward? `Channel` seems to handle direct networking details and is passed directly to the working flows; `Session` is currently only used in the external APIs and it's not clear to me that the original abstraction boundaries set for that type are still relevant. However, based on the previous question, I wonder if it makes sense on the client side to update a `Session` to hold a `Channel` along with the outputs from authentication, or if you had a different plan for the `Session` type.

3. It looks like conversions from bytes (e.g. the message received) to meaningful types all happens inline in the register and authenticate functions. Auditing whether the protocol is sending and receiving the correct message types requires looking through each protocol function and checking that the deserialize line converts to the expected type. Are there benefits to this approach over something like, having a `deserialize` function for each struct in `dams/src/types` that converts the `Vec<u8>` to the expected types?

4. I don't see any new tests with these changes. Are there unit-level tests worth making, or are you satisfied with the end-to-end tests capturing the functionality you added / updated here?

dams-key-server/src/command/authenticate.rs
```diff
@@ -1,191 +1,127 @@
...
-    server_message: ServerLoginStartResult<OpaqueCipherSuite>,
-}
-
 #[derive(Debug)]
 pub struct Authenticate;
 
-pub type AuthenticateStream = ReceiverStream<Result<ServerAuthenticate, Status>>;
-
 impl Authenticate {
     #[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
picklenerd: Yes it can!

picklenerd:
> 

1. We're going to have to figure out a place to store the state. I think we can put it in the DB for the PoC but we might consider something like Redis as a permanent solution. Redis will let us define things like how long the state should exist without us having to manage that manually.
2. I don't have a concrete plan yet but my current thought is that we remove our internal concept of sessions and rename references to it in our current code. I think we should refer to sessions as the multi-request processes bound to a single authentication, if that makes sense. Basically a session is everything that happens while the persistent state from (1) exists. I think the existing client code could use a pretty big structural refactor but I don't think it's important right now. I think you, Vidya, and I should all sync up on how we think it should look once the PoC is done.
3. In this code I just deferred to the `bincode` crate since we were already using it elsewhere. I definitely think it's ok to define these conversions explicitly in the code if that's helpful for auditing. I think using `TryFrom<Message>` instead of our own `deserialize` function is preferable since having it in trait form makes it easy to operate on generic messages, but we can change the implementation of those conversions to whatever we want and it shouldn't break anything as long as we never store raw `Message` objects in the DB.
4. I'm happy with the E2E tests right now since all of this stuff requires communication between the client and server and also requires a database. I think that might be a little too complex for unit tests. I think `Channel` could use some unit tests though. I'll add some.

vidyaap:
> I think the existing client code could use a pretty big structural refactor but I don't think it's important right now. I think you, Vidya, and I should all sync up on how we think it should look once the PoC is done.

Second-ing the sync-up request. I'm running into issues with how to translate steps like "Runs a validity check on the received request and user_id..." from the spec into code since we have the new `Channel` type, the `Session` type and the tonic client object.

marsella:
> 1. We're going to have to figure out a place to store the state. I think we can put it in the DB for the PoC but we might consider something like Redis as a permanent solution. Redis will let us define things like how long the state should exist without us having to manage that manually.

For the PoC, I was thinking something more like a state-holding variable (`Client`?) that we drop when we've completed the one request allowed per authentication. Can Redis can handle decisions like timing out after a certain number of requests as well as general timeouts? Right now there isn't any client-side DB, so we'd have to add that infrastructure to use it in the PoC.
 
> 2. I don't have a concrete plan yet but my current thought is that we remove our internal concept of sessions and rename references to it in our current code. I think we should refer to sessions as the multi-request processes bound to a single authentication, if that makes sense. Basically a session is everything that happens while the persistent state from (1) exists. I think the existing client code could use a pretty big structural refactor but I don't think it's important right now. I think you, Vidya, and I should all sync up on how we think it should look once the PoC is done.

I think we're on the same page about this :+1:

> 3. In this code I just deferred to the `bincode` crate since we were already using it elsewhere. I definitely think it's ok to define these conversions explicitly in the code if that's helpful for auditing. I think using `TryFrom<Message>` instead of our own `deserialize` function is preferable since having it in trait form makes it easy to operate on generic messages, but we can change the implementation of those conversions to whatever we want and it shouldn't break anything as long as we never store raw `Message` objects in the DB.

I'm fine with the `bincode` deserialization and the `TryFrom` approach! I was suggesting something slightly different -- for example, in the key server authenticate command, you have (I've cut out some intervening code):

```rust
let start_message: client::AuthenticateStart = channel.receive().await?;
let uid = super::user_id_from_message(&start_message.user_id)?;
```
I was wondering if it would make sense to have a function on `AuthenticateStart` that consumes it and produces a `UserId` and `CredentialRequest<OpaqueCipherSuite>` by calling the two deserialize functions above.
```rust
let start_message: client::AuthenticateStart = channel.receive().await?;
let (uid, credential_request) = start_message.into_parts()?;
```
The benefit would be that you could look at the `types` module and see the intended types of each message since each one will have a function whose return type defines the expected deserialized type.

> 4. I'm happy with the E2E tests right now since all of this stuff requires communication between the client and server and also requires a database. I think that might be a little too complex for unit tests. ~I think `Channel` could use some unit tests though. I'll add some.~ (edit: Making a `Streaming` is a big pain. I'm not sure it's worth it given that this type is just a wrapper for some other types.)

Fine by me, thanks for checking!

vidyaap:
> ```rust
> let start_message: client::AuthenticateStart = channel.receive().await?;
> let uid = super::user_id_from_message(&start_message.user_id)?;
> let credential_request: CredentialRequest<OpaqueCipherSuite> =
>         dams::deserialize_from_bytes(&start_message.message)?;
> ```

Also realizing that we probably don't need this `user_id_from_message` function anymore. We can deserialize it the way we do the other message fields.

marsella (approved):
Okay, that makes sense to me. Thanks for expanding on your thoughts. I have no other concerns here.

Approving, but please make the issue you mentioned about adding `into_parts` deserializers.

PR #156, opened 2022-08-17
Change remote-client API function names #155

### LINKED ISSUE #155: Change remote-client API function names
As per feedback, we will change the remote-client API function names to be simpler and follow more standard naming conventions:

```
register_passive_user               -> register_user
create_passive_digital_asset_key    -> create_key
request_transaction_signature       -> sign_transaction
import_asset_key                    -> import_key (or make key material an optional argument to create_key)
export_asset_key                    -> export_key
retrieve_public_keys                -> get_keys
retrieve_public_key_by_id           -> get_key_by_id
retrieve_audit_log                  -> get_log
```

vidyaap:
Closes #155

### DIFF
--- a/dams-remote-client/src/api.rs
+++ b/dams-remote-client/src/api.rs
@@ -24,7 +24,7 @@ pub enum Error {}
 ///
 /// TODO #30 (design, implementation): Pass a session.
 #[allow(unused)]
-pub fn register_passive_user(user_id: UserId) -> Result<(), Error> {
+pub fn register_user(user_id: UserId) -> Result<(), Error> {
     todo!()
 }
 
@@ -38,7 +38,7 @@ pub fn register_passive_user(user_id: UserId) -> Result<(), Error> {
 ///
 /// TODO #30 (design, implementation): Pass a session.
 #[allow(unused)]
-pub fn create_passive_digital_asset_key(
+pub fn create_key(
     user_id: UserId,
     blockchain: Blockchain,
     restriction: impl UseRestriction,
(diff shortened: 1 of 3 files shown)


dams-remote-client/src/api.rs
```diff
@@ -144,7 +144,7 @@ pub fn retrieve_public_keys(user_id: UserId) -> Result<Vec<KeyInfo>, Error> {
 ///
 /// TODO #30 (design, implementation): Pass a session.
 #[allow(unused)]
-pub fn retrieve_public_key_by_id(user_id: UserId, key_id: &KeyId) -> Result<KeyInfo, Error> {
+pub fn get_key_by_id(user_id: UserId, key_id: &KeyId) -> Result<KeyInfo, Error> {
```
indomitableSwan, line 147: Same comment:  Maybe`get_pub_key_by_id` here?

indomitableSwan (commented):
Looks good, except for I do have concerns about the semantics of `get_keys` and `get_key_by_id`, because that sounds like you might be retrieving secret key material, which isn't the case here. What are your thoughts on this?

vidyaap:
> Looks good, except for I do have concerns about the semantics of `get_keys` and `get_key_by_id`, because that sounds like you might be retrieving secret key material, which isn't the case here. What are your thoughts on this?

I agree, I like specifying that it's the public material. Added a commit for that.

indomitableSwan (approved):
LGTM. Thanks, Vidya!

Issue #153, opened 2022-08-16
Refactor error handling


picklenerd:
We're currently using several different types of error handling.

1. `anyhow` in the binaries
2. `thiserror` in crypto types and other library code
3. `tonic::Status` anywhere that interacts with `tonic`

We should clean up error handling a bit to be more consistent. 

I suggest:

- [x] Remove `anyhow`.
- [x] Define error types in all of our library code. We can do this manually or with `thiserror`.
- [x] Define conversions from our types to `tonic::Status` so they can be forwarded with `?` in `tonic` handlers.

picklenerd:
Merged https://github.com/boltlabs-inc/key-mgmt/pull/158

PR #158, opened 2022-08-18
Refactor error handling #153

picklenerd:
This PR creates custom error types for the three main crates.

The goal of these error types is to be able to use `?` everywhere instead of having to call `map_err`. This is accomplished by wrapping downstream errors  and forwarding their `std::Error` implementation with `#[error(transparent])`.

`anyhow` has been removed from the three library crates. It's still used in the `dams-tests` crate and can be used in binaries in the future. It's not meant for use in libraries.

### DIFF
--- a/dams-client/src/api/authenticate.rs
+++ b/dams-client/src/api/authenticate.rs
@@ -11,16 +11,16 @@ use opaque_ke::{
 use rand::{CryptoRng, RngCore};
 use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
-use tonic::{transport::Channel, Status};
+use tonic::transport::Channel;
 
-use super::Password;
+use crate::{api::Password, error::DamsClientError};
 
 pub(crate) async fn handle<T: CryptoRng + RngCore>(
     client: &mut DamsRpcClient<Channel>,
     rng: &mut T,
     user_id: &UserId,
     password: &Password,
-) -> Result<[u8; 64], Status> {
+) -> Result<[u8; 64], DamsClientError> {
     // Create channel to send messages to server after connection is established via
     // RPC
     let (tx, rx) = mpsc::channel(2);
@@ -32,8 +32,7 @@ pub(crate) async fn handle<T: CryptoRng + RngCore>(
     let mut channel = ClientChannel::create(tx, server_response);
 
     let client_login_start_result =
-        ClientLogin::<OpaqueCipherSuite>::start(rng, password.as_bytes())
-            .map_err(|_| Status::aborted("LoginStartFailed"))?;
+        ClientLogin::<OpaqueCipherSuite>::start(rng, password.as_bytes())?;
 
     // Handle start step
     let server_start_result =
@@ -56,14 +55,15 @@ async fn authenticate_start(
     channel: &mut ClientChannel,
     client_login_start_result: &ClientLoginStartResult<OpaqueCipherSuite>,
     user_id: &UserId,
-) -> Result<server::AuthenticateStart, Status> {
+) -> Result<server::AuthenticateStart, DamsClientError> {
     let reply = client::AuthenticateStart {
         credential_request: client_login_start_result.message.clone(),
         user_id: user_id.clone(),
     };
 
     channel.send(reply).await?;
-    channel.receive().await
+
+    Ok(channel.receive().await?)
 }
 
 async fn authenticate_finish(
@@ -72,15 +72,12 @@ async fn authenticate_finish(
     password: &Password,
     client_start_result: ClientLoginStartResult<OpaqueCipherSuite>,
     server_start_result: server::AuthenticateStart,
-) -> Result<ClientLoginFinishResult<OpaqueCipherSuite>, Status> {
-    let client_login_finish_result = client_start_result
-        .state
-        .finish(
-            password.as_bytes(),
-            server_start_result.credential_response,
-            ClientLoginFinishParameters::default(),
-        )
-        .map_err(|_| Status::unauthenticated("Authentication failed"))?;
+) -> Result<ClientLoginFinishResult<OpaqueCipherSuite>, DamsClientError> {
+    let client_login_finish_result = client_start_result.state.finish(
+        password.as_bytes(),
+        server_start_result.credential_response,
+        ClientLoginFinishParameters::default(),
+    )?;
 
     let reply = client::AuthenticateFinish {
         credential_finalization: client_login_finish_result.message.clone(),
@@ -94,6 +91,6 @@ async fn authenticate_finish(
     if server_finish.success {
         Ok(client_login_finish_result)
     } else {
-        Err(Status::internal("Server returned failure"))
+        Err(DamsClientError::ServerReturnedFailure)
     }
 }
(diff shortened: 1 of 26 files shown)


dams-client/src/api/authenticate.rs
```diff
@@ -33,8 +33,7 @@ pub(crate) async fn handle<T: CryptoRng + RngCore>(
     let mut channel = ClientChannel::create(tx, server_response);
 
     let client_login_start_result =
-        ClientLogin::<OpaqueCipherSuite>::start(rng, password.as_bytes())
-            .map_err(|_| Status::aborted("LoginStartFailed"))?;
+        ClientLogin::<OpaqueCipherSuite>::start(rng, password.as_bytes())?;
```
picklenerd, line 36: Do we need to obscure this error or is it ok to forward it?

dams/src/error.rs
```diff
@@ -0,0 +1,43 @@
...
+        let message = error.to_string();
+
+        use DamsError::*;
+        match error {
+            InvalidOpaqueDirectory
+            | ProjectDirs
+            | Bincode(_)
+            | Io(_)
+            | OpaqueProtocol(_)
+            | Toml(_) => Status::internal(message),
```
picklenerd, line 40: We can specialize this in the future if we feel like we need different types of `Status`.

dams/src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+    #[error("Could not open user's home directory")]
+    ProjectDirs,
+
+    // Wrapped errors
+    #[error(transparent)]
+    Bincode(#[from] bincode::Error),
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+    #[error("OPAQUE protocol error: {}", .0)]
+    OpaqueProtocol(opaque_ke::errors::ProtocolError),
```
picklenerd, line 25: This one doesn't implement the `std::Error` trait properly so it needs special treatment.

dams-key-server/src/command/register.rs
```diff
@@ -61,15 +61,14 @@ async fn register_start(channel: &mut ServerChannel, context: &Context) -> Resul
         .map_err(|_| Status::aborted("MongoDB error"))?
```
marsella, line 61: todo: I think you can drop this one, I see you have a `DamsServerError::MongoDb`.

dams-key-server/src/command/register.rs
```diff
@@ -21,30 +21,30 @@ impl Register {
...
     // Get server key for OPAQUE
     let server_setup = {
-        let mut rng = context
-            .rng
-            .lock()
-            .map_err(|_| Status::unavailable("Unable to access RNG"))?;
+        let mut rng = context.rng.lock().await;
 
         create_or_retrieve_server_key_opaque(&mut rng, &context.service)
             .map_err(|_| Status::aborted("could not find/create server key"))?
```
marsella, line 50: todo: Can you drop this one, too?

dams/src/error.rs
marsella: Do you mean using `Status::internal` vs `Status::aborted`? Was there a logic before on which one we chose for different errors?

dams-client/src/error.rs
```diff
@@ -0,0 +1,51 @@
...
+    #[error("Select register failed")]
+    SelectRegisterFailed,
+    #[error("Auth start failed")]
+    AuthStartFailed,
+    #[error("Register start failed")]
+    RegisterStartFailed,
+    #[error("Auth finish failed")]
+    AuthFinishFailed,
+    #[error("Register finish failed")]
+    RegisterFinishFailed,
```
marsella, line 32: question: I was going to comment that I'm not sure how many of these details we want to reveal to the client, so we should revisit this in the hardening PR to decide whether to take some out. But when I looked for the use cases, it seemed like most of these are not used anywhere. Is that correct? Can we take them out now if so? Why doesn't clippy flag them as unused?

marsella (commented):
This is a great improvement!

I still see some manual `map_err(|_| Status::aborted("..."))`s laying around. Is there a guide for when to use them and when they can be wrapped into the error types you made? I commented on a couple that looked like they could go away, but I might be missing something.

dams-key-server/src/command/authenticate.rs
```diff
@@ -66,28 +62,22 @@ async fn authenticate_start(
...
     };
     let credential_request: CredentialRequest<OpaqueCipherSuite> =
         dams::deserialize_from_bytes(&start_message.message)?;
 
     let server_login_start_result = {
-        let mut local_rng = context
-            .rng
-            .lock()
-            .map_err(|_| Status::unavailable("Unable to access RNG"))?;
+        let mut local_rng = context.rng.lock().await;
```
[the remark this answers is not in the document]
picklenerd: I replaced `std::sync::Mutex` with `tokio::sync::Mutex`. The `tokio` one is better when you're running inside of `tokio` since it's async and doesn't return a result.

dams/src/error.rs
picklenerd: There's a test here:
https://docs.rs/tonic/latest/src/tonic/status.rs.html#235

```
Use `Aborted` if the client should retry at a higher-level (e.g., restarting a read-modify-write sequence).
```

I'm not sure that really fits. I used `internal` as a general case since all of the `tonic` errors will likely be handled internally. If we want to have more granular `tonic` errors we should definitely have a discussion to figure out how we want it to work.

dams/src/error.rs
picklenerd: It looks like it was implemented normally in `1.2.0` but now it's all weird in `2.0.0-pre.3`. It's not immediately obvious how to untangle what's in there now so I'm not sure if it's worth it unless someone really wants to.

https://docs.rs/opaque-ke/2.0.0-pre.3/src/opaque_ke/errors.rs.html#166

dams-client/src/error.rs
picklenerd: That's a good point. If we need to add these back in the future, we could probably just log the details and return something generic.

picklenerd: I've removed almost all of the ones that aren't in use. The last two are `RegistrationFailed` and `AuthenticationFailed` which I imagine are ok to return.

picklenerd:
> I still see some manual map_err(|_| Status::aborted("..."))s laying around. Is there a guide for when to use them and when they can be wrapped into the error types you made? I commented on a couple that looked like they could go away, but I might be missing something.

I fixed all of the ones I missed. The ones leftover in `channel` are necessary and the ones in `crypto` seem appropriate. Everyone else is clean!

dams-client/src/api/authenticate.rs
marsella: The `ProtocolError` that the opaque library reveals is safe, I think. I added #185 to review later, but you can leave as-is for now.

dams-key-server/src/command/authenticate.rs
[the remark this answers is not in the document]
marsella: Ah, nice.

dams-client/src/error.rs
marsella: Thanks!

marsella (approved):
Thanks for the changes, this looks good to me!

Issue #164, opened 2022-08-19
Remove remote client from key-mgmt repo


picklenerd:
The remote client will be customer-specific and should not be part of the main `key-mgmt` repo. We can simplify the structure of the main repo by moving it to its own private repo.

When we implement the remote client, we can likely just wrap logic from the local client for common features and add any specialized logic on top of that. As such, the local client should just be considered the general client.

This issue is complete when:

- [x] The remote client is moved to its own private repo
- [x] The local client is renamed to `dams-client`

picklenerd:
https://github.com/boltlabs-inc/dams-remote-client

picklenerd:
PR to add the integration tests: https://github.com/boltlabs-inc/dams-remote-client/pull/1

PR #165, opened 2022-08-19
Removed remote client, renamed local client #164

picklenerd:
Moved here: https://github.com/boltlabs-inc/dams-remote-client


indomitableSwan:
I think it would be good for @vidyaap to take a look at this PR. I think there is more that needs to be removed, i.e., there there is still a "dams-remote-client" folder with the API on this branch.

picklenerd:
> I think it would be good for @vidyaap to take a look at this PR. I think there is more that needs to be removed, i.e., there there is still a "dams-remote-client" folder with the API on this branch.

Oh that's just me being forgetful. I left the folder on my computer so I could copy it to the new repo and forgot to delete it on the old repo.

Agree that @vidyaap should take a look though.

picklenerd:
> This looks good! Just a couple of doc edits left over that I could see

Thanks! I updated the second one too just to be consistent and keep this conceptually generic.

vidyaap (approved):

Issue #125, opened 2022-08-11
Implement the registration protocol


marsella:
The [registration protocol](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/cryptographic_flows.md#register) needs to be implemented according to the spec.

This will require a database insertion (#124) and message passing (#123). It will also call cryptographic functions defined in #105 and #113. At time of writing, the cryptographic functions aren't implemented yet. Whoever takes on this issue can decide whether to wait on the implementation tickets for the subset of relevant functions defined in #105. 

Note that this implementation _must_ be reviewed by a cryptographer.

- [x] Implement client side
- [x] Implement server side
- [ ] Add an integration test that a client can register without failing (maybe that they can register and then authenticate?). 
- [ ] Think about whether there are any other tests that would make sense at this point

marsella:
I think that (unlike other flows) the client-side code for this will have to go in the `Session::register` function. Note that due to the current "single-use" nature of the protocol, that function signature should change to not return anything (e.g. the user doesn't get a valid, open session any more).

Issue #124, opened 2022-08-11
Update server database to store `StorageKey`s


marsella:
In the [registration protocol](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/system-functionalities.md#register), the server needs to store a validated, encrypted storage key in a record associated with the user. 

The storage key should only be set once (during registration), and will be retrieved often.

- [ ] Update the [`dams::user::User`] model to include an encrypted storage key (blocked on #113 - this type isn't defined yet)
- [ ] Implement adding an encrypted storage key to a user in  `dams-key-server::database::database`
- [ ] Implement retrieving an encrypted storage key from a user in `dams-key-server::database::database`
- [ ] Test that storing and retrieving a key gets you the same key back.

picklenerd:
On hold while crypto gets finished

Issue #123, opened 2022-08-11
Write tonic helper functions for registration


marsella:
The registration protocol has some additional functionality that needs to be added, describe [in the spec](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/cryptographic_flows.md#register). This involves additional messages passed between the client and server.

Per comments in #98, sending messages requires defining some helper functions to enforce the protocol message ordering. The additional messages include:
1. The client sends a registration request (type not defined at time of writing), including a user id and encrypted storage key.
2. Maybe the server sends a success indicator? Or they just close the channel.

This can be closed when
- [x] Define the registration request type
- [x] Define any tonic infrastructure / helper functions needed to execute the protocol 

NB: @vidyaap or @picklenerd can review this issue text and edit for clarity if it doesn't accurately reflect how tonic works.

picklenerd:
This think this is enough detail for a ticket.

We can break down the helper functions if we want. You can see the existing ones here:
https://github.com/boltlabs-inc/key-mgmt/blob/tonic-registration/dams-key-server/src/command/register.rs#L89

Basically, we have one to ensure that the message is the right type and one to process the message and generate a response. I don't think we need to put these details in the tickets though.

Issue #101, opened 2022-07-28
Complete implementation of registration and authentication protocols


marsella:
The completion of issues #50 and #51 successfully incorporated the OPAQUE registration and authentication protocols into the `Session` type, but there are still some components missing to complete the implementation of our full registration and authentication protocols.

Some of this work overlaps with #30 and #87, in determining how the `Session` will enforce the extra encryption following authentication and how we'll maintain an open channel after those protocols.

This is currently blocking on development of the complete spec for registration and authentication (boltlabs-inc/key-mgmt-spec#10 and boltlabs-inc/key-mgmt-spec#11).

indomitableSwan:
> Authentication is defined as well, but we haven't defined what it means to open an encrypted channel secured by the OPAQUE session_key, so there's no further work to do there.

Can you clarify what you mean by this line?

indomitableSwan:
This is complete. There is some ongoing testing hardening in #216, but this task does not have to be done as part of this epic.

PR #167, opened 2022-08-22
Generate and store

### LINKED ISSUE #69: Write local-client function to implement "generate and store a secret" flow
This function will establish a channel with the server using the pattern established in #119. Once the channel is established, it will implement the "generate and store a secret" flow described [in the spec](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/cryptographic_flows.md#generate-and-store-a-secret).

### LINKED ISSUE #136: Implement the generate-and-store flow
The [generate-and-store flow](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/cryptographic_flows.md#generate-and-store-a-secret) is a multi-message protocol between client and server.
Per comments in #98, sending messages requires defining some helper functions to enforce the protocol message ordering. In generation, these are:

1. The client sends a storage request (type not defined at time of writing), including a user id.
2. The key server sends a `KeyId`
3. The client sends an encrypted `Secret`
4. The key server sends a success message

- [x] Define client and server message types and serialization into / from tonic messages
- [x] Implement client-side protocol - sends request, and encrypted secret for storage; receives new key ID
- [x] Implement server-side protocol - receives request and stores encrypted secret; sends key ID
- [ ] Integration tests

vidyaap:
Closes #136 

I'm making this draft PR to make it easier to ask questions. This is still missing a few changes from #154, so I'll have to rebase with that once it's merged before I mark this ready for review.

I had the following questions:
- A couple of questions about the crypto module that I'd like @marsella's opinion on:
    - `Secret`, `Secret::generate()`, and `StorageKey::encrypt()` all had to be made public for me to use them in the client. Is that okay or were you envisioning using these in a different way?
    - Is it okay that I added `Clone` to `KeyId`?
    - In the `retrieve_storage_key` step, I need an export key - is this supposed to be part of the client state? (right now I'm passing it as a parameter, but that is just to have something that compiles)
    - Also for `retrieve_storage_key`: not really a question, but I need to store the user's `Encrypted<StorageKey>` in the DB as part of a User record, and I can't do that without the rest of register/auth (or wherever the storage key is first generated) being done. 
- Some general/Rust questions:
    - Is making the client wrapper object (or whatever holds client state) going to be part of #154 or should I do it here?
    - Is it bad practice to have multiple nested modules like I made in the local client API? I wanted to separate the arbitrary secrets functionality from the other API function stubs.
    - I see a `BUFFER_SIZE` constant in the `dams` crate but we still have to specify the size when we make mpsc channels to pass to `Channel::create()`. Can I move that somewhere else or make the `BUFFER_SIZE` constant public?

### DIFF
--- a/dams-client/src/api/arbitrary_secrets/generate.rs
+++ b/dams-client/src/api/arbitrary_secrets/generate.rs
@@ -1,7 +1,7 @@
-use crate::{DamsClient, DamsClientError};
+use crate::{api::arbitrary_secrets::LocalStorage, DamsClient, DamsClientError};
 use dams::{
     channel::ClientChannel,
-    crypto::{KeyId, Secret, StorageKey},
+    crypto::{KeyId, StorageKey},
     types::generate::{client, server},
     user::UserId,
 };
@@ -11,19 +11,19 @@ impl DamsClient {
     pub(crate) async fn handle_generate(
         &self,
         channel: &mut ClientChannel,
-    ) -> Result<(KeyId, Secret), DamsClientError> {
+    ) -> Result<(KeyId, LocalStorage), DamsClientError> {
         // Retrieve the storage key
         let storage_key = self.retrieve_storage_key().await?;
 
         // Generate step: get new KeyId from server
         let key_id = get_key_id(channel, self.user_id()).await?;
         // Store step: encrypt secret and send to server to store
-        let secret = {
+        let wrapped_secret = {
             let mut rng = self.rng.lock().await;
             store(channel, self.user_id(), storage_key, &mut rng, &key_id).await?
         };
 
-        Ok((key_id, secret))
+        Ok((key_id, wrapped_secret))
     }
 }
 
@@ -49,7 +49,7 @@ async fn store(
     storage_key: StorageKey,
     rng: &mut StdRng,
     key_id: &KeyId,
-) -> Result<Secret, DamsClientError> {
+) -> Result<LocalStorage, DamsClientError> {
     // Generate and encrypt secret
     let (secret, encrypted) = storage_key.create_and_encrypt_secret(rng, user_id, key_id)?;
     // Serialize and send ciphertext
@@ -65,7 +65,7 @@ async fn store(
     // Await Ok from server
     let result: server::Store = channel.receive().await?;
     if result.success {
-        Ok(secret)
+        Ok(LocalStorage { secret })
     } else {
         Err(DamsClientError::ServerReturnedFailure)
     }
--- a/dams-client/src/api/arbitrary_secrets/generate.rs
+++ b/dams-client/src/api/arbitrary_secrets/generate.rs
@@ -1,4 +1,4 @@
-use crate::{DamsClient, DamsClientError};
+use crate::{client::ClientAction, DamsClient, DamsClientError};
 use dams::{
     channel::ClientChannel,
     crypto::{KeyId, OpaqueExportKey, StorageKey},
@@ -6,36 +6,34 @@ use dams::{
     types::generate::{client, server},
     user::UserId,
 };
-use rand::{CryptoRng, RngCore};
-use tokio::sync::mpsc;
-use tokio_stream::wrappers::ReceiverStream;
-use tonic::{transport::Channel, Status};
-use crate::client::ClientAction;
+use rand::rngs::StdRng;
+use tonic::transport::Channel;
 
 impl DamsClient {
-    pub(crate) async fn handle_generate<T: CryptoRng + RngCore>(
+    pub(crate) async fn handle_generate(
         &self,
         client: &mut DamsRpcClient<Channel>,
         user_id: &UserId,
         export_key: OpaqueExportKey,
     ) -> Result<KeyId, DamsClientError> {
         // Retrieve the storage key
-        let storage_key = self.retrieve_storage_key(client, export_key, user_id).await?;
+        let storage_key = self
+            .retrieve_storage_key(client, export_key, user_id)
+            .await?;
 
         // Create channel to send messages to server
-        let mut channel = Self::create_channel(client, ClientAction::Generate);
+        let mut channel = Self::create_channel(client, ClientAction::Generate).await?;
 
         // Generate step: get new KeyId from server
         let key_id = generate(&mut channel, user_id).await?;
         // Store step: encrypt secret and send to server to store
         {
-            let mut rng = self.rng().lock().await;
-            store(&mut channel, user_id, storage_key, rng, &key_id).await?;
+            let mut rng = self.rng.lock().await;
+            store(&mut channel, user_id, storage_key, &mut rng, &key_id).await?;
         }
 
         Ok(key_id)
     }
-
 }
 
 async fn generate(channel: &mut ClientChannel, user_id: &UserId) -> Result<KeyId, DamsClientError> {
@@ -43,10 +41,7 @@ async fn generate(channel: &mut ClientChannel, user_id: &UserId) -> Result<KeyId
     let generate_message = client::Generate {
         user_id: user_id.clone(),
     };
-    channel
-        .send(generate_message)
-        .await
-        .map_err(|e| Status::aborted(e.to_string()))?;
+    channel.send(generate_message).await?;
 
     // Get KeyId from server
     let generate_result: server::Generate = channel.receive().await?;
@@ -54,18 +49,18 @@ async fn generate(channel: &mut ClientChannel, user_id: &UserId) -> Result<KeyId
     Ok(generate_result.key_id)
 }
 
-async fn store<T: CryptoRng + RngCore>(
+async fn store(
     channel: &mut ClientChannel,
     user_id: &UserId,
     storage_key: StorageKey,
-    rng: &mut T,
+    rng: &mut StdRng,
     key_id: &KeyId,
 ) -> Result<(), DamsClientError> {
     // Generate and encrypt secret
-    let encrypted_secret = storage_key.create_and_encrypt_secret(rng, user_id, key_id);
+    let (_, encrypted) = storage_key.create_and_encrypt_secret(rng, user_id, key_id)?;
     // Serialize and send ciphertext
     let response = client::Store {
-        ciphertext: encrypted_secret,
+        ciphertext: encrypted,
         user_id: user_id.clone(),
     };
     channel.send(response).await?;
--- a/dams-client/src/api/arbitrary_secrets/generate.rs
+++ b/dams-client/src/api/arbitrary_secrets/generate.rs
@@ -1,4 +1,4 @@
-use crate::DamsClientError;
+use crate::{DamsClient, DamsClientError};
 use dams::{
     channel::ClientChannel,
     crypto::{KeyId, OpaqueExportKey, StorageKey},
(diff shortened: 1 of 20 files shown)


picklenerd:
>   * Is making the client wrapper object (or whatever holds client state) going to be part of [Implement message handling system #152 #154](https://github.com/boltlabs-inc/key-mgmt/pull/154) or should I do it here?

It might be better to do it here since (I think?) this is the first endpoint that's going to use it. I don't want to change anything about the implementation in that PR.

>   * Is it bad practice to have multiple nested modules like I made in the local client API? I wanted to separate the arbitrary secrets functionality from the other API function stubs.

Nope I think this is a good practice that keeps the file sizes nice and small. I prefer having small modules like this and re-exporting from the base module if we need anything to be visible externally.

>   * I see a `BUFFER_SIZE` constant in the `dams` crate but we still have to specify the size when we make mpsc channels to pass to `Channel::create()`. Can I move that somewhere else or make the `BUFFER_SIZE` constant public?

This might actually be something we can address in #154. I like how it's kind of hidden away in `Channel` right now so maybe I can extend the channel constructor a bit to cover the local client flow.

marsella:
> * Secret, Secret::generate(), and StorageKey::encrypt() all had to be made public for me to use them in the client. Is that okay or were you envisioning using these in a different way?

Nope, keep those private, please! You can use the `client::create_and_encrypt_secret` function, which bundles those operations together. Although I think `Secret` is already pub on my branch -- are you rebased to the encryption abstraction work yet?

I think you might need to change the signature to take the rng by reference (`&mut (impl CryptoRng + RngCore)`) -- I had hoped to get to that before you did :)

> * Is it okay that I added Clone to KeyId?

I think so. If it's for use in the crypto functions, I don't mind changing the signatures to take references instead of values -- I was a bit sloppy about this.

> * In the retrieve_storage_key step, I need an export key - is this supposed to be part of the client state? (right now I'm passing it as a parameter, but that is just to have something that compiles)

Yes, the `OpaqueExportKey` will be a wrapper around the `export_key` field that is generated when the client finishes OPAQUE [registration](https://docs.rs/opaque-ke/1.2.0/opaque_ke/struct.ClientRegistrationFinishResult.html) or [authentication](https://docs.rs/opaque-ke/1.2.0/opaque_ke/struct.ClientLoginFinishResult.html). It will need to be persisted somehow between authentication and key generation.

> * Also for retrieve_storage_key: not really a question, but I need to store the user's Encrypted<StorageKey> in the DB as part of a User record, and I can't do that without the rest of register/auth (or wherever the storage key is first generated) being done.

Yep. Sorry. Didn't think about this before you started. If it's fully blocking you you could start with the issues in #101 -- in order, probably #124, #123, then #125.

vidyaap:
> Nope, keep those private, please! You can use the `client::create_and_encrypt_secret` ...

Ah thank you, I missed that there was a separate client module!

> I think so. If it's for use in the crypto functions, I don't mind changing the signatures to take references ...

I think references make sense, but I'm not sure whether I will need them as a parameter at all anymore? `Secret::generate()` takes a KeyId but `client::create_and_encrypt_secret` doesn't (unless I'm missing more changes? I'm rebased with develop pre-#154).

> Yes, the `OpaqueExportKey` will be a wrapper around the `export_key` field that is generated when the ...

Got it, this will probably be part of the "client state" wrapper object that we talked about then, right? I think I will hold off on this until the additional registration/auth steps are done.

marsella:
> I think references make sense, but I'm not sure whether I will need them as a parameter at all anymore? Secret::generate() takes a KeyId but client::create_and_encrypt_secret doesn't .

I just missed those parameters! Pointing out [941cf15](https://github.com/boltlabs-inc/key-mgmt/pull/173/commits/941cf156e77a8c3f9a5936cad1d4eefc93219b45) specifically since I added those parameters and changed the calling conventions slightly -- now the `create_and_encrypt_secret` function is a method on the `StorageKey` instead of floating alone in its own module.

dams/src/crypto.rs
```diff
@@ -209,17 +210,18 @@ impl Encrypted<StorageKey> {
         self,
         export_key: OpaqueExportKey,
         user_id: &UserId,
-    ) -> Result<StorageKey, CryptoError> {
+    ) -> Result<StorageKey, DamsError> {
```
picklenerd, line 213: This is a good change!

dams-key-server/src/command.rs
```diff
@@ -1,3 +1,49 @@
...
+    channel: &mut ServerChannel,
+    context: &Context,
+) -> Result<(), DamsServerError> {
+    let request: client::Request = channel.receive().await?;
+    // Find user by ID
+    let user = User::find_user_by_id(&context.db, &request.user_id)
+        .await?
+        .ok_or(DamsServerError::AccountDoesNotExist)?;
+    // Send storage key if set
+    let storage_key = user.storage_key.ok_or(DamsServerError::StorageNotSet)?;
```
picklenerd, line 42: We should make sure that this is getting logged as an error and maybe flagged for manual attention. In most cases this would mean that we have a user in the DB that's in an invalid state. I think the only way this could happen with a valid user is if someone tried to call this while they were in the middle of the registration protocol and haven't sent their storage key yet. This definitely doesn't need to happen in this PR though, so it could probably use its own issue.

I have a PR that should hopefully prevent the invalid user case #212.

dams-key-server/src/error.rs
```diff
@@ -13,6 +13,8 @@ pub enum DamsServerError {
     InvalidUserId,
     #[error("Storage key is already set")]
     StorageKeyAlreadySet,
+    #[error("Storage key is not set for this user")]
+    StorageNotSet,
```
picklenerd, line 17:
```suggestion
    StorageKeyNotSet,
```

dams-client/src/client.rs
```diff
@@ -85,6 +85,18 @@ impl DamsClient {
...
+        self.rng.clone()
+    }
+
+    pub fn tonic_client(&self) -> DamsRpcClient<DamsRpcClientInner> {
+        self.tonic_client.clone()
+    }
+
+    pub fn export_key(&self) -> OpaqueExportKey {
+        self.export_key.clone()
+    }
```
picklenerd, line 98: I think these fields can all be `pub(crate)` instead.

dams-client/src/api/arbitrary_secrets/generate.rs
```diff
@@ -0,0 +1,70 @@
...
+    ) -> Result<(KeyId, Secret), DamsClientError> {
+        // Retrieve the storage key
+        let storage_key = self.retrieve_storage_key().await?;
+
+        // Generate step: get new KeyId from server
+        let key_id = generate(channel, self.user_id()).await?;
+        // Store step: encrypt secret and send to server to store
+        let secret = {
+            let rng = self.rng();
+            let mut rng = rng.lock().await;
```
picklenerd, line 23:
```suggestion
            let mut rng = self.rng.lock().await;
```
If we make the rng public we can do this.

dams-client/src/api/authenticate.rs
```diff
@@ -19,7 +19,7 @@ impl DamsClient {
         rng: &mut T,
         account_name: &AccountName,
         password: &Password,
-    ) -> Result<(OpaqueSessionKey, UserId), DamsClientError> {
+    ) -> Result<(OpaqueSessionKey, OpaqueExportKey, UserId), DamsClientError> {
```
picklenerd, line 22: I think this should have its own type but this is more of a style choice.

dams-key-server/src/database/user.rs
```diff
@@ -61,6 +62,25 @@ pub async fn find_user_by_id(
...
+    db: &Database,
+    user_id: &UserId,
+    secret: Encrypted<Secret>,
+    key_id: KeyId,
+) -> Result<(), DamsServerError> {
+    let collection = db.collection::<User>(constants::USERS);
+    let stored_secret = StoredSecret::new(secret, key_id);
+    let stored_secret_bson = mongodb::bson::to_bson(&stored_secret)?;
+    let filter = doc! { USER_ID: user_id };
+    let update = doc! { "$push": { SECRETS: stored_secret_bson } };
```
picklenerd, line 76: This is really convenient. I was wondering if this was going to be an annoying DB operation but it's not!

dams-client/src/api/arbitrary_secrets/generate.rs
```diff
@@ -0,0 +1,70 @@
...
+            store(channel, self.user_id(), storage_key, &mut rng, &key_id).await?
+        };
+
+        Ok((key_id, secret))
+    }
+}
+
+async fn generate(channel: &mut ClientChannel, user_id: &UserId) -> Result<KeyId, DamsClientError> {
```
marsella, line 31: todo: It seems like this method just gets a new key id from the server. Please either rename it or pull the `create_and_encrypt_secret` method from `store` to here.

dams-key-server/src/server.rs
```diff
@@ -90,6 +92,22 @@ impl DamsRpc for DamsKeyServer {
...
+    ) -> Result<Response<Self::GenerateStream>, Status> {
+        Ok(command::generate::Generate
+            .run(request, self.context())
+            .await?)
+    }
+
+    async fn retrieve_storage_key(
+        &self,
+        request: Request<tonic::Streaming<Message>>,
+    ) -> Result<Response<Self::GenerateStream>, Status> {
```
marsella, line 108: nit: This should be `RetrieveStorageKeyStream`. As an aside, why doesn't clippy throw an unused type warning on L67?

dams-client/src/api/arbitrary_secrets/generate.rs
```diff
@@ -0,0 +1,70 @@
...
+    // Serialize and send ciphertext
+    let response = client::Store {
+        ciphertext: encrypted.clone(),
+        user_id: user_id.clone(),
+    };
+    channel.send(response).await?;
+
+    // TODO spec#39 (design, implementation): Store ciphertext in client-side
+    // storage
+
```
marsella, line 62: suggestion: Can you give this a named type right now? I'm thinking something very simple like creating a `LocalStorage` struct that holds the (unencrypted) secret and the key id and then gets returned from this function. I think that's a little better than returning the unwrapped secret because it's more explicit about its purpose (and it looks like you're already passing that info up to the calling application).

marsella (commented):
This looks great, thanks for all the iterating you've done as the rest of the code has evolved! I have a few comments about naming but otherwise I'm happy with this.

dams-key-server/src/command.rs
vidyaap: Gotcha, I can add a TODO here that points to your issue/PR if that works?

dams-client/src/api/arbitrary_secrets/generate.rs
vidyaap: Sure! Just to confirm though, based on this [spec PR conversation](https://github.com/boltlabs-inc/key-mgmt-spec/pull/125#discussion_r966310012), it looks like we still want to return the key ID and secret as separate objects in a tuple. If that is the case, would the `LocalStorage` type wrap around just the secret, as prep for however we want to handle it later?

dams-client/src/client.rs
vidyaap: I changed export_key and rng, but I left tonic_client as is because otherwise I need to have mutable `self`s in the `generate` and `retrieve_storage_key` functions and that gets weird in the tests when we make a `DamsClient` object.

dams-key-server/src/server.rs
picklenerd: I think it's because the type is required by the trait we're implementing here so it has to be specified whether it's used or not.

vidyaap:
I didn't end up making that "generating without registering" test because it would be testing the same thing as "authenticating without registering".  I also don't think there would be a record to check for whether a secret was added anyway if we didn't register to begin with.

dams-client/src/api/arbitrary_secrets/generate.rs
marsella: Yep, that's perfect!

picklenerd (approved):
Looks good on my side!

dams-client/src/api/arbitrary_secrets/generate.rs
marsella: Nit: now the `store` function should maybe be called `generate_and_store`.

marsella (approved):
One nit but otherwise looks good to me!

PR #169, opened 2022-08-23
Use the HKDF library to create a master key

### LINKED ISSUE #130: Select and integrate HKDF library to derive `MasterKey`
The cryptography module requires an [HMAC-based key derivation function (HKDF)](https://datatracker.ietf.org/doc/html/rfc5869). [RustCrypto has a repository](https://github.com/RustCrypto/KDFs) with an implementation of HKDF that is generic over the hash function ([we use SHA3-256](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/current-development-phase.md#cryptographic-protocol-and-implementation-dependencies)).

context should include the string in the spec and the version.

- [x] Do a cursory search to see if there is another option; compare and contrast
- [ ] Implement master key generation using the HKDF
- [ ] Add dependency choice to the spec via boltlabs-inc/key-mgmt-spec#71

marsella:
Closes #130

This PR uses the Rust-Crypto HKDF library to extract a master key from the export key. To support this, it
- expands the associated data type to support all the relevant context
- expands existing key types, filling in wrappers for export key and master key, adding context to encryption key
- adds one dumb test for master key derivation. Not sure what else to check.

See also: spec update to reflect these changes boltlabs-inc/key-mgmt-spec#71


dams/src/crypto.rs
```diff
@@ -185,13 +223,34 @@ impl Encrypted<StorageKey> {
...
+        let hk = Hkdf::<Sha3_256>::new(None, &self.0);
+        let associated_data = AssociatedData::new()
+            .with_str("OPAQUE-derived Lock Keeper master key")
+            .with_hkdf_parmaters();
+
+        let ad_vec: Vec<u8> = associated_data.clone().into();
+
+        let mut master_key_material = [0u8; 32];
+        hk.expand(&ad_vec, &mut master_key_material)
+            .map_err(CryptoError::KeyDerivationFailed)?;
```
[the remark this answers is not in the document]
picklenerd: Yeah I think it's good to handle it just in case and also good to indicate that it should never return an error. It's probably worth logging this as an error as well so we have a way to see if this is triggering an error when it's not supposed to.

dams/src/crypto.rs
```diff
@@ -185,13 +221,25 @@ impl Encrypted<StorageKey> {
...
 impl OpaqueExportKey {
     /// Derive a [`MasterKey`] from the export key.
     #[allow(unused)]
-    fn derive_master_key(&self) -> MasterKey {
-        todo!()
+    fn derive_master_key(&self) -> Result<MasterKey, CryptoError> {
+        let hk = Hkdf::<Sha3_256>::new(None, &self.0);
+        let associated_data = AssociatedData::new()
+            .with_str("OPAQUE-derived Lock Keeper master key")
+            .with_hkdf_parameters();
```
indomitableSwan, line 233: I would remove the `.with_hkdf_parameters()` context; it's not yet clear that we will need it. Maybe we should expand the issue description for https://github.com/boltlabs-inc/key-mgmt-spec/issues/57 so that we remember to revisit.

dams/src/crypto.rs
```diff
@@ -185,13 +221,25 @@ impl Encrypted<StorageKey> {
...
-pub struct OpaqueExportKey;
+pub struct OpaqueExportKey(GenericArray<u8, U32>);
 
 impl OpaqueExportKey {
     /// Derive a [`MasterKey`] from the export key.
     #[allow(unused)]
-    fn derive_master_key(&self) -> MasterKey {
-        todo!()
+    fn derive_master_key(&self) -> Result<MasterKey, CryptoError> {
+        let hk = Hkdf::<Sha3_256>::new(None, &self.0);
```
indomitableSwan, line 230: Nit: I would prefer a comment here that documents usage of `Hkdf`, e.g., "Produce a uniformly distributed secret using an export key as input key material".

dams/src/crypto.rs
```diff
@@ -185,13 +221,25 @@ impl Encrypted<StorageKey> {
...
+        let associated_data = AssociatedData::new()
+            .with_str("OPAQUE-derived Lock Keeper master key")
+            .with_hkdf_parameters();
+
+        let mut master_key_material = [0u8; 32];
+        hk.expand((&associated_data).into(), &mut master_key_material)
```
indomitableSwan, line 236: Nit: I would prefer a comment here that documents usage of `Hkdf`.

Dependency comment: I do not like the API for this dependency very much. Quite goofy that they don't just provide a single function call for HKF that takes the input key material, optional salt, info string, output length and outputs your key.

dams/src/crypto.rs
```diff
@@ -304,16 +352,29 @@ mod test {
         Ok(())
     }
 
+    // In practice, an export key will be a pseudorandom output form OPAQUE.
```
indomitableSwan, line 355: Typo: "form" --> "from"

dams/src/crypto.rs
```diff
@@ -304,16 +352,29 @@ mod test {
...
+    // Instead, we'll use the encryption key generation function to simulate the
+    // same thing.
+    pub(crate) fn export_key(rng: &mut (impl CryptoRng + RngCore)) -> OpaqueExportKey {
+        OpaqueExportKey(EncryptionKey::new(rng).key)
+    }
```
indomitableSwan, line 360: Does `pub(crate)` mean someone could call this function outside these unit tests?

dams/src/crypto.rs
```diff
@@ -304,16 +352,29 @@ mod test {
...
-    #[should_panic(expected = "not yet implemented")]
-    fn derive_master_key_not_implemented() {
-        let _master_key = OpaqueExportKey.derive_master_key();
+    fn derive_master_key_not_obviously_broken() {
+        let mut rng = rand::thread_rng();
+        let export_key = export_key(&mut rng);
+        let master_key = export_key.derive_master_key().unwrap();
+
+        // make sure the master key isn't the same as the export key
+        assert!(master_key.0.key != export_key.0)
```
indomitableSwan, line 369: I guess other "not obviously broken" test might be to make sure it's not all zeroes, and that adding information using `expand` actually results in a different key.

dams/src/crypto/client.rs
```diff
@@ -43,18 +43,20 @@ pub fn create_and_encrypt_secret(
...
+    use crate::crypto::test::export_key;
 
     #[test]
     #[should_panic(expected = "not yet implemented")]
     fn create_and_encrypt_storage_key_not_implemented() {
-        let rng = rand::thread_rng();
-        let _ = create_and_encrypt_storage_key(rng, OpaqueExportKey);
+        let mut rng = rand::thread_rng();
+        let export_key = export_key(&mut rng);
+        let _ = create_and_encrypt_storage_key(&mut rng, export_key);
```
indomitableSwan, line 53: Ah. Is this why we set the visibility for `export_key` the way we did? Are there other ways we might structure testing to avoid future misuse? Or perhaps even renaming would help. I suppose if we end up with a low-level secret generation building block, we could also use that in testing when we need a key distributed uniformly at random for testing purposes.

indomitableSwan (changes requested):
Looks just about there! A couple of documentation typos/nits, some recommendations for simplification, and a couple comments/questions on testing.

dams/src/crypto.rs
marsella: Yeah, it's a bit weird. The documentation specifically says
> you expand it (perhaps multiple times) into some Output Key Material (OKM) 

I found a deprecated version of RustCrypto that does have it as a single function, but I'm guessing they changed it to support the multiple-key use case. Anyway, I moved these two calls together so they're more obviously a single thing.

dams/src/crypto.rs
marsella: Yes, but it won't compile unless it's called from a test (since this whole module is `#[cfg(test)]`). I have this visibility so I can make a sample export key for testing in the `client` submodule.

dams/src/crypto.rs
indomitableSwan: Yeah, it's fine to _allow_ access to extract and expand functionality, but odd that they wouldn't just provide the most basic use case as one function call.

dams/src/crypto.rs
marsella: Added these. I'm a little leery of the last one because it has a strong dependence on the implementation of the function -- if we use a different HKDF implementation, I'm not sure this will be a useful test (or we'll have to change it to also use the new HKDF). It doesn't test the API of the function, ya know? 

Anyway, this also inspired me to add another test that different export keys produce different master keys.

dams/src/crypto/client.rs
marsella: Yep! I can rename right now. 

One option is for me to move the two functions in this submodule to be methods on the export key and storage key, respectively. Then all the tests would be in one module and `export_key` could be private. I have been thinking about doing this for a while, but waiting until I got to the actual implementation of these functions to do it.

dams/src/crypto.rs
```diff
@@ -55,6 +60,30 @@ impl From<AssociatedData> for Vec<u8> {
...
+    fn with_str(self, ad: &str) -> Self {
+        AssociatedData(format!("{}\n{}", self.0, ad))
+    }
+
+    fn with_aead_parameters(self) -> Self {
+        self.with_str("ChaCha20Poly1305 with 96-bit nonce.")
+    }
+
+    fn with_hkdf_parameters(self) -> Self {
+        self.with_str("HMAC-based extract-and-expand key derivation function using SHA3-256")
```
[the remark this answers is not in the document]
marsella: Removed!

dams/src/crypto/client.rs
indomitableSwan: I like that option.

dams/src/crypto.rs
indomitableSwan: Yeah, that's a good point, and possibly an argument for creating our own KDF abstraction (emphasis on KDF here, not HKDF) eventually. I think this test is fine as is for now, though.

indomitableSwan (approved):
This looks good to me. Responded to a couple of comments, but nothing blocking.

picklenerd (approved):

dams/src/crypto/client.rs
marsella: I got confused and did this on the wrong PR. See [941cf15](https://github.com/boltlabs-inc/key-mgmt/pull/173/commits/941cf156e77a8c3f9a5936cad1d4eefc93219b45) in PR #173.

PR #171, opened 2022-08-23
Server cleanup

### LINKED ISSUE #160: More documentation in README.md on how to run the server
Add some more information on how to run a key server, include some of the following:
- How to run mongodb? (Although there is already a link to mongodb, I believe it is only the download page)
- What is the exact command that you have to run?
- What configuration needs to be in place?

### LINKED ISSUE #161: Move the mongodb .env variables to config file
Use the designated Server.toml config file to add mongodb url and database name, which need to be specified in an environment variable at this point.

### LINKED ISSUE #162: Allow for multiple endpoints for the server
It used to be possible to define multiple endpoints in the Server.toml configuration file, e.g. to define a IPv4 and IPv6 endpoint. We can reintroduce that now that the tonic refactor was successful.

### LINKED ISSUE #163: Use https instead of http when connecting to server
In the integration tests we use a http connection instead of https, we should change that, and make sure http connections are not allowed for our server and client.

vidyaap:
Closes #160 
Closes #161
Closes #162
Closes #163

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,13 +47,28 @@ To run the doctests locally:
 cargo test --all-features --doc --verbose
 ```
 
-To run all unit and integration tests (make sure MongoDB is running first):
+To run all unit and integration tests:
 
+- Start MongoDB in one terminal window (see the [MongoDB docs](https://www.mongodb.com/docs/manual/reference/configuration-options/) for default mongod.conf paths based on your OS):
+```bash
+mongod --config {path_to_mongod.conf}
+```
+
+- Open another terminal window, navigate to this repo and run:
 ```bash
 cargo test --all-features --all-targets
 ```
 
 We follow test-driven development practices and the test suite should be a close mapping to the functionality we currently implement at any given stage of development.
+
+## Running the server locally
+
+To run the server locally, make sure MongoDB is running as above and then run:
+```bash
+cargo run --bin key-server-cli server --config {path_to_server_config} run
+```
+
+There is an example server config file, `dev/Server.toml`. This will start the server on two endpoints, one for IPv4 and one for IPv6, and contains information to connect to a local instance of MongoDB.
 
 ## Build documentation
 
--- a/dams-client/src/api.rs
+++ b/dams-client/src/api.rs
@@ -176,9 +176,17 @@ pub enum Error {
 ///
 /// The returned client should be passed to the remaining API functions.
 pub async fn connect(address: String) -> Result<DamsRpcClient<Channel>, anyhow::Error> {
-    DamsRpcClient::connect(address)
-        .await
-        .map_err(|_| anyhow!("Could not connect to server"))
+    // TODO #174 (design, implementation): determine whether the https link
+    // is secure before passing to tonic
+    if address.starts_with("https:") {
+        DamsRpcClient::connect(address)
+            .await
+            .map_err(|_| anyhow!("Could not connect to server"))
+    } else {
+        Err(anyhow!(
+            "Tried to connect to a server without an https link"
+        ))
+    }
 }
 
 /// Generate a new, distributed digital asset key with the given use
--- a/dams-key-server/src/server.rs
+++ b/dams-key-server/src/server.rs
@@ -85,12 +85,13 @@ impl DamsRpc for DamsKeyServer {
 }
 
 pub async fn start_tonic_server(config: Config) -> Result<(), anyhow::Error> {
-    let db = database::connect_to_mongo().await?;
+    let db =
+        database::connect_to_mongo(&config.database.mongodb_uri, &config.database.db_name).await?;
 
     let dams_rpc_server = DamsKeyServer::new(db, config)?;
     let addr = dams_rpc_server.service.address;
     let port = dams_rpc_server.service.port;
-    info!("{}", TestLogs::ServerSpawned(addr.to_string()));
+    info!("{}", TestLogs::ServerSpawned(format!("{}:{}", addr.to_string(), port.to_string())));
     Server::builder()
         .add_service(DamsRpcServer::new(dams_rpc_server))
         .serve(SocketAddr::new(addr, port))
--- a/dams-key-server/src/server.rs
+++ b/dams-key-server/src/server.rs
@@ -8,7 +8,10 @@ use dams::{
     types::Message,
     TestLogs,
 };
-use futures::FutureExt;
+use futures::{
+    stream::{FuturesUnordered, StreamExt},
+    FutureExt,
+};
 use mongodb::Database;
 use rand::{rngs::StdRng, SeedableRng};
 use std::{
@@ -16,8 +19,9 @@ use std::{
     net::SocketAddr,
     sync::{Arc, Mutex},
 };
+use tokio::signal;
 use tonic::{transport::Server, Request, Response, Status};
-use tracing::info;
+use tracing::{error, info};
 
 #[allow(unused)]
 #[derive(Debug)]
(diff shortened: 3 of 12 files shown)


README.md
```diff
@@ -47,14 +47,29 @@ To run the doctests locally:
 cargo test --all-features --doc --verbose
 ```
 
-To run all unit and integration tests (make sure MongoDB is running first):
+To run all unit and integration tests:
 
+- Start MongoDB in one terminal window:
+```bash
+mongod --config {path_to_mongo.conf}
```
picklenerd, line 54: Is there a default config that gets installed with mongo? Would it be useful to have our own config in this repo?

vidyaap: Yes but the default path changes by OS. I can list out the main ones here if that would be better?

dams-tests/tests/common.rs
```diff
@@ -129,7 +126,8 @@ async fn client_test_config() -> dams::config::client::Config {
 
 /// Encode the customizable fields of the keymgmt server Config struct for
 /// testing.
-async fn server_test_config() -> dams::config::server::Config {
+pub async fn server_test_config() -> dams::config::server::Config {
+    let _ = fs::create_dir("tests/gen");
```
picklenerd, line 130: I think it's ok to `.expect` here since it's in the tests. If there's a problem creating the directory this will swallow up the error and we won't know about it until later when the directory is expected to exist.

dams-tests/tests/common.rs
```diff
@@ -142,8 +140,19 @@ async fn server_test_config() -> dams::config::server::Config {
...
     });
+    // Format string for database information
+    let database = HashMap::from([
+        ("mongodb_uri", "mongodb://localhost:27017"),
+        ("db_name", "dams-test-db"),
+    ])
+    .into_iter()
+    .fold("\n[database]".to_string(), |acc, (key, value)| {
+        format!("{}\n{} = \"{}\"", acc, key, value)
+    });
```
picklenerd, line 151: I'm gonna be honest, I really like this kind of functional wizardry. It always makes me happy.

I think something like this might be more clear though:

```suggestion
    let database = r#"
        [database]
        mongodb_uri = "mongodb://localhost:27017"
        db_name = "dams-test-db"
    "#;
```

dams-tests/tests/common.rs
```diff
@@ -142,8 +140,19 @@ async fn server_test_config() -> dams::config::server::Config {
...
 
-    write_config_file(SERVER_CONFIG, services.to_string());
+    // Format service string and database string into full config
+    let config = format!("{}\n{}", services, database);
```
picklenerd, line 154: See comment above. I think this whole config string could be a [raw string literal](https://rahul-thakoor.github.io/rust-raw-string-literals/).

dev/Server.toml
```diff
@@ -11,3 +11,7 @@ private_key = "localhost.key"
 certificate = "localhost.crt"
 opaque_path = "dev/opaque"
 opaque_server_key = "dev/opaque/server_setup"
+
```
picklenerd, line 14: Could you add the port to these just so it's easy to find?

dams-key-server/src/server.rs
```diff
@@ -85,17 +85,46 @@ impl DamsRpc for DamsKeyServer {
...
-    Server::builder()
-        .add_service(DamsRpcServer::new(dams_rpc_server))
-        .serve(SocketAddr::new(addr, port))
-        .await
-        .map_err(|e| anyhow!("Cannot start server: {:?}", e))
+            async move {
+                let dams_rpc_server = DamsKeyServer::new(db, config, service)?;
+                let addr = dams_rpc_server.service.address;
+                let port = dams_rpc_server.service.port;
+                info!("{}", TestLogs::ServerSpawned(addr.to_string()));
```
picklenerd, line 105: Can you add the port here so it gets printed at startup? I feel like this might break those tests that we were troubleshooting before but I think it would be nice for users to have the full path in the console.

README.md
picklenerd: We could probably just add a link to this page: https://www.mongodb.com/docs/manual/reference/configuration-options/

dams-client/src/api.rs
```diff
@@ -176,9 +176,13 @@ pub enum Error {
 ///
 /// The returned client should be passed to the remaining API functions.
 pub async fn connect(address: String) -> Result<DamsRpcClient<Channel>, anyhow::Error> {
-    DamsRpcClient::connect(address)
-        .await
-        .map_err(|_| anyhow!("Could not connect to server"))
+    if address.starts_with("http:") {
```
picklenerd, line 179: Is this sufficient to prove that it's a secure connection? It seems like it might be but I don't actually know.

vidyaap: No, I forgot to leave a comment/question on this. I was being quite literal here about https links, but really it should at least check for "anything that isn't https" rather than "is it http".  I was wondering if there was a better way to do the check in general since the function expects a String.

picklenerd: I was thinking maybe there might be a function in the type that gets returned by tonic that would tell us whether or not it's https but I didn't see one when I looked through the tonic code.

Maybe we could connect to the path using stuff from `std` and figure it out that way before connecting to tonic. Or maybe this really is sufficient.

I think it's worth making a new issue for this, writing a TODO in the code, and leaving this as-is in this PR unless someone knows for sure whether or not this is enough.

dams-tests/tests/common.rs
picklenerd: Having a file would be good. We could also add a function to the config type to load from a string and have `Config::load` call that function once the file is read to a string.

dams-client/src/api.rs
vidyaap: Well if we gave the function an https string but the link didn't have the right certs, or didn't exist, then we wouldn't be able to connect anyway and we get an error back for that. I'm just not sure if waiting for that error is the appropriate behavior. Still, I can make the issue+TODO.

picklenerd: That's a good point. Maybe we could add a simple health check endpoint and hit that during startup.

dams-tests/tests/common.rs
```diff
@@ -73,7 +70,7 @@ pub async fn setup(db: Database) -> ServerFuture {
     // Note: hard-coded to match the 2-service server with default port.
     let checks = vec![await_log(
         Party::Server,
-        TestLogs::ServerSpawned(SERVER_ADDRESS.to_string()),
+        TestLogs::ServerSpawned(SERVER_ADDRESS.to_string() + ":1113"),
```
picklenerd, line 73:
```suggestion
        TestLogs::ServerSpawned(format!("{}:1113", SERVER_ADDRESS)),
```
`+` works but using `format!` is the idiomatic way to do this.

dams-client/src/api.rs
picklenerd: We can do that in a PR attached to the new issue.

picklenerd (approved):

dams-tests/tests/common.rs
marsella: Sorry I missed this! The reason is that the config struct was defined as `#[non-exhaustive]` so the only way to construct it was to use the `load` function which takes a `Path`.

vidyaap: Thanks, that makes sense! I was wondering why we didn't start off with a file in the tests though, instead of writing a string to a file and then back?

marsella: Hmm. Either I didn't know that was an option or I thought the string -> file approach was going to be more flexible and easier to modify when we tested lots of complicated config options (aka I was optimizing too early).

Issue #134, opened 2022-08-12
Implement public cryptography for generate-and-store flow


marsella:
The [generate-and-store]((https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/cryptographic_flows.md#generate-and-store-a-secret) flow has  few pieces of external-facing cryptography that need to be implemented.

- [x] Implement generation of key IDs as defined in 2.ii.a-c

- [x] Implement the generate-and-encrypt bundle (defined in crypto/client) as defined in 3.i-ii (not including the network transmission)

marsella:
Adding dependency on #131 because key ID generation requires hashing the `UserId` and I want to have the correct representation for that before working on this one. See initial draft in 134-generate-cryptography branch.

Issue #131, opened 2022-08-12
Implement public cryptography for registration


marsella:
The crypto module provides some "wrapper" functions that bundle together multiple cryptographic operations correctly for use in protocols. For registration, this includes the `create_and_encrypt_storage_key` function. This is steps 1iii, 1iv, and 1v in [the registration protocol](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/cryptographic_flows.md#register). Close this issue when it's implemented.

The registration step also requires `UserId` generation by the key server; see the [spec](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/systems-architecture.md#opening-a-registration-session) . 

The session holds the session key as plain bytes. Replace this with a proper named type like `SessionKey(u8; 64)`.

indomitableSwan:
@marsella Shouldn't this ticket be implemented by a cryptographer?

PR #173, opened 2022-08-23
Generate a secret

### LINKED ISSUE #132: Implement secret generation protocol
The cryptographic module requires several kinds of secrets to be [generated with a single protocol](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/cryptographic_flows.md#generate-a-secret).

The protocol itself is pretty straightforward. The most difficult part is to decide how to represent and store additional context when generating a secret. Different secrets will require different, overlapping contexts:
- An arbitrary `Secret` needs the user id and key id it's attached to
- A `StorageKey` might have other context, like a fixed string explaining what it is
- Both should specify the current version of key-mgmt

- [ ] Decide (and log) any additional context that isn't noted here
- [ ] Generate an appropriate abstraction to define different contexts
- [ ] Implement protocol

marsella:
This closes #132.

Generates a secret.

#132 has a bunch of todos about figuring out how to apply contexts correctly. The `AssociatedData` abstraction we figured out in a previous PR still looks like it works here. I'm pushing out the filling-in of the context to the calling function (which will still be within the crypto module) -- that will be implemented in #131 for making the storage key, #134 for making an arbitrary secret.

I took the liberty of fixing some secret-generation-related public APIs. This affects @vidyaap's work in #167 -- she noticed that I had missed some parameters. Those are now added and I fixed the RNG type to adhere to our best practices.


indomitableSwan (approved):
Just one non-blocking comment.

dams/src/crypto/client.rs
```diff
@@ -34,7 +37,9 @@ pub fn create_and_encrypt_storage_key(
 /// 2. Encrypt it under the [`StorageKey`], using an AEAD scheme
 #[allow(unused)]
 pub fn create_and_encrypt_secret(
-    rng: impl CryptoRng + RngCore,
+    rng: &mut (impl CryptoRng + RngCore),
+    user_id: &UserId,
+    key_id: &KeyId,
     storage_key: StorageKey,
 ) -> Encrypted<Secret> {
     todo!()
```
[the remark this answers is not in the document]
marsella: Right now, the protocol implementation is completely shielded from having to construct `AssociatedData` directly. If we ditch this abstraction, they'll have to manually create it and pass it to the secret generation function. I would prefer to avoid this, so we can keep that abstraction private and avoid user error (e.g. from missing user or key ID, or from putting something else in there).

We can revisit once the client exists, but I'm not sure what kind of improvement to expect from using that here. If it has the RNG and the user ID, it could reduce one parameter, but I don't think it will store the key ID or the storage key.

dams/src/crypto.rs
```diff
@@ -261,16 +260,23 @@ impl KeyId {
...
     /// Generate a new secret of length `len`.
-    fn generate(rng: impl CryptoRng + RngCore, len: u32, user_id: UserId, key_id: KeyId) -> Self {
-        todo!()
+    fn generate(
+        rng: &mut (impl CryptoRng + RngCore),
+        len: usize,
+        associated_data: AssociatedData,
+    ) -> Self {
+        Self {
+            material: std::iter::repeat_with(|| rng.gen()).take(len).collect(),
```
marsella, line 277: @indomitableSwan FYI - I looked into other ways of generating a `Vec` of random values. There's a nice way to use a single call to `gen` if the length is constant and less than 32, but allowing variable length I think forces us to use this approach.

indomitableSwan: ah, interesting. Could you add both methods to our internal resource on using rngs?

marsella: Yep, updated!

indomitableSwan (approved):
LGTM.

PR #180, opened 2022-08-25
Message fields now use concrete types #168

### LINKED ISSUE #168: Message types should handle byte conversions for their fields
Message types have fields with the type `Vec<u8>`. These bytes are currently converted to more useful types within the the handlers. This makes it more difficult to see how these type conversions are handled as a whole and complicates the code in the handlers.

For example, we have the following type:

```rust
    pub struct AuthenticateStart {
        pub message: Vec<u8>,
        pub user_id: Vec<u8>,
    }
```

The `message` field is converted in the handler with:

```rust
```

The ideal solution would be to type these fields with their real types within the message types and let them be automatically converted during deserialization. This may require custom deserializers for some types. If conversion requires additional parameters, this solution will not work.

If custom deserializers are too complicated for short-term goals or conversion requires additional parameters, we can implement `into_parts()` functions for message types that deconstruct the message type, convert their fields, and return a tuple containing the converted fields.

picklenerd:
This PR removes a layer of `bincode` serialization/deserialization and changes the fields within each message type to be a concrete type rather than raw bytes. 

Closes #168

### DIFF
--- a/dams-client/src/api/authenticate.rs
+++ b/dams-client/src/api/authenticate.rs
@@ -7,7 +7,6 @@ use dams::{
 };
 use opaque_ke::{
     ClientLogin, ClientLoginFinishParameters, ClientLoginFinishResult, ClientLoginStartResult,
-    CredentialResponse,
 };
 use rand::{CryptoRng, RngCore};
 use tokio::sync::mpsc;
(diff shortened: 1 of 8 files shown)


marsella (approved):
This looks good to me. The issue mentioned a potential problem with custom deserializers, did that not end up being a problem? Would the issue arise if the`impl_message_conversion` macro needs different deserialize calls for different messages?

Please update the PR description to link it to the issue.

picklenerd:
> The issue mentioned a potential problem with custom deserializers, did that not end up being a problem? Would the issue arise if theimpl_message_conversion macro needs different deserialize calls for different messages?

I thought we might have to write them for the opaque types but they already implement `Serialize` and `Deserialize` so `bincode` can handle them naturally. Once I untangled this a bit I realized we were just running through `bincode` twice. Once for the `Message` and again for each field. Now we just run `bincode` on the `Message` and it takes care of the fields at the same time.

If we replace `bincode` with another encoding scheme, we may have to implement custom serialization depending on whether or not it supports `serde`. Hopefully we can still cover that in the macro but it'll depend on the requirements.

PR #182, opened 2022-08-25
Finish Registration protocol #125

### LINKED ISSUE #123: Write tonic helper functions for registration

### LINKED ISSUE #124: Update server database to store `StorageKey`s

### LINKED ISSUE #125: Implement the registration protocol

picklenerd:
This PR adds the storage key to the database and implements the registration protocol.

In order to ensure we have an authenticated channel before creating the storage key, I added a new operation called `CreateStorageKey`. The client-side register protocol is now executed in three steps:

1. Register
2. Authenticate
3. CreateStorageKey

I had to change the inner type of `OpaqueExportKey` to `Box<[u8; 64]>` to match the output from OPAQUE. Some testing code had to change as a result of this.

I also had to change the inner type of `UserId` to be a `String` to get it to play nice with the database. The user id I still a UUID generated in the same way. 

This task really needs an integration test case, but there's no way to check database values with the current test harness. Instead of making major changes to the test harness or duplicating all of the setup code, I made a ticket to add this test case #208 . We're currently working on refactoring the integration tests so we can take this requirement into consideration.

Closes #123
Closes #124
Closes #125

### DIFF
--- a/dams-key-server/src/command/create_storage_key.rs
+++ b/dams-key-server/src/command/create_storage_key.rs
@@ -1,0 +1,81 @@
+use crate::{
+    database::user::{find_user, set_storage_key},
+    error::DamsServerError,
+    server::Context,
+};
+
+use dams::{
+    channel::ServerChannel,
+    types::{
+        create_storage_key::{client, server},
+        Message, MessageStream,
+    },
+    user::UserId,
+};
+use tokio_stream::wrappers::ReceiverStream;
+use tonic::{Request, Response};
+
+#[derive(Debug)]
+pub struct CreateStorageKey;
+
+impl CreateStorageKey {
+    pub async fn run<'a>(
+        &self,
+        request: Request<tonic::Streaming<Message>>,
+        context: Context,
+    ) -> Result<Response<MessageStream>, DamsServerError> {
+        let (mut channel, rx) = ServerChannel::create(request.into_inner());
+
+        let _ = tokio::spawn(async move {
+            let user_id = send_user_id(&mut channel, &context).await?;
+            store_storage_key(user_id, &mut channel, &context).await?;
+
+            Ok::<(), DamsServerError>(())
+        });
+
+        Ok(Response::new(ReceiverStream::new(rx)))
+    }
+}
+
+async fn send_user_id(
+    channel: &mut ServerChannel,
+    context: &Context,
+) -> Result<UserId, DamsServerError> {
+    let request: client::RequestUserId = channel.receive().await?;
+    let user = find_user(&context.db, &request.account_name).await?;
+
+    if let Some(user) = user {
+        if user.storage_key.is_some() {
+            return Err(DamsServerError::StorageKeyAlreadySet);
+        }
+
+        let reply = server::SendUserId {
+            user_id: user.user_id.clone(),
+        };
+
+        channel.send(reply).await?;
+
+        Ok(user.user_id)
+    } else {
+        Err(DamsServerError::AccountDoesNotExist)
+    }
+}
+
+async fn store_storage_key(
+    user_id: UserId,
+    channel: &mut ServerChannel,
+    context: &Context,
+) -> Result<(), DamsServerError> {
+    let client_message: client::SendStorageKey = channel.receive().await?;
+
+    if client_message.user_id != user_id {
+        return Err(DamsServerError::InvalidUserId);
+    }
+
+    set_storage_key(&context.db, &user_id, client_message.storage_key).await?;
+
+    let reply = server::CreateStorageKeyResult { success: true };
+    channel.send(reply).await?;
+
+    Ok(())
+}
--- a/dams-key-server/src/command/register.rs
+++ b/dams-key-server/src/command/register.rs
@@ -55,7 +55,7 @@ async fn register_start(
     let user = User::find_user(&context.db, &start_message.account_name).await?;
 
     if user.is_some() {
-        Err(DamsServerError::AccountAlreadyExists)
+        Err(DamsServerError::InvalidUserId)
     } else {
         // Registration can continue if user ID doesn't exist yet
         let server_registration_start_result = ServerRegistration::<OpaqueCipherSuite>::start(
--- a/dams-key-server/src/server.rs
+++ b/dams-key-server/src/server.rs
@@ -62,6 +62,7 @@ pub struct Context {
 impl DamsRpc for DamsKeyServer {
     type RegisterStream = dams::types::MessageStream;
     type AuthenticateStream = dams::types::MessageStream;
+    type CreateStorageKeyStream = dams::types::MessageStream;
 
     async fn register(
         &self,
@@ -77,6 +78,15 @@ impl DamsRpc for DamsKeyServer {
         request: Request<tonic::Streaming<Message>>,
     ) -> Result<Response<Self::AuthenticateStream>, Status> {
         Ok(command::authenticate::Authenticate
+            .run(request, self.context())
+            .await?)
+    }
+
+    async fn create_storage_key(
+        &self,
+        request: Request<tonic::Streaming<Message>>,
+    ) -> Result<Response<Self::CreateStorageKeyStream>, Status> {
+        Ok(command::create_storage_key::CreateStorageKey
             .run(request, self.context())
             .await?)
     }
--- a/dams/src/types/create_storage_key.rs
+++ b/dams/src/types/create_storage_key.rs
@@ -1,0 +1,41 @@
+pub mod client {
+    use crate::{
+        crypto::{Encrypted, StorageKey},
+        impl_message_conversion,
+        user::{AccountName, UserId},
+    };
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Debug, Deserialize, Serialize)]
+    pub struct RequestUserId {
+        pub account_name: AccountName,
+    }
+
+    #[derive(Debug, Deserialize, Serialize)]
+    pub struct SendStorageKey {
+        pub user_id: UserId,
+        pub storage_key: Encrypted<StorageKey>,
+    }
+
+    // TODO #186: These messages need to be authenticated
+    impl_message_conversion!(RequestUserId, SendStorageKey);
+}
+
+pub mod server {
+    use crate::{impl_message_conversion, user::UserId};
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Debug, Deserialize, Serialize)]
+    pub struct SendUserId {
+        pub user_id: UserId,
+    }
+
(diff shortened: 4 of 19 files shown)


dams-key-server/src/command/register.rs
```diff
@@ -64,46 +69,71 @@ async fn register_start(
...
+        let _object_id = User::create_user(
+            &context.db,
+            &user_info.user_id,
+            &user_info.account_name,
+            finish_message.storage_key,
+            &server_registration,
+        )
+        .await?;
+    } else {
+        return Err(DamsServerError::InvalidUserId);
```
marsella, line 136: This error overlaps with `AccountAlreadyExists`. I might keep the `InvalidUserId` above but replace this one.

marsella (changes requested):
I appreciate your move toward reduced network traffic, but the message ordering is not correct here. The user ID that the client gets from the server needs to be sent over an authenticated channel, which can only happen _after_ the opaque protocol is complete. 

I know we don't have authenticated channels yet, but I think it's better to separate the messages from now and flag them in the docs with #186, [as in authentication](https://github.com/boltlabs-inc/key-mgmt/blob/develop/dams/src/types/authenticate.rs#L48-L51) so it doesn't slip under our radar later.

dams-key-server/src/command/register.rs
picklenerd: Agreed. I think `AccountAlreadyExists` gives too much info.

dams/src/types/register.rs
```diff
@@ -14,20 +19,23 @@ pub mod client {
     /// pass user ID and registration-finish message from OPAQUE
     pub struct RegisterFinish {
         pub registration_upload: RegistrationUpload<OpaqueCipherSuite>,
+        pub user_id: UserId,
+        pub storage_key: Encrypted<StorageKey>,
```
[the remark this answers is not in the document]
picklenerd: If that's the case, maybe creating a storage key should be a separate operation since the client is going to have to call register, then authenticate, then create the key.

picklenerd:
> I appreciate your move toward reduced network traffic, but the message ordering is not correct here. The user ID that the client gets from the server needs to be sent over an authenticated channel, which can only happen after the opaque protocol is complete.

I wasn't trying to do anything about network traffic. I was looking at the wrong page of the spec based on the link in the ticket: https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/system-functionalities.md#register

I think this is the correct page: https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/systems-architecture.md#opening-and-using-a-registration-session

dams-key-server/src/command/create_storage_key.rs
```diff
@@ -0,0 +1,81 @@
...
+    channel: &mut ServerChannel,
+    context: &Context,
+) -> Result<(), DamsServerError> {
+    let client_message: client::SendStorageKey = channel.receive().await?;
+
+    if client_message.user_id != user_id {
+        return Err(DamsServerError::InvalidUserId);
+    }
+
+    set_storage_key(&context.db, &user_id, client_message.storage_key).await?;
```
marsella, line 75: What happens if this fails? Should we send back a `CreateStorageKeyResult { success: false }`?

dams-key-server/src/command/create_storage_key.rs
```diff
@@ -0,0 +1,81 @@
...
+
+#[derive(Debug)]
+pub struct CreateStorageKey;
+
+impl CreateStorageKey {
+    pub async fn run<'a>(
+        &self,
+        request: Request<tonic::Streaming<Message>>,
+        context: Context,
+    ) -> Result<Response<MessageStream>, DamsServerError> {
```
marsella, line 26: question: I can't quite tell what the overall failure behavior is here. Does the client get some kind of error message if the server generates a `DamsServerError` here, or will the connection just drop?

dams-key-server/src/server.rs
```diff
@@ -78,6 +79,15 @@ impl DamsRpc for DamsKeyServer {
             .run(request, self.context())
             .await?)
     }
+
+    async fn create_storage_key(
+        &self,
+        request: Request<tonic::Streaming<Message>>,
+    ) -> Result<Response<Self::AuthenticateStream>, Status> {
```
marsella, line 86: Should this create a `CreateStorageKeyStream`? I see they map to the same type.

marsella (commented):
The protocol looks good to me! I have a couple inline questions about error handling and expected behaviors.

I also can't quite tell how the server can recognize that they're dealing with the same client between registration+authentication and creating the storage key, and whether they persist the information (the session key) they'll need soon to authenticate messages in this flow. This might be something we can address in #186, but we should make sure it's documented there if we'll need changes to the server.

More on error handling: How to we want to handle cleaning the server state if there are errors in creating & storing the storage key? A user won't be able to "try again" by running the registration protocol because their account will already exist. I'm not sure if this needs to be addressed right now, but we should think about what server behavior we'd expect in the long run (deleting the whole record and starting over? Adding a conditional to circumvent the "register" part of registration if the account exists without a storage key?)

dams-key-server/src/server.rs
picklenerd: Yes it should! These type aliases are required by the `tonic` interface. Maybe it would be better to make them all return `MessageStream` and just let those type aliases sit around doing nothing.

dams-key-server/src/command/create_storage_key.rs
picklenerd: Right now the client will get whatever error is returned wrapped in a `tonic::Status`. I think we'll need to flesh out how we want errors displayed to the client in the future but I think this is best for now.

dams/src/types/create_storage_key.rs
```diff
@@ -0,0 +1,42 @@
+pub mod client {
+    use crate::{
+        crypto::{Encrypted, StorageKey},
+        impl_message_conversion,
+        user::{AccountName, UserId},
+    };
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Debug, Deserialize, Serialize)]
+    /// pass user ID and registration-start message from OPAQUE
```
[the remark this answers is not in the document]
picklenerd: Updated. I was going to write similar comments but I don't think they add anything.

dams-key-server/src/command/create_storage_key.rs
picklenerd: I updated the `set_storage_key` docs with descriptions of all of the possible errors.

picklenerd: It's not possible to get a `success` false at this point. We could probably use a better type for this. Something I've seen in the past is a generic `Success(bool)` type. Or we could do a basic two-variant enum.

picklenerd:
> I also can't quite tell how the server can recognize that they're dealing with the same client between registration+authentication and creating the storage key, and whether they persist the information (the session key) they'll need soon to authenticate messages in this flow. This might be something we can address in https://github.com/boltlabs-inc/key-mgmt/issues/186, but we should make sure it's documented there if we'll need changes to the server.

I think we never settled on how to store the session key. I remember we talked about potentially using redis but we could also store them in the `DamsKeyServer` for now. It looks like session key handling is described in #186 but it seems like we could break this off into its own issue since #186 is also concerned with the MAC solution.

picklenerd:
> More on error handling: How to we want to handle cleaning the server state if there are errors in creating & storing the storage key? A user won't be able to "try again" by running the registration protocol because their account will already exist. I'm not sure if this needs to be addressed right now, but we should think about what server behavior we'd expect in the long run (deleting the whole record and starting over? Adding a conditional to circumvent the "register" part of registration if the account exists without a storage key?)

This is a really good point. I think it would be easier to just delete the whole database entry. I'll make a new issue to address this.

marsella:
> > I also can't quite tell how the server can recognize that they're dealing with the same client between registration+authentication and creating the storage key, and whether they persist the information (the session key) they'll need soon to authenticate messages in this flow. This might be something we can address in #186, but we should make sure it's documented there if we'll need changes to the server.
> 
> I think we never settled on how to store the session key. I remember we talked about potentially using redis but we could also store them in the `DamsKeyServer` for now. It looks like session key handling is described in #186 but it seems like we could break this off into its own issue since #186 is also concerned with the MAC solution.

Oh, right. Thanks. I think it would be good to make a separate issue (maybe an issue belonging to that epic) about this specifically.

dams-key-server/src/command/create_storage_key.rs
marsella: Ah, I see. As you say below, we might want to revisit how we expose errors to the client but this should be fine.

marsella (approved):

picklenerd:
Created #213 to handle session keys

PR #183, opened 2022-08-26
172 client abstraction

### LINKED ISSUE #172: Create initial Client abstraction in dams-client
The client needs a way to store state between requests and reduce boilerplate in handler functions. We would like to rename the `Session` type in the client to `Client` and use it as a general abstraction for client operations. A `Client` object can be passed to request handlers to provide all of the necessary functionality to complete a request.

`Client` responsibilities:

- Store state that is used between requests
- Store common infrastructure such as RNGs
- Store a `DamsRpcClient` that can be used to create channels for each request type

The `impl` for `Client` should provide functions that create a channel for each endpoint. The logic for these functions should contain the boilerplate we're using to create channels in the current handler functions along with any other setup code that makes protocol implementation simpler:

```rust
    // Create channel to send messages to server after connection is established via RPC
    let stream = ReceiverStream::new(rx);

    // Server returns its own channel that is uses to send responses
    let server_response = client.authenticate(stream).await?.into_inner();

    ClientChannel::create(tx, server_response)
```

This will allow us to create a channel at the top of each handler with something like `let mut channel = client.authenticate_channel().await?`. Having a generic `Client::channel(some_way_to_dispatch_request_type)` would be ideal but may be challenging. Ping @picklenerd if you want some ideas to get started.

This issue supersedes #30. I believe this `Client` type can handle the extensions described in that issue when we need to implement them in the future.

vidyaap:
Closes #172 

Aside from general review I also have these questions:
- I had to remove the Default impl for `DamsClient` (i.e. `Session`) because there's no default for the tonic client object. We talked about calling the `connect` function to get a tonic client, but because it's async, I wasn't able to await it in the standard Default impl. For now, I removed the session parameter from the rest of the API functions and tests but is there a way for me to make a dummy Default for DamsClient anyway? Can async_trait do anything here?
- @marsella Is it appropriate for the storage key to be a part of the state for this object? Not to replace it being in persistent storage, but because most of the flows have a "retrieve_storage_key" step that would be nice not to have to duplicate.

Also not sure if the docs still make sense with me just replacing "Session" with "DamsClient"...


dams-client/src/api.rs
```diff
@@ -70,47 +74,74 @@ impl Password {
...
+        }
+    }
+
     /// Open a new mutually authenticated session between a previously
     /// registered user and a key server.
     ///
-    /// Output: If successful, returns an open [`Session`] between the specified
-    /// [`UserId`] and the configured key server.
+    /// Output: If successful, returns a [`DamsClient`] holding a
+    /// `tonic`client object, a session key, and configuration information.
```
picklenerd, line 117:
```suggestion
    /// `tonic` client object, a session key, and configuration information.
```

dams-client/src/api.rs
```diff
@@ -121,39 +152,56 @@ impl Session {
...
     /// a mutually authenticated session with the server.
     ///
     /// This only needs to be called once per user; future sessions can be
-    /// created with [`Session::open()`].
+    /// created with [`DamsClient::open()`].
     ///
-    /// Output: If successful, returns an open [`Session`] between the specified
-    /// [`UserId`] and the configured key server.
+    /// Output: If successful, returns a [`DamsClient`] holding a
+    /// `tonic`client object, a session key, and configuration information.
```
picklenerd, line 162:
```suggestion
    /// `tonic` client object, a session key, and configuration information.
```

dams-client/src/api.rs
```diff
@@ -70,47 +74,74 @@ impl Password {
...
 
 #[allow(unused)]
-impl Session {
+impl DamsClient {
+    /// Connect to the gRPC client and return it to the client app.
+    ///
+    /// The returned client should be stored as part of the [`DamsClient`]
+    /// state.
+    async fn connect(address: String) -> Result<DamsRpcClient<Channel>, SessionError> {
+        if address.starts_with("https:") {
```
picklenerd, line 104:
```suggestion
    async fn connect(address: Uri) -> Result<DamsRpcClient<Channel>, SessionError> {
        if address.scheme() == Some(&Scheme::HTTPS) {
```
I think this should take a [Uri](https://docs.rs/http/latest/http/uri/struct.Uri.html) instead of a String. It's a bit more explicit and `tonic` can take it in `DamsRpcClient::connect`. 

`http::uri::Uri` is re-exported as `tonic::transport::Uri` so it might be good to use the `tonic` import in case the versions diverge.

It might also be convenient to store it as a `Uri` in the config, but `Uri` doesn't support `serde` so we'd have to work around that. I know a really nice trick for this if you want to hop on a call and go over it. We could also just add a `server_uri()` function to the config that calls `Uri::from_str(self.server_location)`.

dams-client/src/api.rs
```diff
@@ -70,47 +74,74 @@ impl Password {
...
+            Err(SessionError::ServerConnectionFailed)
+        }
+    }
+
     /// Open a new mutually authenticated session between a previously
     /// registered user and a key server.
     ///
-    /// Output: If successful, returns an open [`Session`] between the specified
-    /// [`UserId`] and the configured key server.
+    /// Output: If successful, returns a [`DamsClient`] holding a
```
picklenerd, line 116: I think it might actually be better to cut this off after `returns a [`DamsClient`]`. Users can check the `DamsClient` type to see what it does/contains. I can see this second part becoming out of date in the future as we add things to the client.

dams-client/src/api.rs
```diff
@@ -121,39 +152,56 @@ impl Session {
...
+        // Create channel to send messages to server after connection is established via
+        // RPC
+        let (tx, rx) = mpsc::channel(2);
+        let stream = ReceiverStream::new(rx);
+
+        // Server returns its own channel that is uses to send responses
+        let server_response = match action {
+            ClientAction::Register => client.register(stream).await,
+            ClientAction::Authenticate => client.authenticate(stream).await,
+        }
```
picklenerd, line 197: I think this is a great approach. I'm a big fan of this kind of manual dispatch in Rust.

dams-client/src/api.rs
```diff
@@ -70,47 +74,74 @@ impl Password {
...
+        config: &Config,
+    ) -> Result<Self, SessionError> {
+        let server_location = config.server_location.as_str();
+        let mut client = Self::connect(server_location.to_string()).await?;
+        Self::authenticate(client, rng, user_id, password, config).await
+    }
+
+    async fn authenticate<T: CryptoRng + RngCore>(
+        mut client: DamsRpcClient<Channel>,
+        rng: &mut T,
```
picklenerd, line 131: I think this function should create its own RNG and store it in the `DamsClient`. We wanted to avoid creating new RNGs for each request but I think it's ok to create a new one for each `DamsClient`.

picklenerd: Just realized that `register` needs this to have the `rng` param. It makes sense to keep it as a param here but it should take it by value and stick it in the `DamsClient` so we don't have to keep passing it around. I think `open` and `register` should create their own RNG and pass it to `authenticate` unless there's something else I'm missing.

dams-client/src/api/authenticate.rs
```diff
@@ -1,36 +1,23 @@
...
-use tokio::sync::mpsc;
-use tokio_stream::wrappers::ReceiverStream;
-use tonic::{transport::Channel, Status};
+use tonic::Status;
 
 use super::Password;
 
 pub(crate) async fn handle<T: CryptoRng + RngCore>(
-    client: &mut DamsRpcClient<Channel>,
+    mut channel: ClientChannel,
```
picklenerd, line 16: This is great. I like that it takes the channel by value so you know it's going to be closed at the end of this function.

dams/src/config/client.rs
```diff
@@ -1,25 +1,11 @@
...
-    pub connection_timeout: Option<Duration>,
-    #[serde(default = "defaults::max_pending_connection_retries")]
-    pub max_pending_connection_retries: usize,
-    #[serde(with = "humantime_serde", default = "defaults::message_timeout")]
-    pub message_timeout: Duration,
-    #[serde(default = "defaults::max_message_length")]
-    pub max_message_length: usize,
-    #[serde(default = "defaults::max_note_length")]
-    pub max_note_length: u64,
+    pub server_location: String,
```
picklenerd, line 8:
```suggestion
    pub server_uri: Uri,
```

I mentioned this in another comment but this might be convenient. I think `uri` is a better name for this either way just to indicate what should be in there.

picklenerd:
> * I had to remove the Default impl for `DamsClient` (i.e. `Session`) because there's no default for the tonic client object. We talked about calling the `connect` function to get a tonic client, but because it's async, I wasn't able to await it in the standard Default impl. For now, I removed the session parameter from the rest of the API functions and tests but is there a way for me to make a dummy Default for DamsClient anyway? Can async_trait do anything here?

`Default` is basically just a standard trait-ified way to write a `fn new() -> Self` function. It's rarely useful as a trait implementation unless you're trying to `#[derive(Default)]` something that contains it. It's totally ok to just make `async fn new() -> Self` and call that instead of `DamsClient::default()`.

marsella (commented):
> Is it appropriate for the storage key to be a part of the state for this object? Not to replace it being in persistent storage, but because most of the flows have a "retrieve_storage_key" step that would be nice not to have to duplicate.

I think for now, we should keep it separate to avoid diverging from the spec. I appreciate the efficiency improvements of not re-retrieving it every time -- we are going to revisit storage key retrieval for security improvements in boltlabs-inc/key-mgmt-spec#67, I will update that to also look into client-side caching. I think the ideal outcome from that issue is that we can put the _encrypted_ storage key in the `Client`, so it can be decrypted it any time the client is authenticated and has the export key, but isn't dragged around in the clear.

dams-client/src/api.rs
marsella: I agree. Part of the reason these comments are currently so explicit about contents is because we wrote them when this was just an API and we hadn't filled in what the `Session`/`DamsClient` contained.

dams-client/src/api.rs
vidyaap: Yep, this makes sense, I just forgot to add the RNG to the struct fields.

> For example, whoever calls client.generate_and_store() shouldn't have to pass an RNG.

Also wanted to clarify this part of your comment - are you suggesting making the API functions part of the `DamsClient` instead of separate like they are now? That would make sense to me; it would just change all the current API functions and tests.

dams-client/src/api.rs
```diff
@@ -70,47 +74,74 @@ impl Password {
...
-/// communication session between an asset owner and a key server
-/// that provides mutual authentication, confidentiality, and integrity.
-/// An open `Session` is required to interact with this API.
-///
-/// A session can be ended manually, or it might time out and require
-/// re-authentication (that is, creation of a new [`Session`]).
+/// A `DamsClient` is an abstraction over client operations. It is used
+/// to hold state, common infrastructure, and a `tonic` client object.
+/// Authenticating to the DAMS returns a `DamsClient`, and this object
+/// must be passed to the remaining API functions.
```
marsella, line 80: I'm not quite sure what "an abstraction over client operations" means; I think it's fine here to be more explicit about the functionality this handles. E.g.:
- Are there any assumptions or behaviors we expect out of this type?
- Does this interface manage all communication with the server? If so, particular mention that it handles confidentiality, integrity, and authentication of communication with the server (and maybe drop "must be passed to remaining API functions" -- that's implied if it handles all the comms).
- maybe describe the lifetime of a client. I think it handles a single authenticated session, and doesn't persist past one request / the end of authentication?

dams-client/src/api.rs
```diff
@@ -70,47 +74,74 @@ impl Password {
...
+/// Options for actions the client can take.
+pub(crate) enum ClientAction {
+    Register,
+    Authenticate,
 }
 
 #[allow(unused)]
-impl Session {
+impl DamsClient {
+    /// Connect to the gRPC client and return it to the client app.
```
marsella, line 99: Does this actually form a connection with the server? Be explicit about it -- I think it opens an initial connection with the server but doesn't authenticate the client, right?

marsella (commented):
I have a couple comments about documentation -- I think it's worth trying to be more explicit about the purpose of & assumptions around the `DamsClient` and its fields so that we don't get them fuzzed up later.

I also saw a bunch of TODO #172s -- are you planning to fix those in this PR? If not, I would make a new issue describing the work that needs to be done and switch the tag to that issue so we don't lose it later.

vidyaap:
> I have a couple comments about documentation -- I think it's worth trying to be more explicit about the purpose of & assumptions around the `DamsClient` and its fields so that we don't get them fuzzed up later.

Edited those docs a bit; let me know if that covers what you were looking for!

> I also saw a bunch of TODO #172s -- are you planning to fix those in this PR? If not, I would make a new issue describing the work that needs to be done and switch the tag to that issue so we don't lose it later.

Those TODOs had to do with the `Default` issue in my first comment. I just deleted them because @picklenerd and I decided to remove the API function tests (dams-tests/test/client.rs) completely. They didn't work with this client refactor and would have to be rethought anyway, but let us know if you want to discuss further!

dams-client/Cargo.toml
```diff
@@ -15,6 +15,7 @@ async-trait = "0"
 bincode = "1"
 dams = { version = "0.1.0", path = "../dams" }
 futures = "0"
+http = "0"
```
picklenerd, line 18: note: `0.x` crates should specify a minor version. I made a ticket #192 to fix these so don't worry about this one.

dams-client/src/api/authenticate.rs
```diff
@@ -1,54 +1,43 @@
...
-        user_id,
-        password,
-        client_login_start_result,
-        server_start_result,
-    )
-    .await?;
-
-    Ok(client_login_finish_result.session_key.into())
+impl DamsClient {
+    pub(crate) async fn handle_authentication<T: CryptoRng + RngCore>(
```
picklenerd, line 16: Nice :grin:

picklenerd (approved):
This looks great

PR #184, opened 2022-08-26
Fix user ID / account name distinction

### LINKED ISSUE #177: Create separate `UserId` and `AccountName` types
Create two separate types for (internal) user id and (external / opaque) account name. Try to replace current user ids with the appropriate one (I think right now it's `AccountName`).

See spec, especially the changes in boltlabs-inc/key-mgmt-spec#99 for additional context.

### LINKED ISSUE #178: Update database to hold both types of user id, account name
With the changes in boltlabs-inc/key-mgmt-spec#99, the database needs to maintain a mapping between a unique user ID and a unique account name (e.g. email address).

- [ ] Update user model to hold both of these
- [ ] Update "create new record" to take both of these and reject if non-unique
- [ ] Update registration function to create user id at the appropriate time.

### LINKED ISSUE #179: Update authentication to handle separate user id, account name
I'm not sure if there are any changes that won't be captured by #177, but if there are, log them here and then do them.

Refer to boltlabs-inc/key-mgmt-spec#99 for possible changes.

marsella:
This has elements of #131, plus closes #177, closes #178, closes #179, closes boltlabs-inc/key-mgmt-spec#102 (parent epic).

I unblocked user ID authentication by making #186 and sending the user id unauthenticated for now, but documenting aggressively around it that it'll need to be updated once the underlying message authentication mechanisms are in place.

Question about databases -- where should we enforce uniqueness of user id and account name? Should the calling module check-then-insert, or should the db insert function fail if one of those fields already exists?

Thought for the future: generate / store / retrieve all take a User ID but at test-description time (compile time?) we don't know what that is -- we only know the account name we chose. We'll have to add some infrastructure to persist the user ids that are returned from register / authenticate and use them in subsequent operations (or not use them if we want to test e.g. authenticating as X and trying to generate as Y fails).


dams/src/opaque_storage.rs
```diff
@@ -99,16 +55,15 @@ pub fn create_or_retrieve_server_key_opaque(
...
     use std::{
         env::temp_dir,
         net::{IpAddr, Ipv4Addr},
-        str::FromStr,
+        path::Path,
     };
 
     #[tokio::test]
-    async fn test_store_and_retrieve() -> Result<(), Error> {
+    async fn opaque_server_key_is_retrievable() {
```
marsella, line 66: The `temp_dir` stuff here makes this kind of annoying to reproduce -- if the test fails, it doesn't get removed, so then the first assertion will always fail and you have to go manually delete it. I don't think it's worth fixing though because we will probably relocate this code to mongodb soon anyway.

dams-client/src/api.rs
```diff
@@ -103,14 +110,17 @@ impl Session {
...
+        account_name: &AccountName,
         password: &Password,
     ) -> Result<Self, SessionError> {
-        let result = Self::authenticate(client, rng, user_id, password).await;
+        let result = Self::authenticate(client, rng, account_name, password).await;
         match result {
-            Ok(result) => {
+            Ok(session_key) => {
+                // TODO: receive User ID over authenticated channel (under session_key)
+                let user_id = todo!();
```
marsella, line 120: Here's the problem: this needs to be received & the MAC checked under the `SessionKey`. We don't have authentication in place yet, of course, but I wasn't sure if there was a correct API to use. I imagine the API will usually be via the `Session` anyway, which won't exist yet.

I could send it un-authenticated with the success indicator for now, but that's a security issue that would _need_ to be fixed before this goes anywhere. Or we could "manually" authenticate it on the server side?

dams/src/opaque_storage.rs
picklenerd: I can work on a PR for this today.

dams-client/src/api.rs
```diff
@@ -34,7 +34,8 @@ impl ToString for Password {
 }
 
 impl FromStr for Password {
-    type Err = DamsClientError;
+    //type Err = DamsClientError;
+    type Err = Infallible;
```
marsella, line 38: Which one of these is better? `Infallible` is a better description of the error, but there isn't a specific `DamsClientError` variant that absorbs it. I'm actually a bit confused about why this compiles.

dams-key-server/src/command/register.rs
```diff
@@ -84,12 +86,26 @@ async fn register_finish(
...
+        server_registration,
+    )
+    .await?;
 
     // reply with the success:true if successful
-    let reply = server::RegisterFinish { success: true };
+    let reply = server::RegisterFinish {
+        success: true,
+        user_id,
+    };
```
marsella, line 108: @marsella Check spec to see if the user id is actually supposed to be sent here.

dams-key-server/src/database/user.rs
```diff
@@ -17,22 +17,24 @@ use opaque_ke::ServerRegistration;
 
 /// Create a new [`User`] with their authentication information and insert it
 /// into the MongoDB database.
+///
+/// TODO: add constraint that user id and account name are both unique.
```
marsella, line 21: Resolve question in description and remove this TODO before merging.

dams-tests/tests/main.rs
```diff
@@ -203,14 +213,21 @@ impl Test {
...
-                    .map_err(|e| e.into()),
+                // Register with the given account name.
+                Register(account_name, password) => {
+                    Session::register(client, rng, account_name, password)
+                        .await
+                        .map(|_| ())
+                        .map_err(|e| e.into())
+                }
+                // Authenticate with the saved user id, if one exists
+                // TODO #179: This should take an account name, not use a previously saved user id.
```
marsella, line 224: Remove TODO

dams/src/user.rs
```diff
@@ -39,23 +63,26 @@ impl UserId {
...
+            account_name,
             secrets: Vec::new(),
             server_registration,
         }
     }
 
-    pub fn into_server_registration(self) -> ServerRegistration<OpaqueCipherSuite> {
-        self.server_registration
+    pub fn into_parts(self) -> (ServerRegistration<OpaqueCipherSuite>, UserId) {
+        (self.server_registration, self.user_id)
```
marsella, line 86: Is there a more standard way to destructure a type that doesn't have public fields?

dams-client/src/api.rs
picklenerd: `Infallible` is for when you don't have an error type but need to specify one. We have `DamsClientError` now so we can specify it and it will let us use `?` in any function that returns `DamsClientError`. I agree that it's more descriptive for this particular error but I don't think the caller needs to know that it's infallible. They should treat it like it can return error. I think we risk having the caller `.unwrap()` safely and then changing the conversion down the line so that it's actually possible to fail.

Any error type would compile here since it never actually returns an error.

dams/src/user.rs
picklenerd: Nope this is how I always see it done. Do the fields need to be private?

marsella: Yeah, I think so. It's a public type but the fields shouldn't be changed by an arbitrary party. We only call this when we're done with the `User` itself, so it's not like we're cloning the whole thing just to get a copy of these.

dams-key-server/src/database/user.rs
marsella: @picklenerd or @vidyaap Can you comment on this, please? More details in the PR description.

dams-client/src/api.rs
marsella: Resolved by #186

dams-key-server/src/database/user.rs
vidyaap: I think it would be best to check the uniqueness constraint in the database `create_user` function itself.  If we anticipate still wanting to find a user by their user ID, it might be worth separating the `find` function into a `find_by_user_id` and `find_by_account_name` and just use those to check in `create_user`.

marsella: I found the [Unique Index](https://www.mongodb.com/docs/manual/core/index-unique/) feature on Mongo, but I'm not sure where I would enforce it -- I think Vidya said something about Mongo automatically instantiating databases, so there's no initialization code? I'm not even sure where we even throw the `RegistrationFailed` error in the "register the same user twice" test; is there already a unique index enforced on the account name?

marsella: Ah, thanks. That sounds good to me!

vidyaap: I tried to figure out how we could use the unique index with the way we define the models and [this post](https://stackoverflow.com/questions/71284576/rust-mongodb-set-unique-fields) was what I saw people doing. In our code that would look like putting something like this at the end of the database connection function (dams-key-server/src/database.rs, after line 27):

```
  let options = IndexOptions::builder().unique(true).build();
  let model = IndexModel::builder()
      .keys(doc! {"user_id": 1, "account_name": 1})
      .options(options)
      .build();
  db.collection::<User>(constants::USERS)
      .create_index(model, None)
      .await
      .expect("error creating index!");
```

I still have to test whether this will allow us to avoid manually checking though.

dams-key-server/src/command/register.rs
```diff
@@ -84,9 +86,20 @@ async fn register_finish(
...
+        UserId::new(rng.deref_mut())
+    };
+
     // add the new user to the DB
-    let _object_id =
-        User::create_user(&context.db, &finish_message.user_id, server_registration).await?;
+    let _object_id = User::create_user(
+        &context.db,
+        &user_id,
+        &finish_message.account_name,
```
marsella: @indomitableSwan The server doesn't make sure that the account name passed in this message is the same as the one passed in the first message of OPAQUE and used to create the server registration (e.g. in `register_start` and `register_finish`). Should we manually verify that? We do check that it's fresh but I think we could get an inconsistency in the server DB. Or, we could just save the account name after the first message and not require the client to pass it a second time.

dams-key-server/src/database/user.rs
marsella: Yeah, I think I prefer this solution because it avoids an unlikely (?) race condition where another task adds a user with the same account name between when we check that the things are fresh when we actually add the user.

In this branch that would go on L24, I think. I will add it and you can double-check the location.

vidyaap: Sounds good! A possible concern I thought of after though: we should make sure that adding this won't overwrite the existing User collection every time we connect. I don't think that's happening but I wasn't able to test it with the way the integration tests are now.

marsella: Ooh, good point. Maybe I can try to write a unit test for that.

picklenerd: We can add a few random characters to the end of usernames in the test cases so that they're unique even if someone repeats names.

picklenerd: I like this uniqueness constraint for the general case too. This seems the most natural solution.

marsella: Okay, I tried to use the unique index but I was getting really inconsistent test failures. I wrote up the problem in #197 and rolled back the changes in [3c95b4c](https://github.com/boltlabs-inc/key-mgmt/pull/184/commits/3c95b4ca00cc040d66ea464cb0e37295b64fcb72).

The current status is that there is manual verification that name + id are fresh, but not at time of insertion (just at the specified point in the protocol).

marsella: Also: I wrote a unit test to check that the user collection doesn't get overwritten when we reconnect. It worked locally but failed in the CI with
```
Error: MongoDb(Error { kind: ServerSelection { message: "Server selection timeout: No available servers. 
Topology: { Type: Unknown, Servers: [ { Address: localhost:27017, Type: Unknown, 
Error: Connection refused (os error 111) }, ] }" }, labels: {}, wire_version: None, source: None })
```
Since it doesn't really pertain to this PR, I took it out and tagged the commit with #197, which maybe isn't quite the right place to take care of it, but I didn't want to lose it.

dams/src/opaque_storage.rs
picklenerd: #191

dams-client/src/client.rs
```diff
@@ -123,19 +134,24 @@ impl DamsClient {
...
-    /// Output: If successful, returns a [`DamsClient`].
+    /// Output: Returns Ok if successful. To perform further operations, use
+    /// [`Self::authenticated_client()`].
     pub async fn register(
-        user_id: &UserId,
+        account_name: &AccountName,
         password: &Password,
         config: &Config,
-    ) -> Result<Self, DamsClientError> {
+    ) -> Result<(), DamsClientError> {
```
marsella, line 143: @indomitableSwan Point of interest: I changed this to not return a DamsClient so that we will adhere to the spec and prevent a user from using a "register" session for other operations.

dams-key-server/src/database/user.rs
jwantz: Just throwing in my two cents here... I believe that any time that we know a given column that is stored in the DB contains a unique value, then it should have a unique constraint. In the code snippet that was referenced within this comment ([here](https://github.com/boltlabs-inc/key-mgmt/pull/184/files#discussion_r958852157)), please keep in mind that this was creating a compound constraint on two columns (meaning, the combination of those two columns would be unique, but not necessarily individually).

marsella: Yes, I realized the compound constraint when my tests kept failing! There's a two-unique-constraints version in the rollback commit. I think there's also more work to do on efficiency -- right now we're re-creating the index every time we get a database handle, but it seems to me like we should only need to do that once.

indomitableSwan (commented):
LGTM. High-level responses below.

> I unblocked user ID authentication by making https://github.com/boltlabs-inc/key-mgmt/issues/186 and sending the user id unauthenticated for now, but documenting aggressively around it that it'll need to be updated once the underlying message authentication mechanisms are in place.

OK. I added "security" and "cryptography" labels to #186.

> Question about databases -- where should we enforce uniqueness of user id and account name? Should the calling module check-then-insert, or should the db insert function fail if one of those fields already exists?

This check should be enforced atomically with insertion.

> Thought for the future: generate / store / retrieve all take a User ID but at test-description time (compile time?) we don't know what that is -- we only know the account name we chose. We'll have to add some infrastructure to persist the user ids that are returned from register / authenticate and use them in subsequent operations (or not use them if we want to test e.g. authenticating as X and trying to generate as Y fails).

Agreed. The tests we have are incomplete. Is there an issue for this?

dams-client/src/client.rs
vidyaap: I'm a little confused by how this change works. Does this mean that the calling app will have to call `authenticated_client()` with the same parameters right after `register()` to get a client? And if so, why do we still call `authenticate()` at the end of register()? Also is there no state we need to keep from register() in the DamsClient?

dams-key-server/src/command/register.rs
marsella: Nope, the parameter we pass the second time doesn't get used in an OPAQUE function -- we just put it into the database. I made the change to not pass it a second time.

dams-client/src/client.rs
marsella: > Does this mean that the calling app will have to call authenticated_client() with the same parameters right after register() to get a client? 

Yes.

> And if so, why do we still call authenticate() at the end of register()? 

This is a good question that I am actually not sure of the answer to. @indomitableSwan 

> Also is there no state we need to keep from register() in the DamsClient?

No, there's nothing that is output from `register` that isn't also output from `authenticate`.

marsella:
> > Thought for the future: generate / store / retrieve all take a User ID but at test-description time (compile time?) we don't know what that is -- we only know the account name we chose. We'll have to add some infrastructure to persist the user ids that are returned from register / authenticate and use them in subsequent operations (or not use them if we want to test e.g. authenticating as X and trying to generate as Y fails).
> 
> Agreed. The tests we have are incomplete. Is there an issue for this?

No, no issue. Vidya pointed out that we keep the `user_id` in the `DamsClient`, so as long as we persist the client we'll have the ID. I don't thing this needs a separate issue, I just wanted to note it for whoever starts implementing protocol flows and writing integration tests for them.

indomitableSwan:
> > > Thought for the future: generate / store / retrieve all take a User ID but at test-description time (compile time?) we don't know what that is -- we only know the account name we chose. We'll have to add some infrastructure to persist the user ids that are returned from register / authenticate and use them in subsequent operations (or not use them if we want to test e.g. authenticating as X and trying to generate as Y fails).
> > 
> > 
> > Agreed. The tests we have are incomplete. Is there an issue for this?
> 
> No, no issue. Vidya pointed out that we keep the `user_id` in the `DamsClient`, so as long as we persist the client we'll have the ID. I don't thing this needs a separate issue, I just wanted to note it for whoever starts implementing protocol flows and writing integration tests for them.

Oh I see, I misread this.

indomitableSwan (approved):

PR #188, opened 2022-08-29
Add FromStr impl to Configs #189

picklenerd:
Remove generation of config files from tests

### DIFF
--- a/dams-tests/tests/common.rs
+++ b/dams-tests/tests/common.rs
@@ -1,9 +1,8 @@
 use std::{
-    collections::HashMap,
     fs::{self, File},
-    io::{Read, Write},
-    path::Path,
+    io::Read,
     process::Command,
+    str::FromStr,
     sync::Mutex,
 };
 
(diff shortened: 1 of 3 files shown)


picklenerd:
todo: Make an issue and add it to the commit

dams-tests/tests/common.rs
```diff
@@ -43,6 +40,11 @@ impl Party {
 
 #[allow(unused)]
```
marsella, line 41: This is not on you but we can probably take out this flag now :)

marsella (approved):
This looks good to me. I thought you'd have to take out the `#[non_exhaustive]` tag on the `Config`s, but maybe that's only to prevent manual construction of a Config, rather than construction from a string?

dams-tests/tests/common.rs
picklenerd: `db` isn't used so it still returned an error but I changed it to `_db` so we could avoid the the `allow`. I got rid of the other unnecessary `#[allow(unused)]` blocks too.

PR #193, opened 2022-08-30
Add minor versions to 0.x deps #192

### LINKED ISSUE #192: Add minor version numbers to all `0.x` crates in Cargo.toml
A lot of Rust crates treat minor versions in `0.x` crates as if they were major versions and will include breaking changes with minor version updates. We have some dependencies pinned to `0` that should be updated to include the minor version.

picklenerd:
Closes #192


vidyaap (approved):
Thanks, looks good to me!

PR #196, opened 2022-08-31
Clean server state before integration tests #191

### LINKED ISSUE #191: Clean server state before integration tests
Some test failures leave state behind that can end up being used during subsequent test runs. For example, it's possible for tests to fail and leave without dropping the database on the local machine. We should ensure that the server state is fresh for every test run.

- [ ] Move filesystem and database cleanup to the `setup` step in `common.rs`
    - This has the side benefit that the server state can be inspected after tests run
- [ ] Reset the database before every test case
- [ ] Implement a test state abstraction that can be used to set up the database before each test case

picklenerd:
Moved file and db cleanup to `setup`

Closes #191

### DIFF
--- a/dams-tests/tests/common.rs
+++ b/dams-tests/tests/common.rs
@@ -1,6 +1,7 @@
 use std::{
     fs::{self, File},
     io::Read,
+    path::Path,
     process::Command,
     str::FromStr,
     sync::Mutex,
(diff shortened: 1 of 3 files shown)


marsella (approved):
One clean-up comment, otherwise this looks good!

Issue #107, opened 2022-08-03
Select and integrate AEAD library


marsella:
The cryptography module depends on an authenticated encryption with associated data (AEAD) scheme.

Also, think about context for each encryption scheme and validation of the type (e.g. at deserialization or generation).

- [x] choose a library that implements an AEAD scheme (or chose to implement it ourselves)
- [x] Update encryption-related types to be wrappers around the expected chacha types (key, associated data, encrypted)
- [x] Implement generic encryption + decryption
- [x] implement encryption + decryption of the storage key under the master key
- [x] implement encryption + decryption of a secret under the storage key
- [x] updated all the tagged documentation in the crypto module to reference the selected scheme
- [x] note choice via boltlabs-inc/key-mgmt-spec#70

marsella:
[RFC 5116](https://www.rfc-editor.org/rfc/rfc5116.html) has recommended AEAD schemes based on AES, and links to papers with the full algorithms and security proofs.

indomitableSwan:
Can you link to the relevant issue in key-mgmt-spec? @marsella

indomitableSwan:
In line with our other dependency decisions, it would be reasonable to choose [Rust Crypto's chacha20poly1305 implementation](https://github.com/RustCrypto/AEADs/tree/master/chacha20poly1305).

This repository has at least undergone one audit by NCC.

The other option I might consider is a nonce misuse-resistant scheme like AES-GCM-SIV ([RFC 8452](https://www.rfc-editor.org/rfc/rfc8452.html)), but RustCrypto's implementation of this has not been audited and there are performance drawbacks to choosing this option.

PR #202, opened 2022-09-01
Fill in high-level cryptography for registration

### LINKED ISSUE #131: Implement public cryptography for registration

marsella:
Closes #131, addresses part of #107.

- from 107: adds encryption and decryption of storage key
- from 131: adds utility function to create-and-encrypt storage key

Other features
- Switches bound for things-that-can-be-encrypted to require `TryFrom<Vec<u8>>` instead of `From`.
- Implements conversion to + from bytes for secrets, storage key
- Adds tests for storage key generation, encryption+decryption, and conversions


dams/src/crypto.rs
```diff
@@ -153,7 +166,10 @@ where
 
     /// Decrypt the ciphertext to a `T`.
     ///
-    /// Raises a [`CryptoError::DecryptionFailed`] if decryption fails.
+    /// Raises a [`CryptoError::DecryptionFailed`] if decryption fails or
+    /// [`CryptoError::ConversionError`] if the decrypted plaintext cannot be
+    /// converted into `T`.
```
marsella, line 171: Is it okay to reveal the difference between these two errors, or should it always be a `DecryptionError`?

dams/src/crypto.rs
```diff
@@ -351,19 +408,43 @@ pub struct Secret {
...
 
+impl From<Secret> for Vec<u8> {
+    fn from(secret: Secret) -> Self {
+        // len || material || context
+        iter::once(secret.material.len() as u8)
+            .chain(secret.material)
+            .chain::<Vec<u8>>(secret.associated_data.into())
+            .collect()
+    }
+}
```
marsella, line 432: This is the conversion we use to turn a secret into bytes before encrypting it.

@indomitableSwan: Should the associated data be included in the "encrypted" block, or should this only be key material?

@picklenerd: This (and the converse below) is starting to feel suspiciously like something that should be handled by serde. Depending on Swan's answer, I wonder if a better option here is to either require serde on encrypted types or make a separate trait that documents that the bytes should only include "the things that should be encrypted".

dams/src/crypto.rs
picklenerd: From an engineering perspective, this would be a useful distinction for the caller. It seems like it would be safe to send as long as there aren't any details about T.

dams/src/crypto.rs
picklenerd: Is the idea that the message types would have `Secret` types stored in them? If so, I agree that `serde` would be convenient.

dams/src/crypto.rs
```diff
@@ -351,19 +408,43 @@ pub struct Secret {
...
+
+impl TryFrom<Vec<u8>> for Secret {
+    type Error = CryptoError;
+    fn try_from(bytes: Vec<u8>) -> Result<Self, Self::Error> {
+        // len (1 byte) || material (len bytes) || context (remainder)
+        let len = bytes[0] as usize;
```
picklenerd, line 438:
```suggestion
        let len = bytes.get(0).ok_or(SomeError)? as usize;
```

This can panic.

dams/src/crypto.rs
```diff
@@ -351,19 +408,43 @@ pub struct Secret {
...
+        let (material, associated_data) = bytes.split_at(len + 1);
+
+        Ok(Self {
+            material: material[1..].into(),
```
picklenerd, line 442: This can panic too. I don't actually know the safe method for ranges.

dams/src/crypto.rs
marsella: No, not messages themselves. When we encrypt a type (e.g. `Secret`), we convert to bytes, then feed the bytes to the generic encryption function. For the `encrypt` function, I put in the type constraints that any `T` you want to encrypt has to be converted to and from `Vec<u8>`. But now I'm wondering that if the `T`s aren't going to be fully serialized (e.g. we only encrypt one of the fields of a secret, not both of them), then the constraint that `Vec<u8>: From<T>` is not super descriptive. It doesn't clarify that the `Vec<u8>` should only hold the bits of `T` that you want to encrypt.

Anyway, I feel like I'm over-optimizing at this point, since I'm the one implementing all these conversions. Maybe refining those constraints is an enhancement for later.

picklenerd: If it's not going in the message types, I don't think we need to deal with `serde`. I would only use that if we wanted to convert the types to/from a transport or storage format.

dams/src/crypto.rs
marsella: Ah, thank you. In fact, `split_at` can also panic, so every line of this function was broken!!

dams/src/crypto.rs
```diff
@@ -277,8 +314,17 @@ struct MasterKey(EncryptionKey);
...
-        todo!()
+    fn encrypt_storage_key(
+        self,
+        rng: &mut (impl CryptoRng + RngCore),
+        storage_key: StorageKey,
+        user_id: &UserId,
+    ) -> Result<Encrypted<StorageKey>, CryptoError> {
+        let associated_data = AssociatedData::new()
+            .with_str(&user_id.to_string())
+            .with_str("storage key");
```
marsella, line 325: @indomitableSwan I feel confused about what we're supposed to do with the associated data from the storage key itself. Right now it's just getting encrypted and decrypted as part of the secret.

indomitableSwan: Yes, that's correct. Technically, the spec doesn't include any context for the storage key. (I think I commented on this below as well.)

When I originally read your question, I misread and thought you were asking about the associated data and why we include it ---which enables the client to check that the blob it's decrypting really is their storage key.

indomitableSwan: i.e. for hardening, the protocol implementations really ought to be checking properties like this. Client-side checks are [currently missing in the specification](https://github.com/boltlabs-inc/key-mgmt-spec/issues/120).

dams/src/crypto.rs
```diff
@@ -290,14 +336,15 @@ impl MasterKey {
...
+pub struct StorageKey(Secret);
 
 #[allow(unused)]
 impl StorageKey {
     /// Generate a new 32-byte [`StorageKey`].
-    fn generate() -> Self {
-        todo!()
+    fn generate<Rng: CryptoRng + RngCore>(rng: &mut Rng) -> Self {
+        let associated_data = AssociatedData::new().with_str("storage key");
+        Self(Secret::generate(rng, 32, associated_data))
```
indomitableSwan, line 347: Technically this isn't in the spec, but I don't have a problem with this redundancy. I would expect versioning information here in future versions.

dams/src/crypto.rs
indomitableSwan: > This is the conversion we use to turn a secret into bytes before encrypting it.
> 
> @indomitableSwan: Should the associated data be included in the "encrypted" block, or should this only be key material?
> 
> @picklenerd: This (and the converse below) is starting to feel suspiciously like something that should be handled by serde. Depending on Swan's answer, I wonder if a better option here is to either require serde on encrypted types or make a separate trait that documents that the bytes should only include "the things that should be encrypted".

You encrypt everything. You want the client to be able to store and retrieve the entire key, which includes all fields (i.e., both the secret material and the context). I'm starting to suspect naming this field `associated_data` instead of `context` (in `Secret`) is causing you to be confused in several places. 

I'm also seeing in the spec that the phrase "non-sensitive" [got slipped into the description of "context"](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/system-functionalities.md#generate-a-secret-helper). I'm not sure when that happened (probably me), but it's misleading. The context can be anything and it's considered _part of the secret object itself_. 

Let me know if there are any other changes to the spec that would make this clearer.

dams/src/crypto.rs
```diff
@@ -351,28 +412,61 @@ pub struct Secret {
...
-            material: std::iter::repeat_with(|| rng.gen()).take(len).collect(),
+            material: iter::repeat_with(|| rng.gen()).take(len).collect(),
             associated_data,
         }
     }
 }
 
+impl From<Secret> for Vec<u8> {
+    fn from(secret: Secret) -> Self {
+        // len || material || context
```
indomitableSwan, line 430: we might want to be explicit about the length of the context.

dams/src/crypto.rs
```diff
@@ -387,13 +481,44 @@ mod test {
...
 
-            // make sure converting to & from Vec<u8> gives the same result
+            // Make sure converting to & from Vec<u8> gives the same result
             assert_eq!(data, output_data);
         }
 
         Ok(())
     }
 
+    #[test]
```
indomitableSwan, line 491: You might want to test that leaving the optional context out doesn't cause any problems.

dams/src/crypto.rs
```diff
@@ -439,17 +564,20 @@ mod test {
...
-    fn encrypt_storage_key_not_implemented() {
+    fn storage_key_generation_produces_unique_storage_keys() {
         let mut rng = rand::thread_rng();
-        let master_key = MasterKey(EncryptionKey::new(&mut rng));
-        let _encrypted_storage_key = master_key.encrypt_storage_key(StorageKey);
+        let mut uniq = HashSet::new();
+
+        assert!((0..1000)
+            .map(|_| StorageKey::generate(&mut rng))
+            .all(|storage_key| uniq.insert(storage_key.0.material)));
```
indomitableSwan, line 573: what are we actually testing here? It seems like you're checking that the rng isn't broken.

dams/src/crypto.rs
```diff
@@ -472,7 +602,7 @@ mod test {
         // already exists in the set.
         assert!((0..1000)
             .map(|_| Secret::generate(&mut rng, 32, AssociatedData::default()).material)
-            .all(|nonce| uniq.insert(nonce)))
+            .all(|secret| uniq.insert(secret)))
```
indomitableSwan, line 605: Again, are we actually just testing the rng here?

indomitableSwan (commented):
- I responded to your questions inline.
- Hardening suggestion on the `Secret` abstraction.
- Minor testing suggestion.
- I am confused by a couple of your tests and would like clarification.

Otherwise this looks OK.

dams/src/crypto.rs
marsella: That seems reasonable. I can add.

dams/src/crypto.rs
marsella: Yeah, basically. I mean, it's a general test that storage key generation incorporates some randomness (e.g. we actually use the RNG that gets passed).

dams/src/crypto.rs
marsella: Yeah, testing that we use the RNG in secret generation.

dams/src/crypto.rs
marsella: Since all of the secrets we currently use in the protocol do have context, that field is not optional right now.

indomitableSwan: Hmm. I would like us to follow the spec.

marsella: Made #205 to address this.

dams/src/crypto.rs
marsella: Oh yeah, that's a good point. I think this will get "disappeared" into a `<Protocol>Failed` error when it gets passed to the calling application.

dams/src/crypto.rs
marsella: Thanks! I will keep this as-is. I also renamed the secret field to `context` as you suggest.

dams/src/crypto.rs
marsella: Made a note of the storage key not needing context in #205. Yeah, we might add some of those property checks in this module (e.g. pass some extra parameters like user id so you don't just decrypt, you decrypt-and-verify), but I am going to hold off on adding them right now, if that's okay.

indomitableSwan (approved):
LGTM.

PR #206, opened 2022-09-06
197 db uniqueness

### LINKED ISSUE #197: Improve database uniqueness enforcement
As of #186, the key server database holds a `UserId` and an `AccountName` for each user. These fields must both be unique. During the register flow, the server manually checks this with the `find_user` and `find_user_by_id` methods. If both of these pass, then they add the user to the database.

However, this opens up a possible race condition where a user with the same id / name gets added between the time of the uniqueness check and the database update. We can prevent this by using a MongoDB index that enforces uniqueness of those fields.

@marsella tried to implement this during the course of #178 but ran into a problem where the tests had inconsistent failing behavior because of the uniqueness requirement, even after dropping the entire DB at the start of each test. This sometimes happened in the new unit tests for the database and sometimes in the integration tests. After rebasing, it sometimes didn't happen at all.

@jwantz pointed out
> If we are generating the `UserId` under the hood, then why not let the DB do it for us as part of the insert? (My understanding is that the `user_id` would be the primary key for the table.) Then we really only need to worry about the `account_name`.

- [ ] Decide whether to use an index / at-time-of-insert uniqueness guarantee
- [ ] Fix the tests

vidyaap:
Closes #197 

Wraps the UserId in a Uuid so it doesn't serialize to an array and we can use the MongoDB unique index on it.

Question: Is there a way to make the `DatabaseSpec::new()` function I wrote in `dams` only available to tests? It didn't seem to work when I put `cfg[test]` over it, and I think it's because the unit tests I'm using it for are in a different crate.

### DIFF
--- a/dams-key-server/src/database/user.rs
+++ b/dams-key-server/src/database/user.rs
@@ -112,7 +112,10 @@ mod test {
         let mut rng = rand::thread_rng();
         let mongodb_uri = "mongodb://localhost:27017";
         let db_name = "multiple_connections_dont_overwrite";
-        let db_spec = DatabaseSpec::new(mongodb_uri.to_string(), db_name.to_string());
+        let db_spec = DatabaseSpec {
+            mongodb_uri: mongodb_uri.to_string(),
+            db_name: db_name.to_string(),
+        };
 
         // Clean up previous runs and make fresh connection
         drop_db(mongodb_uri, db_name).await?;
@@ -165,7 +168,10 @@ mod test {
         let mut rng = rand::thread_rng();
         let mongodb_uri = "mongodb://localhost:27017";
         let db_name = "unique_indices_are_enforced";
-        let db_spec = DatabaseSpec::new(mongodb_uri.to_string(), db_name.to_string());
+        let db_spec = DatabaseSpec {
+            mongodb_uri: mongodb_uri.to_string(),
+            db_name: db_name.to_string(),
+        };
 
         // Clean up previous runs and make fresh connection
         drop_db(mongodb_uri, db_name).await?;
--- a/dams-key-server/src/database/user.rs
+++ b/dams-key-server/src/database/user.rs
@@ -68,6 +68,8 @@ mod test {
 
     use super::create_user;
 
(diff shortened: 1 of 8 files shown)


dams-key-server/src/database/user.rs
```diff
@@ -51,3 +49,156 @@ pub async fn find_user_by_id(db: &Database, user_id: &UserId) -> Result<Option<U
...
+    use opaque_ke::{
+        ClientRegistration, ClientRegistrationFinishParameters, ServerRegistration, ServerSetup,
+    };
+    use rand::{CryptoRng, RngCore};
+
+    use crate::{constants, database::connect_to_mongo, DamsServerError};
+
+    use super::create_user;
+
+    fn server_registration(
```
marsella, line 71: Could you add a short comment here to explain that this locally simulates OPAQUE to get a valid `ServerRegistration`?

dams-key-server/src/database/user.rs
```diff
@@ -51,3 +49,156 @@ pub async fn find_user_by_id(db: &Database, user_id: &UserId) -> Result<Option<U
...
+        // Matching AccountNames can't be added.
+        let different_uid = UserId::new(&mut rng)?;
+        assert!(
+            create_user(&db, &different_uid, &account_name, &server_registration)
+                .await
+                .is_err()
+        );
+
+        // Matching both can't be added.
+        eprintln!("one conn:: no match both");
```
marsella, line 195: nit: remove stray debug statement

dams/src/user.rs
```diff
@@ -29,7 +30,8 @@ impl UserId {
         let mut id = [0_u8; 16];
         rng.try_fill(&mut id)
             .map_err(|_| CryptoError::RandomNumberGeneratorFailed)?;
-        Ok(Self(Box::new(id)))
+        let uuid = Uuid::from_bytes(id);
+        Ok(Self(uuid))
```
marsella, line 34: Perfect! I'm glad this worked!

marsella (commented):
A couple nits. I don't know the answer to your question about test helper scoping.

picklenerd (approved):

PR #207, opened 2022-09-06
Add generate cryptography

### LINKED ISSUE #107: Select and integrate AEAD library

### LINKED ISSUE #134: Implement public cryptography for generate-and-store flow

marsella:
Closes #134, closes #107 (checks the last box!)

- Updates several type representations to be more compatible with how they need to operate
   - Storage key is now an encryption key, not a secret. This also required updating the to/from `Vec<u8>` on a storage key. For modularity, I also implemented the conversion for an encryption key.
   - Associated data is now a byte array, not a string.
   - The opaque encryption key is now a byte array, not a generic array, and it has a constructor.
- Adds encryption and decryption of secrets under storage keys
- Adds generation of key IDs.

Specific feedback on tests is particularly welcome.

### DIFF
--- a/dams/src/crypto.rs
+++ b/dams/src/crypto.rs
@@ -18,7 +18,7 @@ use generic_array::{
 use hkdf::Hkdf;
 use rand::{CryptoRng, Rng, RngCore};
 use serde::{Deserialize, Serialize};
-use sha3::Sha3_256;
+use sha3::{Digest, Sha3_256};
 use std::{iter, marker::PhantomData};
 use thiserror::Error;
 use tracing::error;
@@ -386,16 +386,40 @@ impl TryFrom<Vec<u8>> for StorageKey {
 
 /// Universally unique identifier for a secret.
 #[allow(unused)]
-#[derive(Debug, Serialize, Deserialize)]
-pub struct KeyId;
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct KeyId(Box<[u8; 32]>);
 
 #[allow(unused)]
 impl KeyId {
     /// Generate a new, random `KeyId` for the given [`UserId`].
     ///
     /// This is called by the key server.
-    fn generate(rng: impl CryptoRng + RngCore, user_id: UserId) -> Self {
-        todo!()
+    fn generate(rng: &mut (impl CryptoRng + RngCore), user_id: &UserId) -> Self {
+        const RANDOM_LEN: usize = 32;
+        let mut randomness = [0; RANDOM_LEN];
+        rng.fill_bytes(&mut randomness);
+        let domain_separator = b"Lock Keeper key ID";
+
+        let mut hasher = Sha3_256::new();
+        let bytes = hasher
+            .chain_update(domain_separator.len().to_be_bytes())
+            .chain_update(domain_separator)
+            .chain_update([user_id.len() as u8])
+            .chain_update(user_id.as_bytes())
+            .chain_update([RANDOM_LEN as u8])
+            .chain_update(randomness)
+            .finalize();
+
+        // Truncate to 32 bytes, then convert into a 32-byte array.
+        // This `unwrap` is safe because the lengths are hardcoded.
+        Self(Box::new(
+            bytes
+                .into_iter()
+                .take(32)
+                .collect::<Vec<u8>>()
+                .try_into()
+                .unwrap(),
+        ))
     }
 }
 
@@ -660,10 +684,34 @@ mod test {
     }
 
     #[test]
-    #[should_panic(expected = "not yet implemented")]
-    fn key_id_generation_not_implemented() {
-        let thread_rng = rand::thread_rng();
-        let _key_id = KeyId::generate(thread_rng, UserId::default());
+    fn key_id_generation_produces_unique_ids() {
+        let mut rng = rand::thread_rng();
+        let mut uniq = HashSet::new();
+        let user_id = UserId::new(&mut rng).unwrap();
+
+        // Create 1000 key IDs; make sure they're unique
+        // by putting them into a set. Insert will return false if a secret
+        // already exists in the set.
+        assert!((0..1000)
+            .map(|_| KeyId::generate(&mut rng, &user_id))
+            .all(|key_id| uniq.insert(key_id)))
+    }
+
+    #[test]
+    fn key_id_generation_incorporates_user_id() {
+        let mut rng = rand::thread_rng();
+        let mut uniq = HashSet::new();
+        let seed = b"a terrible seed for testing keys";
+
+        // The bad RNG will produce the same randomness for every call to generate.
+        // But the key IDs are still unique!
+        assert!((0..1000)
+            .map(|_| {
+                let user_id = UserId::new(&mut rng).unwrap();
+                let mut bad_rng = StdRng::from_seed(*seed);
+                KeyId::generate(&mut bad_rng, &user_id)
+            })
+            .all(|key_id| uniq.insert(key_id)))
     }
 
     /// Generates random bytes, encrypts them, and returns them along with the
@@ -814,6 +862,8 @@ mod test {
     fn create_and_encrypt_secret_not_implemented() {
         let mut rng = rand::thread_rng();
         let storage_key = StorageKey::generate(&mut rng);
-        let _ = storage_key.create_and_encrypt_secret(&mut rng, &UserId::default(), &KeyId);
+        // TODO #134: use proper gen function
+        let key_id = KeyId(Box::new([0; 32]));
+        let _ = storage_key.create_and_encrypt_secret(&mut rng, &UserId::default(), &key_id);
     }
 }
--- a/dams/src/crypto.rs
+++ b/dams/src/crypto.rs
@@ -4,9 +4,6 @@
 //! transformations between them. Public functions here are mostly wrappers
 //! around multiple low-level cryptographic steps.
 
-#[cfg(test)]
-use std::convert::Infallible;
-
 use chacha20poly1305::{
     aead::{Aead, Payload},
     AeadCore, ChaCha20Poly1305, KeyInit,
@@ -24,6 +21,9 @@ use thiserror::Error;
 use tracing::error;
 
 use crate::user::UserId;
+
+#[cfg(test)]
+use std::convert::Infallible;
 
 /// Errors that arise in the cryptography module.
 ///
@@ -394,10 +394,14 @@ impl KeyId {
     /// Generate a new, random `KeyId` for the given [`UserId`].
     ///
     /// This is called by the key server.
-    fn generate(rng: &mut (impl CryptoRng + RngCore), user_id: &UserId) -> Self {
+    fn generate(
+        rng: &mut (impl CryptoRng + RngCore),
+        user_id: &UserId,
+    ) -> Result<Self, CryptoError> {
         const RANDOM_LEN: usize = 32;
         let mut randomness = [0; RANDOM_LEN];
-        rng.fill_bytes(&mut randomness);
+        rng.try_fill(&mut randomness)
+            .map_err(|_| CryptoError::RandomNumberGeneratorFailed)?;
         let domain_separator = b"Lock Keeper key ID";
 
         let mut hasher = Sha3_256::new();
@@ -411,15 +415,16 @@ impl KeyId {
             .finalize();
 
         // Truncate to 32 bytes, then convert into a 32-byte array.
-        // This `unwrap` is safe because the lengths are hardcoded.
-        Self(Box::new(
+        // This should never produce a ConversoinError because the lengths are
+        // hard-coded.
+        Ok(Self(Box::new(
             bytes
                 .into_iter()
                 .take(32)
                 .collect::<Vec<u8>>()
                 .try_into()
-                .unwrap(),
-        ))
+                .map_err(|_| CryptoError::ConversionError)?,
+        )))
     }
 }
 
@@ -684,7 +689,7 @@ mod test {
     }
 
     #[test]
-    fn key_id_generation_produces_unique_ids() {
+    fn key_id_generation_produces_unique_ids() -> Result<(), CryptoError> {
         let mut rng = rand::thread_rng();
         let mut uniq = HashSet::new();
         let user_id = UserId::new(&mut rng).unwrap();
@@ -693,8 +698,10 @@ mod test {
         // by putting them into a set. Insert will return false if a secret
         // already exists in the set.
         assert!((0..1000)
-            .map(|_| KeyId::generate(&mut rng, &user_id))
-            .all(|key_id| uniq.insert(key_id)))
+            .map(|_| KeyId::generate(&mut rng, &user_id).unwrap())
+            .all(|key_id| uniq.insert(key_id)));
+
+        Ok(())
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


dams/src/crypto.rs
```diff
@@ -85,7 +83,11 @@ impl AssociatedData {
     }
 
     fn with_str(self, ad: &str) -> Self {
-        AssociatedData(format!("{}\n{}", self.0, ad))
+        AssociatedData(self.0.into_iter().chain(ad.bytes().into_iter()).collect())
```
picklenerd, line 86: I dunno if this would actually be better but I think you could use the `write!` macro here.

https://doc.rust-lang.org/std/macro.write.html

dams/src/crypto.rs
```diff
@@ -299,6 +301,12 @@ impl OpaqueExportKey {
     }
 }
 
+impl From<GenericArray<u8, U32>> for OpaqueExportKey {
+    fn from(arr: GenericArray<u8, U32>) -> Self {
+        Self(Box::new(arr.into()))
+    }
+}
+
```
picklenerd, line 309: Nice timing! I was just about to write this myself for registration.

dams/src/crypto.rs
```diff
@@ -359,43 +386,138 @@ impl StorageKey {
...
+        // len (should always be 32) || key || context
+        let len = *bytes.first().ok_or(CryptoError::ConversionError)? as usize;
+        if len != 32 {
+            return Err(CryptoError::ConversionError);
+        }
+
+        // len (1 byte) || material (len bytes) || context (remainder)
+        let key = bytes.get(1..33).ok_or(CryptoError::ConversionError)?;
+        let associated_data: Vec<u8> = bytes
+            .get(len + 1..)
```
picklenerd, line 460: This syntax is a little confusing. I think it would help to make something like `let offset = len + 1` and then do `.get(offset..)`.

dams/src/crypto.rs
```diff
@@ -247,7 +250,7 @@ impl From<GenericArray<u8, U64>> for OpaqueSessionKey {
 /// and corresponding registration result.
 #[allow(unused)]
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct OpaqueExportKey(GenericArray<u8, U32>);
+pub struct OpaqueExportKey(Box<[u8; 32]>);
```
picklenerd, line 253: Does this need to be in a `Box`? I tried removing it and the tests still passed.

marsella: Not necessarily. On a previous project, I used lots of unboxed, fixed-length arrays and we ended up with a stack overflow. I've been defaulting to `Box`es to avoid that happening unexpectedly.

dams/src/crypto.rs
```diff
@@ -652,18 +797,43 @@ mod test {
...
+    }
+
+    #[test]
+    fn key_id_generation_incorporates_user_id() {
+        let mut rng = rand::thread_rng();
+        let mut uniq = HashSet::new();
+        let seed = b"a terrible seed for testing keys";
+
+        // The bad RNG will produce the same randomness for every call to generate.
+        // But the key IDs are still unique!
```
picklenerd, line 829: This is great 😄

dams/src/crypto.rs
picklenerd: That sounds like a good reason. I can see that potentially being a problem on the remote client.

dams/src/user.rs
```diff
@@ -35,6 +29,25 @@ impl UserId {
...
+    type IntoIter = IntoIter<u8, 16>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.into_iter()
+    }
+}
+
+impl Display for UserId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(&format!("{:?}", self.0))
```
picklenerd, line 49:
```suggestion
        write!(f, "{:?}", self.0)
```

picklenerd: If you still want this to print as a string, you could use this:
https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy

It'll replace any invalid UTF-8 bytes with a specific symbol and print everything else correctly.

dams/src/crypto.rs
```diff
@@ -359,43 +386,138 @@ impl StorageKey {
...
+            .chain_update(domain_separator.len().to_be_bytes())
+            .chain_update(domain_separator)
+            .chain_update([user_id.len() as u8])
+            .chain_update(user_id.as_bytes())
+            .chain_update([RANDOM_LEN as u8])
+            .chain_update(randomness)
+            .finalize();
+
+        // Truncate to 32 bytes, then convert into a 32-byte array.
+        // This should never produce a ConversoinError because the lengths are
```
indomitableSwan, line 511: typo in `ConversionError`

dams/src/transaction.rs
```diff
@@ -31,18 +31,6 @@ pub struct TransactionApprovalRequest {
...
-        Self {
-            key_id: KeyId,
-            user_id: UserId::default(),
-            asset_id: AssetId,
-            tar_id: TarId,
-            transaction: Transaction,
-        }
-    }
-}
-
```
indomitableSwan, line 45: why are you changing anything in transactions.rs in this PR?

indomitableSwan (changes requested):
Looks OK, modulo some minor comments. 
- The abstractions seem a bit heavy-handed to me, but they are probably fine. The only place I suggest we might want to simplify now is with respect to `StorageKey`: I don't see why we have`encrypt_data` as a helper function. Seems like extra code that isn't particularly helpful.
- There are some naming inconsistencies/weirdness we might want to consider updating, either here or in a cleanup issue:
  - For `EncryptionKey`, we might want to consider the same change we made for `Secret`, where we use `context` as the field name for the context instead of `associated_data`.
  - It's a bit odd that to encrypt a secret we use `encrypt_data`, but when we later decrypt it's `decrypt_secret`.
- It also looks like some changes that do not belong in this PR were made, specifically with respect to transactions.rs. These should be reverted and the file left unchanged.

dams/src/transaction.rs
marsella: `KeyId` is no longer a unit struct, so we'd have to implement a default constructor for `KeyId` in order to create this default `TransactionApprovalRequest`. I don't think that would be a useful or safe thing to have around, and we're not using this default impl anywhere, so I decided to remove this rather than add an unsafe constructor for `KeyId`.

dams/src/user.rs
marsella: Ah, I just moved this around in the file. But since you mentioned it:

We only use this for the database lookup -- `find_user_by_id()` uses the `doc!` macro that seems to require a string as input. I'm not sure if the lossy version would be OK for that, or if this version even works if there's a non-UTF character in there. I was actually hoping to remove it in favor of the `IntoIterator` since that's all I need for making associated data, but I couldn't tell if the `doc` macro could take non-string parameters. Maybe @vidyaap knows?

dams/src/crypto.rs
```diff
@@ -105,7 +107,7 @@ pub struct Encrypted<T> {
 }
 
 /// A well-formed symmetric encryption key for an AEAD scheme.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
```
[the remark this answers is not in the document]
marsella: I'll change it here, I think that's a good call.

dams/src/crypto.rs
```diff
@@ -359,43 +386,138 @@ impl StorageKey {
...
         rng: &mut (impl CryptoRng + RngCore),
         user_id: &UserId,
         key_id: &KeyId,
-    ) -> Encrypted<Secret> {
-        todo!()
+    ) -> Result<Encrypted<Secret>, CryptoError> {
+        let ad = AssociatedData::new()
+            .with_bytes(user_id.clone())
+            .with_bytes(key_id.clone())
+            .with_str("client-generated");
```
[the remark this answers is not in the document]
marsella: No, it is not particularly useful. I will combine. But for the record, one of the associated datas is the context for the generated secret, and one is the associated data for the encryption. They're technically, I think, separate things in the spec, even though they have the same value.

dams/src/user.rs
vidyaap: For the `doc` macro, it needs to be either something that implements `Into<Bson>` and `Clone`, or a `str` or bytes. Not sure if that is helpful for this though.

dams/src/crypto.rs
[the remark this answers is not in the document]
marsella: It does make the tests nicer so we can get the secret that was generated instead of losing it in the "generate-and-encrypt" functionality. I'm using cloned RNGs right now to get around that (e.g. generate the secret manually with an RNG using the same state), but that's kind of fragile anyway -- if we change around the order of things in the function, we have to change the tests.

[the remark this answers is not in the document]
indomitableSwan: > No, it is not particularly useful. I will combine. But for the record, one of the associated datas is the context for the generated secret, and one is the associated data for the encryption. They're technically, I think, separate things in the spec, even though they have the same value.

I was just thinking about this and realized I forgot to write down a comment. (I probably shouldn't pause during reviews.) It's a problem that `create_and_encrypt` doesn't return the actual generated secret, because the secret is meant to be passed to local storage (which right now the PoC spec just leaves this as a stub that stores the secret in the clear).

I think it's an odd choice to not pass the secret's `context` in directly as the associated data for encryption. If we decided at some point to make them different (which we won't, because they are the same value in the spec as part of the design, I just didn't give them a new name because that is an implementation detail and the spec is not pseudocode), then we could adjust the code as needed. Is there a way that I can make this more explicit in the spec repo? If we _want_ the spec repo to be pseudocode and dictate things like "pass the secret's context field into the AEAD", then I really need more help writing. To date, I haven't been trying to design the implementation, just write down the protocol.

dams/src/transaction.rs
indomitableSwan: Ah I see. This makes sense. Thank you for explaining.

dams/src/crypto.rs
picklenerd, line 86:
```suggestion
        AssociatedData(self.0.into_iter().chain(ad.bytes()).collect())
```
I think the `Bytes` type that `bytes()` returns is already an iterator.

dams/src/user.rs
marsella: Hmm, the `as_bytes` method isn't compiling and this isn't causing any errors right now, so I think I will leave it as-is for now, if that's okay with @picklenerd.

dams/src/crypto.rs
marsella: Good call. I actually switched this to call into the `with_bytes` method for consistency.

dams/src/crypto.rs
marsella: Fixed! Both here and in the other bytes-parsing functions.

dams/src/crypto.rs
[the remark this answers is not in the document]
indomitableSwan: Ah yes. That does make sense! It's a bit hard to say whether it's premature optimization, because certainly careless changes here could result in problems, as you point out. However, I can think of a couple different ways we might want to handle arbitrary contexts for secrets (which we don't support at the moment), some of which might be sensitive values, so perhaps we should decide on that abstraction when we are adding that feature, rather than now. If we're thinking about it now, I would probably adjust the secret to have a `default` context and an `arbitrary` context, and then just use the `default` directly in encryption for the associated data.

I'll also observe that in the protocol here we do not send the associated data over to the server for storage, because the server already knows these values. That is perhaps easy to misunderstand, and now that you've mentioned the issue, I am not clear if our encryption abstraction lines up to this or not. I also think it doesn't matter here, i.e., we could change the spec without any problems. We would just have to be sure that the key server checks the received associated data for correctness before storage (or otherwise doesn't rely on the content of the data for determining association with `user_id` and `key_id` in its database).

dams/src/crypto.rs
```diff
@@ -354,32 +346,15 @@ impl StorageKey {
...
-        key_id: &KeyId,
-    ) -> Result<Encrypted<Secret>, CryptoError> {
-        let ad = AssociatedData::new()
-            .with_bytes(user_id.clone())
-            .with_bytes(key_id.clone())
-            .with_str("client-generated");
-
-        Encrypted::encrypt(rng, &self.0, secret, &ad)
-    }
-
```
indomitableSwan, line 355: Given that we have created `StorageKey` as a wrapper around `EncryptionKey`, perhaps we should provide basic `encrypt` and `decrypt` (non-public) inner methods that are just wrappers around our encryption abstraction? I'm not sure what the argument are for and against this.

indomitableSwan, line 355: Given that we have created `StorageKey` as a wrapper around `EncryptionKey`, perhaps we should provide basic `encrypt` and `decrypt` (non-public) inner methods that are just wrappers around our encryption abstraction? I'm not sure what the argument are for and against this.

dams/src/crypto.rs
```diff
@@ -391,14 +366,17 @@ impl StorageKey {
...
+        let context = AssociatedData::new()
             .with_bytes(user_id.clone())
             .with_bytes(key_id.clone())
             .with_str("client-generated");
-        let secret = Secret::generate(rng, 32, ad);
+        let secret = Secret::generate(rng, 32, context.clone());
 
-        self.encrypt_data(rng, secret, user_id, key_id)
+        Ok((
+            secret.clone(),
```
indomitableSwan, line 377: why do we have to clone the secret? i.e., when we made our encryption abstraction, do we want to pass the secret in directly or a reference to the secret?

indomitableSwan (approved):
I am satisfied with the changes, but have a couple more questions/suggestions, and if you are making more changes based on these, please flag me again for another pass.

dams/src/crypto.rs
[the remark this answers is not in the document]
marsella: I added boltlabs-inc/key-mgmt-spec#123 to address your second concern.

dams/src/crypto.rs
marsella: The `encrypt` function requires a secret passed by value. Made #209 to reduce this.

dams/src/crypto.rs
marsella: I agree, but I would rather do this once we are ready to abstract more things with the same `StorageKey`.

PR #211, opened 2022-09-07
Add validation to storage key decryption

### LINKED ISSUE #133: Implement client-side cryptography for `retrieve_storage_key`
In several places throughout the protocol, the client will have to retrieve their encrypted storage key from the server. The functionality is [defined in the spec](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/cryptographic_flows.md#retrieve_storage_key-functionality), but might be modified by boltlabs-inc/key-mgmt-spec#67. 

Based on progress of the spec modification, implement the client side cryptography for storage key retrieval. Maybe this means implementing the old version and making a new issue to refactor when #67 is complete, or maybe it means blocking on #67. The implementation should be contained in the `crypto/client.rs::decrypt_storage_key` function and encompass the linked spec section, steps 3.i - 3.iii.

marsella:
Closes #133.

The issue said we'd have to implement the whole decryption block, but we already did that when testing for #107. Instead, this just adds a validation step that the associated data is what was expected, and a couple of tests trying to craft malicious input.

### DIFF
--- a/dams/src/crypto.rs
+++ b/dams/src/crypto.rs
@@ -211,7 +211,16 @@ impl Encrypted<StorageKey> {
     pub fn decrypt_storage_key(
         self,
         export_key: OpaqueExportKey,
+        user_id: &UserId,
     ) -> Result<StorageKey, CryptoError> {
+        // Check that the associated data is correct.
+        let expected_aad = AssociatedData::new()
+            .with_bytes(user_id.clone())
+            .with_str(StorageKey::domain_separator());
+        if self.associated_data != expected_aad {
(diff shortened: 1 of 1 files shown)


indomitableSwan (approved):
This looks good. It makes me ponder our session-handling choices thus far, though. Ideally, perhaps we shouldn't be passing the export key around, but instead just the derived keys we need during an open session (in this case, the `master_key`). Then if a session is compromised after establishment (e.g., cache-timing attack), the export key itself is not compromised. Tagging @jwantz on this thought.

picklenerd (approved):

PR #214, opened 2022-09-09
Pull generic crypto types into submodule

### LINKED ISSUE #230: Refactor crypto module
Right now, all cryptographic operations are contained in one massive module. Ideally they would be separated in a way that would support future development, especially as we start to add more complex secret / key types. 

Separate the underlying pure cryptography from application-specific types and wrappers. Make an abstraction over the Secret type, in particular.

marsella:
This pulls the "generic" cryptography into a submodule, so the only types available to the rest of the crate are our application-specific types. Most changes are highlighted inline (the diff is not particularly useful). I am waiting to rebase this PR so that those inline comments don't get dropped from the diff.

The only thing not noted inline is that I added a `Secret` wrapper around the generic `Secret`. These have the same name for now because I didn't want to cause reverberating changes across the rest of the library right now. 

Fixes #230


dams/src/crypto.rs
```diff
@@ -969,20 +481,18 @@ mod test {
...
-        let fake_storage_key: Encrypted<StorageKey> = Encrypted {
-            ciphertext: encrypted_key.ciphertext,
-            associated_data: encrypted_key.associated_data,
-            nonce: encrypted_key.nonce,
-            original_type: PhantomData,
-        };
+
+        // Serialize the fake key, and pretend it's a storage key when you deserialize. 
+        let fake_storage_key: Encrypted<StorageKey> =
+            serde_json::from_str(&serde_json::to_string(&encrypted_fake_key).unwrap()).unwrap();
```
marsella, line 495: Updated this test to the more realistic deserialize-into-wrong-type, rather than hand-craft the wrong type. This does make me wonder what the `PhantomData` is actually doing in a normal deserialization, if we can just steamroller over it here.

dams/src/crypto/generic.rs
```diff
@@ -0,0 +1,529 @@
...
+}
+
+impl<'a> From<&'a AssociatedData> for &'a [u8] {
+    fn from(associated_data: &'a AssociatedData) -> Self {
+        associated_data.0.as_ref()
+    }
+}
+
+impl AssociatedData {
+    pub(super) fn new() -> Self {
```
marsella, line 64: Here and elsewhere: almost nothing in this module is `pub` -- it's all `pub(super)`, which I'm hoping means that only the crypto module can access it.

dams/src/crypto/generic.rs
```diff
@@ -0,0 +1,529 @@
...
+    // Get the length of the key. This should always return 32.
+    pub fn len(&self) -> usize {
+        self.key.len()
+    }
+
+    /// Get the context for this key
+    pub(super) fn context(&self) -> &AssociatedData {
+        &self.context
+    }
+}
```
marsella, line 130: Added some utility functions for testing in the crypto module. It wouldn't be terrible if these became generally public, but they didn't need to be right now.

dams/src/crypto/generic.rs
```diff
@@ -0,0 +1,529 @@
...
+
+        Ok(())
+    }
+
+    #[test]
+    fn encryption_keys_are_32_bytes() {
+        let mut rng = rand::thread_rng();
+        let key = EncryptionKey::new(&mut rng);
+        assert_eq!(32, key.len());
+    }
```
marsella, line 386: New test.

dams/src/crypto.rs
```diff
@@ -4,188 +4,20 @@
...
-
-        // Decrypt!
-        Ok(cipher
-            .decrypt(&self.nonce, payload)
-            .map_err(|_| CryptoError::DecryptionFailed)?
-            .try_into()?)
-    }
-}
+use generic::{AssociatedData, EncryptionKey};
+pub use generic::{CryptoError, Encrypted, Secret};
```
marsella, line 20: `Secret` now lives in the `generic` module, which I think paves the way toward making more complicated abstractions for different types of secrets/keys that wrap around the basic type. For now, though, we just "forward" it out to the rest of the crate.

dams/src/crypto/generic.rs
```diff
@@ -0,0 +1,529 @@
...
+        }
+    }
+
+    // Use the given bytes as a symmetric AEAD encryption key.
+    pub(super) fn from_bytes(key_material: [u8; 32], context: AssociatedData) -> Self {
+        Self {
+            key: key_material.into(),
+            context,
+        }
+    }
```
marsella, line 116: Added new constructor to cleanly create keys in the `crypto` module.

dams/src/crypto.rs
```diff
@@ -352,13 +184,7 @@ impl StorageKey {
...
     /// Generate a new 32-byte [`StorageKey`].
     fn generate<Rng: CryptoRng + RngCore>(rng: &mut Rng) -> Self {
-        let context = AssociatedData::new().with_str(Self::domain_separator());
-        let key = Secret::generate(rng, 32, context.clone());
-
-        Self(EncryptionKey {
-            key: *chacha20poly1305::Key::from_slice(&key.material),
-            context,
-        })
+        Self(EncryptionKey::new(rng))
```
marsella, line 187: Replaced the secret generation with the standard new-encryption-key function. This removes the uncalled-for storage key context that I accidentally put in. It calls the chacha random-key-generation function instead of our own, but they do almost exactly the same thing, just generate random bytes from the RNG.

I was going to keep the `Secret` version because that's what the spec says, but I was having trouble with the current abstraction boundaries on guaranteeing that the length is correct without unwrapping something. Secret generation gives you back a `Vec<u8>`, but the `from_bytes` method I made requires a `[u8; 32]`. Open to alternate suggestions here.

dams/src/crypto.rs
picklenerd: I don't think it's factored into the deserialization process since there's no actual data there. I think the `PhantomData` is only useful to tell you what you should expect the data to be.

Maybe it's possible to enforce this with a custom deserializer. I think it might still be challenging since type information doesn't exist at runtime but it might be possible.

dams/src/crypto/generic.rs
picklenerd: As long as it's not re-exported by the `crypto` module this should be true. `pub(super)` might even block re-exports.

dams/src/crypto/generic.rs
```diff
@@ -0,0 +1,530 @@
+use chacha20poly1305::{
+    aead::{Aead, Payload},
+    AeadCore, ChaCha20Poly1305, KeyInit,
+};
+
```
indomitableSwan, line 5: There is an extra line here.

dams/src/crypto.rs
indomitableSwan: This is fine --- just adjust the spec so that it doesn't get confusing later.

indomitableSwan (approved):
LGTM.

Question: In crypto.rs, L102, we have an implementation note that suggests the `OpaqueExportKey` struct is meant to be a wrapper around the opaque_ke's "export key". Is this documented somewhere else? Are we intending to change this struct?

dams/src/crypto.rs
marsella: Hmm. We might have to be careful about how we genericize storage and retrieval once we start supporting multiple types of stored secrets. The client side will have to know what type to decrypt into. So we might not be able to reuse a single "retrieve" function, or we'll have to add some kind of indicator to the `StoredSecret` type.

dams/src/crypto.rs
marsella: Great! Made boltlabs-inc/key-mgmt-spec#128

marsella:
> Question: In crypto.rs, L102, we have an implementation note that suggests the `OpaqueExportKey` struct is meant to be a wrapper around the opaque_ke's "export key". Is this documented somewhere else? Are we intending to change this struct?

Thanks! Good catch, I removed that documentation. The `opaque_ke` type _is_ a 64-byte array, but they use a `GenericArray` instead of a fixed-size one. I think the main role of that doc was to remind myself not to try to do anything fancy there.

dams/src/crypto.rs
picklenerd: Yeah that'll be a little complicated depending on the shape of the generic secrets. Serde gives a lot of options for enums so if the data is all shaped differently it'll be pretty easy, otherwise we can tag the data with a type hint.
https://serde.rs/enum-representations.html#internally-tagged

picklenerd (approved):

PR #215, opened 2022-09-09
Delete user from if CreateStorageKey fails #212

### LINKED ISSUE #212: Clean server state if registration fails after the user is created
Registration is a three step process.

1. OPAQUE registration
2. Authentication
3. Creation of a storage key

The user is inserted into the database during the first step, but the database entry is invalid until the the third step is finished. If the user tries to register again with the same account name, registration will fail.

The simplest solution would be to delete the user from the database if any other step fails. Alternatively, we could detect an incomplete registration and skip ahead to step 2.

picklenerd:
I decided to go with deleting the user here. I think this minimizes the chances that something goes wrong and the OPAQUE registration gets out of sync.

This is another one that needs an integration test that can insert itself into the middle of the process. I think the easiest way to trigger this would be to manually run through the registration process and insert the wrong user id in the `client::SendStorageKey` step. #217

Closes #212

### DIFF
--- a/dams-key-server/src/command/create_storage_key.rs
+++ b/dams-key-server/src/command/create_storage_key.rs
@@ -1,5 +1,5 @@
 use crate::{
-    database::user::{find_user, set_storage_key},
+    database::user::{delete_user, find_user, set_storage_key},
     error::DamsServerError,
     server::Context,
 };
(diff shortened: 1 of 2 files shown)


picklenerd:
For authentication failure, the client knows that it's in the middle of the registration protocol so it could trigger deletion itself, but that's a pretty impactful change to the interface that we definitely need to discuss in detail.

picklenerd:
I made an issue for open questions so I'm going to merge this as-is: #219

Issue #216, opened 2022-09-09
Refactor integration tests to support complex tests


picklenerd:
The integration tests currently do a good job of end to end testing for error flows but they don't support comparing input and output values and they don't support complex test cases that involve looking at the database or breaking the process in the middle of a protocol.

We should maintain the current test harness for end-to-end testing and extend it to support comparing successful results. This will be useful for situations like the generate+retrieve tests where we need to verify that the retrieved secret is the same as the generated secret.

We should introduce a parallel set of tests that perform impure operations that look at the database or tap into the message passing process. This will be useful for situations like registration where we need to trigger an error during the creation of a storage key and verify that the user was deleted.

There will be common setup and teardown code between these two types of tests so we need to pull that out of the end-to-end tests and move it to the `common` module. The implementation of the impure tests that need to be written for existing code can be handled in separate issues.

This issue is complete when:

- [x] Setup and teardown is moved to the `common` module.
- [x] The end-to-end testing  module should be renamed from `main.rs` to something like `end_to_end.rs` to reflect their purpose.
- [ ] End-to-end tests are extended with the ability to compare operation inputs and outputs.

PR #218, opened 2022-09-09
Refactored integration tests #216

picklenerd:
This PR addresses the first two requirements of #216. We can work on the third requirement after the retrieve operation is merged.

- Tests are now divided between end-to-end tests and more general integration tests.
- Setup and teardown code were fully moved to common module.

### DIFF
--- a/dams-tests/tests/common.rs
+++ b/dams-tests/tests/common.rs
@@ -13,40 +13,47 @@ use tokio::{task::JoinHandle, time::Duration};
 use tracing::info_span;
 use tracing_futures::Instrument;
 
-use dams::{defaults::server::LOCAL_SERVER_URI, timeout::WithTimeout, TestLogs};
+use dams::{
+    config::{client::Config as ClientConfig, server::Config as ServerConfig},
+    defaults::server::LOCAL_SERVER_URI,
+    timeout::WithTimeout,
+    TestLogs,
+};
 
 pub const ERROR_FILENAME: &str = "tests/gen/errors.log";
-
 pub const SERVER_ADDRESS: &str = "127.0.0.1";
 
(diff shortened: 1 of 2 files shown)


dams-tests/tests/common.rs
```diff
@@ -13,37 +13,44 @@ use tokio::{task::JoinHandle, time::Duration};
...
+async fn generate_files(db: Database) {
     // Delete data from previous run
     let gen_path = Path::new("tests/gen/");
     // Swallow error if path doesn't exist
     let _ = fs::remove_dir_all(&gen_path);
-    fs::create_dir(&gen_path).expect("Unable to create directory tests/gen");
+    fs::create_dir_all(&gen_path).expect("Unable to create directory tests/gen");
 
     // Ensure that the test DB is fresh
     db.drop().await.expect("Failed to drop database");
```
vidyaap, line 56: nit: Dropping the database feels like it should go in `setup` or `start_server` instead

dams-tests/tests/common.rs
```diff
@@ -54,7 +61,9 @@ pub async fn setup(db: Database, server_config: dams::config::server::Config) ->
         .spawn()
         .expect("Failed to generate new certificates");
     tokio::time::sleep(tokio::time::Duration::from_millis(1000)).await;
+}
 
+async fn start_server(server_config: ServerConfig) -> ServerFuture {
```
vidyaap, line 66: I can't comment on the actual line, but I think the `allow` tag on line 75 is from old code, and can be removed.

vidyaap (commented):
Look good to me! Just two minor nits.

vidyaap (approved):
Thanks, this looks good!

Issue #220, opened 2022-09-12
Add config parameter to disable TLS


picklenerd:
We need a temporary flag to disable TLS while we troubleshoot problems with it. The demo project is having trouble connecting.

PR #221, opened 2022-09-12
Added temp flag to disable TLS #220

picklenerd:
We're temporarily disabling TLS for the demo repo. Integration tests still run with TLS enabled but users running the server themselves have the option to disable it now.

This PR simply adds back the old code and puts it in a new function.


jakinyele (approved):
Looks good

Issue #223, opened 2022-09-12
Create database abstraction


picklenerd:
We're currently calling the database by providing functions that take a `mongodb::Database` as a parameter. This is creating a tight coupling to MongoDB that's spreading across the codebase.

We should wrap the `mongodb::Database` handle with our own type and provide methods for all of the free-functions we're currently using.

- [ ] Create an internal `Database` type that wraps a `mongodb::Database`
- [ ] Move free-functions to methods within the `Database` type
- [ ] Provide a temporary `drop` method that can be used in the integration tests

PR #225, opened 2022-09-12
Create Database abstraction #223

picklenerd:
This PR creates a wrapper around the MongoDB database handle. This gives us an easier interface for working with the database and decouples our code from MongoDB.

### DIFF
--- a/dams-key-server/src/command.rs
+++ b/dams-key-server/src/command.rs
@@ -3,7 +3,7 @@ pub mod create_storage_key;
 pub mod generate;
 pub mod register;
 
-use crate::{database::user as User, server::Context, DamsServerError};
+use crate::{server::Context, DamsServerError};
 use dams::{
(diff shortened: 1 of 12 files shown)


vidyaap (approved):
Thanks, this looks good to me!  

Non-blocking question: is this providing as much independence from Mongo as we thought? I'm noticing all the public code is still Mongo-based, but I think I was imagining something that had a similar structure to the impl of the `DamsRpc` trait in server.rs - i.e. something that just dispatches calls to the right functions. That would be a larger task though; I was just curious if that would be worth doing in the future.

picklenerd:
Yeah you're right that we're not abstracting over Mongo at all, but it at least hides Mongo from any library consumers so we could change the database and anyone calling create_user shouldn't have to change anything.  We'd still have to update create_user to work with the new database.

PR #228, opened 2022-09-13
Retrieve

### LINKED ISSUE #73: Implement the retrieve flow
The [retrieve](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/system-functionalities.md#retrieve-a-secret) flow is a multi-message protocol between client and server.

1. The client checks local storage to see if the desired key is there. (The following steps occur if not)
2. The client sends a retrieval request (type not defined at time of writing), including a user ID and key ID.
3. The key server checks the request, retrieves the appropriate key, and sends to the client
4. The client returns the appropriate value to the calling app based on the context.

- [ ] Define client and server message types and serialization into / from tonic messages.
- [ ] Implement client-side protocol
- [ ] Implement server-side protocol
- [ ] Integration tests

vidyaap:
Closes #73 

Aside from regular review I have a few questions:
- @indomitableSwan There is a section in the spec that says to "update the database record to indicate that the secret has been retrieved by the client". Besides logging, was there something else you wanted to happen here?
- Logging
    - I'm having issues logging things that only have one helper function call in the server (i.e. retrieve and retrieve_storage_key) because I can't get the UserId/AccountName out of the result and then call `.log()` on the same result.
    - Is it reasonable/possible to only call `.log` on results that contain an `impl LogIdentifier` and make all the "last" helper functions for each endpoint return a UserId/AccountName in its result so that we can just pull it from there instead of passing it in?
- Testing
    - There's a dummy test in there that just makes sure no errors happen when retrieving, but how can I test that we got the right secret back? I have thoughts, but they might be easier to discuss on call.
    - Right now, I have to generate the secret in the same block as retrieving it so that I have access to the `KeyId` to pass to retrieve. Is there a way we can return things from the execute function to be available to future operations, so that I can separate the generate and retrieve operations?

### DIFF
--- a/dams-client/src/api/arbitrary_secrets.rs
+++ b/dams-client/src/api/arbitrary_secrets.rs
@@ -2,19 +2,12 @@ use crate::{DamsClient, DamsClientError};
 use dams::{
     crypto::{KeyId, Secret, StorageKey},
     types::retrieve_storage_key::{client, server},
-    ClientAction,
+    ClientAction, RetrieveContext,
 };
 use serde::{Deserialize, Serialize};
 
 mod generate;
 mod retrieve;
-
-/// Options for the asset owner's intended use of a secret
-#[derive(Debug, Deserialize, Serialize)]
-pub enum RetrieveContext {
-    LocalOnly,
-    Export,
-}
 
 /// Ways of returning a key from the retrieval process based on usage
 /// [`Context`]
@@ -66,7 +59,7 @@ impl DamsClient {
     pub async fn retrieve(
         &self,
         key_id: &KeyId,
-        context: Option<RetrieveContext>,
+        context: RetrieveContext,
     ) -> Result<RetrieveResult, DamsClientError> {
         let mut client_channel =
             Self::create_channel(&mut self.tonic_client(), ClientAction::Retrieve).await?;
--- a/dams-client/src/api/arbitrary_secrets.rs
+++ b/dams-client/src/api/arbitrary_secrets.rs
@@ -10,7 +10,7 @@ mod generate;
 mod retrieve;
 
 /// Ways of returning a key from the retrieval process based on usage
-/// [`Context`]
+/// [`RetrieveContext`]
 #[derive(Debug, Deserialize, Serialize)]
 pub enum RetrieveResult {
     None,
--- a/dams-client/src/api/arbitrary_secrets.rs
+++ b/dams-client/src/api/arbitrary_secrets.rs
@@ -12,6 +12,12 @@ mod generate;
 const SECRET_LENGTH: u32 = 32;
 
 #[derive(Debug, Deserialize, Serialize)]
+pub enum Context {
+    LocalOnly,
+    Export
+}
+
+#[derive(Debug, Deserialize, Serialize)]
 pub struct LocalStorage {
     pub(crate) secret: Secret,
 }
--- a/dams-client/src/api/arbitrary_secrets.rs
+++ b/dams-client/src/api/arbitrary_secrets.rs
@@ -7,14 +7,22 @@ use dams::{
 use serde::{Deserialize, Serialize};
 
 mod generate;
+mod retrieve;
 
-#[allow(unused)]
-const SECRET_LENGTH: u32 = 32;
-
+/// Options for the asset owner's intended use of a secret
 #[derive(Debug, Deserialize, Serialize)]
 pub enum Context {
     LocalOnly,
-    Export
+    Export,
+}
+
+/// Ways of returning a key from the retrieval process based on usage
+/// [`Context`]
+#[derive(Debug, Deserialize, Serialize)]
+pub enum RetrieveResult {
+    None,
+    ArbitraryKey(LocalStorage),
+    ExportedKey(Vec<u8>),
 }
 
 #[derive(Debug, Deserialize, Serialize)]
@@ -52,5 +60,17 @@ impl DamsClient {
         let mut client_channel =
             Self::create_channel(&mut self.tonic_client(), ClientAction::Generate).await?;
         self.handle_generate(&mut client_channel).await
+    }
+
+    /// Retrieve an arbitrary secret from the key server by [`KeyId`]
+    pub async fn retrieve(
+        &self,
+        key_id: &KeyId,
+        context: Option<Context>,
+    ) -> Result<RetrieveResult, DamsClientError> {
+        let mut client_channel =
+            Self::create_channel(&mut self.tonic_client(), ClientAction::Retrieve).await?;
+        self.handle_retrieve(&mut client_channel, key_id, context)
+            .await
     }
 }
--- a/dams-client/src/api/arbitrary_secrets.rs
+++ b/dams-client/src/api/arbitrary_secrets.rs
@@ -11,7 +11,7 @@ mod retrieve;
 
 /// Options for the asset owner's intended use of a secret
 #[derive(Debug, Deserialize, Serialize)]
-pub enum Context {
+pub enum RetrieveContext {
     LocalOnly,
     Export,
 }
@@ -66,7 +66,7 @@ impl DamsClient {
     pub async fn retrieve(
         &self,
         key_id: &KeyId,
-        context: Option<Context>,
+        context: Option<RetrieveContext>,
     ) -> Result<RetrieveResult, DamsClientError> {
         let mut client_channel =
             Self::create_channel(&mut self.tonic_client(), ClientAction::Retrieve).await?;
--- a/dams-client/src/api/arbitrary_secrets/retrieve.rs
+++ b/dams-client/src/api/arbitrary_secrets/retrieve.rs
@@ -1,11 +1,12 @@
 use crate::{
-    api::arbitrary_secrets::{LocalStorage, RetrieveContext, RetrieveResult},
+    api::arbitrary_secrets::{LocalStorage, RetrieveResult},
     DamsClient, DamsClientError,
 };
 use dams::{
     channel::ClientChannel,
     crypto::KeyId,
     types::retrieve::{client, server},
+    RetrieveContext,
 };
 
 impl DamsClient {
@@ -13,7 +14,7 @@ impl DamsClient {
         &self,
         channel: &mut ClientChannel,
         key_id: &KeyId,
-        context: Option<RetrieveContext>,
+        context: RetrieveContext,
     ) -> Result<RetrieveResult, DamsClientError> {
         // Retrieve the storage key
         let storage_key = self.retrieve_storage_key().await?;
@@ -24,6 +25,7 @@ impl DamsClient {
         let request = client::Request {
             user_id: self.user_id().clone(),
             key_id: key_id.clone(),
+            context: context.clone(),
         };
         channel.send(request).await?;
 
@@ -38,12 +40,12 @@ impl DamsClient {
 
         // Return appropriate value based on Context
         match context {
-            None => Ok(RetrieveResult::None),
-            Some(RetrieveContext::LocalOnly) => {
+            RetrieveContext::Null => Ok(RetrieveResult::None),
+            RetrieveContext::LocalOnly => {
                 let wrapped_secret = LocalStorage { secret };
                 Ok(RetrieveResult::ArbitraryKey(wrapped_secret))
             }
-            Some(RetrieveContext::Export) => Ok(RetrieveResult::ExportedKey(secret.into())),
+            RetrieveContext::Export => Ok(RetrieveResult::ExportedKey(secret.into())),
         }
     }
 }
--- a/dams-client/src/api/arbitrary_secrets/retrieve.rs
+++ b/dams-client/src/api/arbitrary_secrets/retrieve.rs
@@ -1,5 +1,5 @@
 use crate::{
-    api::arbitrary_secrets::{Context, LocalStorage, RetrieveResult},
+    api::arbitrary_secrets::{LocalStorage, RetrieveContext, RetrieveResult},
     DamsClient, DamsClientError,
 };
 use dams::{
@@ -13,7 +13,7 @@ impl DamsClient {
         &self,
         channel: &mut ClientChannel,
         key_id: &KeyId,
-        context: Option<Context>,
+        context: Option<RetrieveContext>,
     ) -> Result<RetrieveResult, DamsClientError> {
         // Retrieve the storage key
         let storage_key = self.retrieve_storage_key().await?;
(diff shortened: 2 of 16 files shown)


dams-client/src/api/arbitrary_secrets.rs
```diff
@@ -7,9 +7,23 @@ use dams::{
...
+    Export,
+}
+
+/// Ways of returning a key from the retrieval process based on usage
+/// [`Context`]
+#[derive(Debug, Deserialize, Serialize)]
+pub enum RetrieveResult {
+    None,
+    ArbitraryKey(LocalStorage),
+    ExportedKey(Vec<u8>),
```
vidyaap, line 25: @indomitableSwan FYI, this is what I was asking about for exported keys. What gets passed back here is the result of [this function](https://github.com/boltlabs-inc/key-mgmt/blob/develop/dams/src/crypto.rs#L531-L541). Is that right?

dams-client/src/api/arbitrary_secrets.rs
```diff
@@ -7,9 +7,23 @@ use dams::{
 use serde::{Deserialize, Serialize};
 
 mod generate;
+mod retrieve;
 
-#[allow(unused)]
-const SECRET_LENGTH: u32 = 32;
+/// Options for the asset owner's intended use of a secret
+#[derive(Debug, Deserialize, Serialize)]
+pub enum Context {
```
picklenerd, line 14: Maybe this could be something like `SecretContext` or `RetrieveContext`.

dams-tests/tests/main.rs
```diff
@@ -296,6 +331,7 @@ enum Operation {
     Register(AccountName, Password),
     Authenticate(AccountName, Password),
     Generate(AccountName, Password),
+    Retrieve(AccountName, Password),
```
marsella, line 334: nit: Maybe rename this to `GenerateAndRetrieve` or something, to capture what it's actually doing right now.

dams-client/src/api/arbitrary_secrets/retrieve.rs
```diff
@@ -0,0 +1,49 @@
...
+        channel.send(request).await?;
+
+        // Get StoredSecret from server
+        let server_response: server::Response = channel.receive().await?;
+
+        // Decrypt secret
+        let secret = server_response
+            .stored_secret
+            .secret
+            .decrypt_secret(storage_key)?;
```
marsella, line 37: thought: I imagine there is some validation here that I didn't really give you the tools to check correctly. Like, we should compare the associated data to make sure it actually contains this user's user id and key id (the spec says to check it for "client-generated" indicator, which all keys have right now). 

@indomitableSwan should I make an issue to check that the associated data has the expected value? We could do it in the `decrypt_secret` method if we pass user and key ids as parameters.

dams-key-server/src/command/retrieve.rs
```diff
@@ -0,0 +1,54 @@
...
+    },
+};
+use tokio_stream::wrappers::ReceiverStream;
+use tonic::{Request, Response};
+
+#[derive(Debug)]
+pub struct Retrieve;
+
+impl Retrieve {
+    pub async fn run<'a>(
```
marsella, line 17: todo: The retrieve action needs to be logged, I think somewhere in this file.

dams/src/types/retrieve.rs
```diff
@@ -0,0 +1,26 @@
+pub mod client {
+    use crate::{crypto::KeyId, impl_message_conversion, user::UserId};
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Debug, Deserialize, Serialize)]
+    /// pass user ID and key ID to server
+    pub struct Request {
+        pub user_id: UserId,
+        pub key_id: KeyId,
+    }
```
marsella, line 10: question (@indomitableSwan): Is the intended use context supposed to be sent to the server? I mean the null / local only / export distinction. It isn't specified in the request description but step 2.ii.e says (on the server side)
> The intended use must match one of the expected options.

indomitableSwan:
> Closes #73
> 
> Aside from regular review I have a few questions:
> 
> * @indomitableSwan There is a section in the spec that says to "update the database record to indicate that the secret has been retrieved by the client". Besides logging, was there something else you wanted to happen here?

We will want to include some kind of indicator flag in the database record. I think the involves an adjustment to the data modelling we are currently using.

marsella (commented):
This looks good! Overall protocol looks good, I noted a few nits that seem slightly off.

You already noted the missing retrieval indicator. This is something we want to have for when we add remotely-generated keys, so that users can tell if the key has ever been taken out of the system. It's just a per-key boolean flag (e.g. it's false until we call retrieve on a key id, then true forever, but doesn't affect any other protocol behavior).

I agree that the tests are not quite capturing what we want right now since we can only test the honest case of generating and immediately retrieving. I'm not sure if there is a nice way to maintain the simple declarative-ness of the tests when subsequent operations depend on data generated by previous operations. I'm thinking about options where the test suite implements some version of local storage -- like, you assign a key a name when you generate it, and then use that name to retrieve it specifically (and the `execute` function maintains the mapping of name -> key id).

indomitableSwan:
> This looks good! Overall protocol looks good, I noted a few nits that seem slightly off.
> 
> You already noted the missing retrieval indicator. This is something we want to have for when we add remotely-generated keys, so that users can tell if the key has ever been taken out of the system. It's just a per-key boolean flag (e.g. it's false until we call retrieve on a key id, then true forever, but doesn't affect any other protocol behavior).
> 
An aside: It's actually for the servers to be able to tell, more than the users.

> I agree that the tests are not quite capturing what we want right now since we can only test the honest case of generating and immediately retrieving. I'm not sure if there is a nice way to maintain the simple declarative-ness of the tests when subsequent operations depend on data generated by previous operations. I'm thinking about options where the test suite implements some version of local storage -- like, you assign a key a name when you generate it, and then use that name to retrieve it specifically (and the `execute` function maintains the mapping of name -> key id).

dams/src/types/retrieve.rs
indomitableSwan: Ah, yes, that should be sent with the request. Sounds like a bug in the spec.

dams-client/src/api/arbitrary_secrets.rs
indomitableSwan: Yes, that looks right to me. @marsella does the spec line up to this, though? I think we may have introduced an inconsistency.

picklenerd:
> There's a dummy test in there that just makes sure no errors happen when retrieving, but how can I test that we got the right secret back? I have thoughts, but they might be easier to discuss on call.

I'm planning to address this with the second PR for #216.  I was just waiting for Retrieve to be finished so I had a test case.

Would it be possible to add a test case that uses an invalid key id?

> Right now, I have to generate the secret in the same block as retrieving it so that I have access to the KeyId to pass to retrieve. Is there a way we can return things from the execute function to be available to future operations, so that I can separate the generate and retrieve operations?

I was thinking of doing this similarly to how @marsella described where you can store a named value on one end and compare the named value on the other. This might get messy but I was planning to experiment a bit.

dams/src/types/retrieve.rs
vidyaap: I don't think I understand what we'd be comparing to on the server-side if we pass it over? Is it just something that would be confirmed by the fact that we can deserialize into an `Option<Context>` type? And if so, should I change the `Context` type to hold the None case too, rather than wrapping the Context in an Option?

picklenerd:
> - I'm having issues logging things that only have one helper function call in the server (i.e. retrieve and retrieve_storage_key) because I can't get the UserId/AccountName out of the result and then call .log() on the same result.
> - Is it reasonable/possible to only call .log on results that contain an impl LogIdentifier and make all the "last" helper functions for each endpoint return a UserId/AccountName in its result so that we can just pull it from there instead of passing it in?

I think we talked about having some kind of logging context where we store the operation type and some other metadata so that it's automatically added to log entries. Maybe the user id can go in there.

This might be a bad long term solution, but we could grab it from the database at the top of each handler and stick it in the logging context. Once we have session management set up (with redis or whatever) we can get the user id when we check the session.

It looks like everything except `Generate` and `RetrieveStorageKey` have an `AccountName` parameter in the first message. If those two could be changed to take an account name, we could add it to the base Message type. If not, we could make a trait.

dams-key-server/src/command/generate.rs
```diff
@@ -24,9 +27,12 @@ impl Generate {
 
         let _ = tokio::spawn(async move {
             // Generate step: receive UserId and reply with new KeyId
-            let key_id = generate(&mut channel, &context).await?;
+            let (key_id, user_id) = generate(&mut channel, &context).await?;
```
picklenerd, line 30: I see what you mean about this being clunky but I think it's fine for now.

vidyaap:
> You already noted the missing retrieval indicator. This is something we want to have for when we add remotely-generated keys, so that users can tell if the key has ever been taken out of the system. ...

Ah, that makes sense. Would it still be accurate to the spec if I added this flag to the `StoredSecret` type instead of the actual secret? I think that would be clearer to read/easier to work with.

dams-client/src/api/arbitrary_secrets.rs
marsella: The spec says to return just the secret, but this function returns the secret and its context.

dams-key-server/src/command/retrieve.rs
vidyaap: Yep, this was the logging issue I was describing in the first comment. I'm trying to figure out how to pass the UserId to the log function from here.

dams/src/types/retrieve.rs
marsella: Yep, there's no comparison on the server side, we'll just check that it deserializes. We can put it in the log, though.

I don't have a strong preference between option or plain context. The `null` option is kind of just for internal use & testing, so it might be easier to feature gate it or remove it entirely if it's an enum variant rather than the whole type, though.

dams-key-server/src/command/retrieve.rs
```diff
@@ -0,0 +1,54 @@
...
+    // Find user by ID
+    let user = User::find_user_by_id(&context.db, &request.user_id)
+        .await?
+        .ok_or(DamsServerError::AccountDoesNotExist)?;
+
+    // Find secret based on key_id
+    let stored_secret = user
+        .secrets
+        .into_iter()
+        .find(|x| x.key_id == request.key_id)
```
picklenerd, line 47: This seems like it should be stored as a map but I don't think that will work if the `KeyId` is bytes since we had trouble indexing with byte arrays. I think we might need to move this filtering to the database query in the future but I think this is fine for now.

marsella:
> Would it still be accurate to the spec if I added this flag to the `StoredSecret` type instead of the actual secret? I think that would be clearer to read/easier to work with.

Yep, that seems like a perfect place to keep it.

dams-key-server/src/command/retrieve.rs
picklenerd: I think you can log it in the `retrieve` function below after you get the first message.

marsella: Oh, sorry! I missed that.

vidyaap:
> It looks like everything except Generate, Retrieve, and RetrieveStorageKey have an AccountName parameter in the first message ...

Yeah that makes sense. The issue I'm having is more that Retrieve and RetrieveStorageKey don't have a second message so I'm not able to get the user_id out of the result, even in the clunky way that I'm doing for generate.

> I think we talked about having some kind of logging context where we store the operation type and some other metadata so that it's automatically added to log entries. Maybe the user id can go in there.

I'm not sure I understand how the context would work, with the `LogExt` being a trait right now.  If we make it a struct, then can we still attach `.log` to results? Also, do we have to keep that structure?

Also, I'm realizing logging in general is incomplete because I don't think we're logging failures on "first" messages, only "last" messages in a sequence.

picklenerd:
> I'm not sure I understand how the context would work, with the `LogExt` being a trait right now. If we make it a struct, then can we still attach `.log` to results? Also, do we have to keep that structure?

I think the easiest way would be to attach the metadata to the server's `Context` struct and pass the whole `Context` to the log function. That would give it a database handle, user id, and operation all in a single parameter.

vidyaap:
> I think the easiest way would be to attach the metadata to the server's Context struct and pass the whole Context to the log function. That would give it a database handle, user id, and operation all in a single parameter.

I thought the server context shouldn't be tied to a single user?

dams-key-server/src/command/retrieve.rs
vidyaap: Re: logging in the retrieve function - I think logging there would only log the result of that one message rather than the whole operation.

dams-key-server/src/command/retrieve.rs
```diff
@@ -0,0 +1,45 @@
...
+impl Retrieve {
+    pub async fn run<'a>(
+        &self,
+        request: Request<tonic::Streaming<Message>>,
+        context: Context,
+    ) -> Result<Response<MessageStream>, DamsServerError> {
+        let (mut channel, rx) = ServerChannel::create(request.into_inner());
+
+        let _ = tokio::spawn(async move {
+            // Generate step: receive UserId and reply with new KeyId
```
marsella, line 25: nit: This doc is not correct.

dams-key-server/src/command/retrieve.rs
picklenerd: I think that's the best we can do for now. I made a ticket to update the logging interface: #232

picklenerd:
> I thought the server context shouldn't be tied to a single user?

A new context gets created for each request so we can inject these parameters when we create it.

dams-key-server/src/command/retrieve.rs
vidyaap: Okay, I logged the last non-channel error operation in retrieve for now, and put a TODO in for #232.

dams-client/src/api/arbitrary_secrets.rs
marsella: Made boltlabs-inc/key-mgmt-spec#127 to remove inconsistency.

dams-key-server/src/command/retrieve.rs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+async fn retrieve(channel: &mut ServerChannel, context: &Context) -> Result<(), DamsServerError> {
+    // Receive UserId from client
+    let request: client::Request = channel.receive().await?;
+    // Find secret based on key_id
+    // TODO #232: move this log so that we log the entire operation
+    let stored_secret = User::get_user_secret(&context.db, &request.user_id, &request.key_id)
+        .await
+        .log(
```
picklenerd, line 43: I think this needs to `create_log_entry` directly since the extension method will only be called if there's an error.

Issue #232, opened 2022-09-13
Improve logging interface


picklenerd:
The current logging interface works well in the middle of a protocol but it's not currently possible to log operations before the first message is passed.

In order to get create a log entry at the beginning of a protocol, we need either the user id or the account name. If we add the account name to the base message type, we would be able to create a log entry before processing the first message. The updated message type would look something like this:

```proto
message Message {
  bytes content = 1;
  string account_name = 2;
}
```

With this change to the message type, we can add the account name and operation type to the server's `Context` type. This will allow us to update the `log` function to take a `Context` as a parameter instead of specifying a bunch of redundant parameters.

marsella:
If we do this, we should also add validation once we process the content to make sure the claimed account name matches the account that's actually getting operated on.

vidyaap:
Note from dev sync - we should look at how to add headers to the tonic RPC messages so that we can pull that metadata instead of relying on the messages in the stream to get the user_id from.

PR #236, opened 2022-09-14
Fix error reporting #229

### LINKED ISSUE #229: Fix error reporting
Error details are currently being swallowed up by some part of the `tower` stack. This makes troubleshooting very difficult.

I suspect that the errors are being swallowed up in the `start_tonic_server` function of `server.rs`. We can start by logging any errors returned in the threads spawned during this function.

## Error examples
- If the server certificate is invalid, it returns obscure connection errors.
- If a user tries to register twice, the client gets a `DamsChannel(NoMessageReceived)` error.

This issue is complete when:

- [x] All errors are logged through the `tracing` crate
- [x] All logged errors contain the full details of the returned error
- [x] Protocol error details should be hidden from the client but should return something more useful than `NoMessageReceived`

picklenerd:
The reason errors were disappearing is because we spawn a lot of `tokio` tasks but drop their `JoinHandle`  which means any error returned by that task will disappear into the ether. 

One way to solve this would be to collect all of these join handles somewhere and have a separate task that loops through them and checks for errors. This might be the right solution in the future but I think it's a little too complex right now and it could also cause its own set of new problems.

Instead, I did some refactoring that puts the logic from these tasks into separate functions so that the task itself can check for errors, log the error, and then decide what to do next.

For the main server loop, I added a `handle_connection` function. This function calls the main `tonic` service so any errors returned by that service are now logged. All of the service functions were moved to their own module to reduce the size of `server.rs`.

For the operations, I added a new `Operation` trait that takes some of the boilerplate from out of the operations themselves and allows them to return errors like usual. If an error is returned, the `Operation` trait will log the error on the server and then send an error across the channel that is appropriate  for a client to see. 

There are a few other small refactoring items that happened along the way:

- `ChannelError` variants were moved into `DamsError` for ease of use.
- The `main_with_cli` function was moved from `server.rs` into a `run` method on the `Cli` type to clean up the server module.
- `RetrieveStorageKey` was given its own module to match the other operations.
- `LogExt` was renamed to `AuditLogExt` and its method was renamed to `audit_log`. I had originally created a similar trait for server logs and renamed this to disambiguate. I ended up deleting the server log trait but I think this is a good change.
- `AuditLogExt` was moved to the `log` module to reduce the size of the error module.
- The `command` module was renamed to `operations` for consistency.

Closes #229


picklenerd:
I made some assumptions about which errors should be sanitized for the client here. There are a couple general themes:

- `tonic` errors are returned to the client.
- Protocol-specific errors are returned to the client.
- Protocol errors that would reveal the existence or non-existence of an `AccountName` or `UserId` are sent as `"Invalid account name"` or `"Invalid user id"`.  This is a middle ground between useful details and exposing too much. It's worth considering what we want to do for these cases.
- All other errors will be sent as `"Internal server error"` with no further detail.

@indomitableSwan @marsella

indomitableSwan:
> I made some assumptions about which errors should be sanitized for the client here. There are a couple general themes:
> 
> * `tonic` errors are returned to the client.
> * Protocol-specific errors are returned to the client.
> * Protocol errors that would reveal the existence or non-existence of an `AccountName` or `UserId` are sent as `"Invalid account name"` or `"Invalid user id"`.  This is a middle ground between useful details and exposing too much. It's worth considering what we want to do for these cases.
> * All other errors will be sent as `"Internal server error"` with no further detail.
> 
> @indomitableSwan @marsella

- I do not think we need to pass the `user_id` out to the calling application at all. Under what circumstance would we want to pass out a `"invalid user id"` message? @marsella?
- The only time the client should ever learn that an account name already exists is during a registration call, and in this case, the error _should_ be passed out to the calling application (for usability). Otherwise, the client should never learn this information from the server in the first place.

indomitableSwan:
Additional thought: It sounds like we need to do some hardening with respect to what the key server is willing to pass back to the client for an error message.

picklenerd:
> I do not think we need to pass the `user_id` out to the calling application at all. Under what circumstance would we want to pass out a `"invalid user id"` message?

Some of the client message types send a user id to the server. These can probably be changed to use an account name. Adding account name to all requests was also suggested for #232 so that would be helpful in general.

picklenerd:
> The only time the client should ever learn that an account name already exists is during a registration call, and in this case, the error _should_ be passed out to the calling application (for usability). Otherwise, the client should never learn this information from the server in the first place.

Right now, all of the requests need the client to specify a user id or account name. When we implement sessions on the server, we can store the user id along with the session key and then we won't have to pass either in messages after authentication anymore.

Account name is also required for authentication.

picklenerd:
> When we implement sessions on the server, we can store the user id along with the session key and then we won't have to pass either in messages after authentication anymore.

This actually isn't true. We still need to pass the account name or some other identifier to pull up the session on the server.

marsella (commented):
Thanks for this refactor, this is a very useful improvement. Your error-sanitizing assumptions seem good to me, but I don't have that much grounding in best practices for this. Perhaps as a part of hardening, we'll want to re-evaluate whether it's fine to show _all_ the protocol errors or if we want to genericize it a bit (since the client application will basically have the option to retry or give up), but I don't think it's an issue to share them all now.

> I do not think we need to pass the user_id out to the calling application at all. Under what circumstance would we want to pass out a "invalid user id" message? @marsella?

Yeah, I think the account name -> user id map happens inside our library (e.g. once we figure out the user id, we store it in the client). I think you can wrap errors from account names _and_ user ids into a single "Invalid account" or "Invalid account name".

> The only time the client should ever learn that an account name already exists is during a registration call, and in this case, the error should be passed out to the calling application (for usability). Otherwise, the client should never learn this information from the server in the first place.

I agree that I think it's fine to reveal "account name already registered", but any other user id / account name errors can be wrapped up into one generic error. In authentication and flows other than registration, the "account name already exists" error isn't relevant (since it _must_ exist for those things to happen).

dams-key-server/src/server/service.rs
```diff
@@ -0,0 +1,115 @@
...
+            // Spawn a task to handle each connection
+            let _ = tokio::spawn(async move {
+                match handle_connection(http, conn, tls_acceptor, svc).await {
+                    Ok(()) => Ok(()),
+                    Err(e) => {
+                        error!("{}", e);
+                        Err(e)
+                    }
+                }
+            });
```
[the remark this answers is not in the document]
picklenerd: Yep exactly. We really don't even need to return the error here since it won't go anywhere but I think it's good future-proofing.

dams-key-server/src/error.rs
```diff
@@ -53,37 +47,27 @@ impl From<opaque_ke::errors::ProtocolError> for DamsServerError {
...
     fn from(error: DamsServerError) -> Status {
-        Status::internal(error.to_string())
-    }
-}
+        match error {
+            // Errors that are safe to return to the client
+            DamsServerError::InvalidAccountName
+            | DamsServerError::InvalidUserId
+            | DamsServerError::StorageKeyAlreadySet
+            | DamsServerError::StorageKeyNotSet
```
vidyaap, line 55: suggestion: I'm not sure that the storage key errors are invalid argument errors. I think aborted or unavailable might be better.

vidyaap (commented):
Looks good to me, barring anything left from the comment threads, and one Status variant comment I had!

dams-key-server/src/error.rs
picklenerd: Good catch. I made it `internal` since it's from the internal state on the server and not a problem with the connection.

picklenerd:
> Yeah, I think the account name -> user id map happens inside our library (e.g. once we figure out the user id, we store it in the client). I think you can wrap errors from account names and user ids into a single "Invalid account" or "Invalid account name".
> I agree that I think it's fine to reveal "account name already registered", but any other user id / account name errors can be wrapped up into one generic error. In authentication and flows other than registration, the "account name already exists" error isn't relevant (since it must exist for those things to happen).

@marsella Updated with these suggestions.

marsella (approved):
Thanks for fixing!

picklenerd:
Made some updates to the client errors based on some of Vidya's feedback. I got rid of the generic `Registration failed` and `Authentication failed` errors since the user can't take any action based on those. I added error variants for account/login issues and we can continue to fill out specific errors as we progress.

indomitableSwan:
> > I do not think we need to pass the `user_id` out to the calling application at all. Under what circumstance would we want to pass out a `"invalid user id"` message?
> 
> Some of the client message types send a user id to the server. These can probably be changed to use an account name. Adding account name to all requests was also suggested for #232 so that would be helpful in general.
> 
> The client is also required to pass a user_id during storage key generation per the spec so we'd need an error for that.

I think we're talking about two different things here. The _calling application_ means code written on top of our library, i.e., what the Forte devs are writing.

PR #238, opened 2022-09-15
216 test refactor

### LINKED ISSUE #216: Refactor integration tests to support complex tests

vidyaap:
Closes #216 

There is more work to do to make this structure a bit more flexible (especially to test incorrect behavior on purpose) but this first pass at keeping state in the `Test` object should let us make more meaningful generate and retrieve tests.

### DIFF
--- a/dams-tests/tests/end_to_end.rs
+++ b/dams-tests/tests/end_to_end.rs
@@ -1,11 +1,13 @@
 mod common;
 
 use common::{get_logs, LogType, Party};
-use Operation::{Authenticate, Generate, GenerateAndRetrieve, Register};
+use Operation::{Authenticate, Generate, Register, Retrieve};
 use Party::{Client, Server};
 
-use dams::{config::client::Config, user::AccountName, RetrieveContext};
-use dams_client::{client::Password, DamsClient, DamsClientError};
+use dams::{config::client::Config, crypto::KeyId, user::AccountName, RetrieveContext};
+use dams_client::{
+    api::arbitrary_secrets::RetrieveResult, client::Password, DamsClient, DamsClientError,
+};
 use rand::{distributions::Alphanumeric, Rng};
 use serde_json::Value;
 use std::{collections::HashMap, fs::OpenOptions, str::FromStr};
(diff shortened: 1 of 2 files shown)


dams-tests/tests/end_to_end.rs
```diff
@@ -66,197 +76,267 @@ pub async fn end_to_end_tests() {
...
+    fn generate_tag() -> String {
+        rand::thread_rng()
+            .sample_iter(&Alphanumeric)
+            .take(7)
+            .map(char::from)
+            .collect()
+    }
+
+    fn new(name: String, operations: Vec<(Operation, Outcome)>) -> Self {
+        let tag = Self::generate_tag();
```
picklenerd, line 255:
```suggestion
    fn new(name: impl Into<String>, operations: Vec<(Operation, Outcome)>) -> Self {
        let name = name.into();
        let tag = Self::generate_tag();
```

picklenerd (approved):
This looks great! I added one little suggestion to make the constructor a little more convenient but it's not a blocker.

PR #240, opened 2022-09-15
Test to verify correct storage key is stored #208

### LINKED ISSUE #208: Add integration test case to verify that the storage key is stored properly in the database
This issue is a leftover from #124.

This test case won't work with the current test harness. We're currently refactoring the integration testing structure so it would be best to work the requirements for this type of test into the refactoring. Once the new testing infrastructure supports this, we can write the test.

picklenerd:
This PR checks a generated storage key against the storage key stored in the database after registration.

This test has to insert itself into the normal registration flow so that it can keep a copy of the storage key it generates and sends to the server. In order to avoid a bunch of duplicate logic, I made the `DamsClient::authenticate` method public. This method is not abusable or easy to misuse so I think it's safe to be public. I added a note that the caller should prefer `DamsClient::authenticate_client`.

Unfortunately I couldn't do the same thing with `handle_registration` since it can be used to create a broken user who doesn't have a storage key. I messed around with feature flags but ultimately I just copied the handler code to into the test module. 

Having written this test, I'm wondering if this sort of test case is even useful. I think the same error conditions can be tested through normal generate and retrieve tests that depend on the storage key being correct.

Closes #208

### DIFF
--- a/dams-client/src/client.rs
+++ b/dams-client/src/client.rs
@@ -65,7 +65,7 @@ pub struct DamsClient {
 /// Connection type used by `DamsRpcClient`.
 /// This would normally be `tonic::transport:Channel` but TLS makes it more
 /// complicated.
-type DamsRpcClientInner = hyper::Client<
+pub type DamsRpcClientInner = hyper::Client<
     HttpsConnector<HttpConnector>,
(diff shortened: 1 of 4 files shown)


picklenerd:
We decided that this test case isn't useful since the generate/retrieve test cases cover the same functionality.

Issue #242, opened 2022-09-19
Add remote generation methods


marsella:
The public method for creating new `Secret`s only works for locally (client-side) generation. 

- [ ] Add a new method for remote generation, per [the spec](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/system-functionalities.md#remote-only-secret-generation-and-storage):
> Runs the generate protocol on input `(32, rng, user_id||key_id||"server-generated")` to get a secret `arbitrary_key`

- [x] Fill in the method for remote generation of signing keys.

marsella:
Remote generation of signing keys is already implemented correctly in #262.

@indomitableSwan, I have a memory that we decided to _not_ support remote generation of arbitrary secrets (because the key server cannot do anything with them other than retrieve them, so it's better to only allow them to be stored encrypted). Is that correct? If so, this issue can be closed.

indomitableSwan:
> Remote generation of signing keys is already implemented correctly in #262.
> 
> @indomitableSwan, I have a memory that we decided to _not_ support remote generation of arbitrary secrets (because the key server cannot do anything with them other than retrieve them, so it's better to only allow them to be stored encrypted). Is that correct? If so, this issue can be closed.

I think that's the correct decision at this point in time.

marsella:
Great. Then I will close this, with a reference that the piece we chose not to implement is in the (closed) PR #310.

Issue #241, opened 2022-09-19
Update create+encrypt signing key


marsella:
Once the signing key is correctly created, update the encryption / decryption methods (under the storage key).

These currently assume a trivial signing key and don't provide any meaningful associated data. However, the spec says there is some associated data, which we should add to encryption and check at decryption.
> Computes `ciphertext = Enc(storage_key, arbitrary_key, user_id||key_id||"client-generated") `

marsella:
I think this was mostly completed under #262. We still need to check the associated data is correct at decryption.

Issue #235, opened 2022-09-14
Implement ECDSA signing


marsella:
Choose an ECDSA library and integrate it into the outline defined in #234.

- [x] Choose ECDSA library
- [x] Update key, message, and signature types to be wrappers around the library-defined types (and update conversions+tests appropriately)
- [x] Update generic signing and verification methods to call the library
- [x] Add correctness tests (sign + verify works, verifying the wrong message doesn't work, signing twice produces different signatures, flipping bits in the signature doesn't work)
- [x] Make sure signing key pair domain separator still matches dependency.
- [x] Update other generation functions (remote, import) to use library.
- [x] Update `Export` conversion functions to put the associated data and key material in the correct places (and add tests)

marsella:
I reviewed available ECDSA-over-secp256k1 libraries. 
It seems to me like the Rust Crypto version is the best option, with the big caveat that it might not work as well in the SGX environment. But it would be compatible with all the other crypto libraries we use, I trust the cryptographers building it, and I think it has the nicest API. 

Would be nice to get @indomitableSwan and @picklenerd to weigh in on this before I fully commit. See more detailed notes below:

[secp256k1](https://docs.rs/secp256k1/latest/secp256k1/)
- This is a wrapper around a C? library by Peter Wuille, a bitcoin core developer. Hosted by the rust bitcoin community.
- No security claims re: audits or constant-time behavior.
- abstractions seem fine. Generating messages isn't super misuse-resistant. We can use their custom hasher or hash things ourselves -- I don't love the idea of having multiple implementations of the same hash function.
- It seems to have a no-std option, which might be helpful when we get into SGX.
- No shared API with, say, an EdDSA implementation.
- 177 crates use it

[RustCrypto k256](https://docs.rs/k256/latest/k256/)
- Pure Rust secp256k1 elliptic curve library with ECDSA sign/verify
- Not independently audited. Probably, but not definitely constant time.
- lower-level but more misuse-resistant API (message are just bytes, hashing seems hidden). Uses the same hash library as us.
- not sure about std
- implements the signature trait that is shared with eddsa over curve 25519 (is that the curve we want?)
- 72 crates use it

[libsecp256k1](https://docs.rs/libsecp256k1/latest/libsecp256k1/)
- Pure rust elliptic curve library (for "fast ECDSA signatures", but not clear if they have any particular optimizations) by parity tech, blockchain infra company
[...]

PR #243, opened 2022-09-19
Add outline of methods for ECDSA keys

### LINKED ISSUE #234: Outline and document ECDSA signing
To ease development of protocol flows, we should implement a non-panicking outline of all the cryptography used in the remote signing flow. 

- [ ] Define (unit) types for signing keys, messages (maybe this is just `Into<Vec<u8>>`), and signatures
- [ ] Implement type conversions (signing key -> `Vec<u8>`, string -> message, others?)
- [ ] Either make a new abstraction or make a new function signature to encrypt and decrypt signing keys under storage keys.
- [ ] Function signatures to sign a message and verify a signature (non-panicking!)

marsella:
Closes #234 

The goal is to provide non-panicking methods for dealing with ECDSA signing keys (including creating locally and remotely, encrypting for storage, and signing and verifying with them).

This rearranges the crypto module a little bit more by pulling out `Secret`s and `SigningKey`s into their own modules along with all the relevant functions (e.g. encrypting and decrypting for storage). Open to feedback about whether this is useful (to see all the relevant methods in one place) or confusing (because e.g. `StorageKey` functionality is spread out among multiple files, even though it all lives in once place in the docs).

I added trivial implementations to most of the methods so they don't panic, and tests that demonstrate that the implementation is trivial. There are issues referencing every method that has an incorrect implementation (#235 for sign, verify, and all the type instantiations, #241 for encrypting under the storage key, #242 for filling in remote generation).


dams/src/crypto.rs
```diff
@@ -15,67 +15,14 @@ use tracing::error;
...
-impl Encrypted<Secret> {
-    /// Decrypt a secret. This should be run as part of the subprotocol to
-    /// retrieve a secret from the server.
-    ///
-    /// This must be run by the client.
-    pub fn decrypt_secret(self, storage_key: StorageKey) -> Result<Secret, DamsError> {
-        let decrypted = self.decrypt(&storage_key.0)?;
-        Ok(decrypted)
-    }
-}
```
marsella, line 51: NB: This is copied without change to `arbitrary_secrets.rs`

dams/src/crypto.rs
```diff
@@ -15,67 +15,14 @@ use tracing::error;
...
-            .with_str(StorageKey::domain_separator());
-        if self.associated_data != expected_aad {
-            return Err(CryptoError::DecryptionFailed.into());
-        }
-
-        let master_key = export_key.derive_master_key()?;
-        let decrypted = self.decrypt(&master_key.0)?;
-        Ok(decrypted)
-    }
-}
```
marsella, line 78: NB: This is copied without change to another place in this file.

dams/src/crypto.rs
```diff
@@ -203,30 +151,6 @@ impl StorageKey {
...
-            .with_bytes(user_id.clone())
-            .with_bytes(key_id.clone())
-            .with_str("client-generated");
-        let secret = Secret(generic::Secret::generate(rng, 32, context.clone()));
-
-        Ok((
-            secret.clone(),
-            Encrypted::encrypt(rng, &self.0, secret, &context)?,
-        ))
-    }
```
marsella, line 229: NB: This is copied without change to `arbitrary_secrets.rs`.

dams/src/crypto.rs
```diff
@@ -518,24 +469,4 @@ mod test {
...
-        let (secret, encrypted_secret) = storage_key
-            .clone()
-            .create_and_encrypt_secret(&mut rng, &user_id, &key_id)?;
-
-        // Decrypt the secret
-        let decrypted_secret = encrypted_secret.decrypt_secret(storage_key)?;
-        assert_eq!(decrypted_secret, secret);
-
-        Ok(())
-    }
```
marsella, line 540: NB: Copied without change to `arbitrary_secrets.rs`.

dams/src/crypto/arbitrary_secret.rs
```diff
@@ -0,0 +1,115 @@
...
+
+            // Convert to Vec<u8> and back
+            let secret_vec: Vec<u8> = secret.clone().into();
+            let output_secret: Secret = secret_vec.try_into()?;
+
+            assert_eq!(secret, output_secret);
+        }
+
+        Ok(())
+    }
```
marsella, line 95: NB: This test is new.

marsella (commented):
Marking the bits of of code about secrets that are copied as-is to the new module.

dams/src/crypto/signing_key.rs
```diff
@@ -0,0 +1,202 @@
...
+        rng: &mut (impl CryptoRng + RngCore),
+        user_id: &UserId,
+        key_id: &KeyId,
+    ) -> Self {
+        SigningKeyPair
+    }
+
+    /// Domain separator for use in serializing signing keypairs.
+    fn domain_separator() -> &'static str {
+        "ECDSA signing key"
```
indomitableSwan, line 41: Won't we want to be more specific than this?

dams/src/crypto/signing_key.rs
```diff
@@ -0,0 +1,202 @@
...
+    ) -> Result<(SigningKeyPair, Encrypted<SigningKeyPair>), DamsError> {
+        let context = AssociatedData::new();
+        let signing_key = SigningKeyPair::generate(rng, &context);
+
+        Ok((
+            signing_key.clone(),
+            Encrypted::encrypt(rng, &self.0, signing_key, &context)?,
+        ))
+    }
+}
```
indomitableSwan, line 134: Similarly this choice feels unnatural, as opposed to having this be a method on `SigningKeyPair`.

indomitableSwan (commented):
This looks OK, but I would like to have a better idea of the overall vision for incorporating new secret types into the library, both now with respect to signing keys, but more broadly for other primitives we might wish to add in the future. Can you describe your design thoughts here?

Is the idea that the implementations for arbitrary secrets and signing keys will be disjoint? Are the choices we are making here ones that will ease implementation efforts and modularity of the code base? How about when we extend to add additional secret types?

For example, the protocols for generation and retrieval will be identical aside from the underlying key type/cryptography. Similarly, we will have overlapping notions of custody/shared control via an external policy engine.

Some inline responses to high-level PR description:
> Closes #234
> 
> The goal is to provide non-panicking methods for dealing with ECDSA signing keys (including creating locally and remotely, encrypting for storage, and signing and verifying with them).
> 

I think the PR achieves this goal, although we will need methods to validate public keys. Perhaps you intend this to be taken care of in serialization?

> This rearranges the crypto module a little bit more by pulling out `Secret`s and `SigningKey`s into their own modules along with all the relevant functions (e.g. encrypting and decrypting for storage). Open to feedback about whether this is useful (to see all the relevant methods in one place) or confusing (because e.g. `StorageKey` functionality is spread out among multiple files, even though it all lives in once place in the docs).
> 
See inline comments on this, but I don't think this is a big deal.

> I added trivial implementations to most of the methods so they don't panic, and tests that demonstrate that the implementation is trivial. There are issues referencing every method that has an incorrect implementation (#235 for sign, verify, and all the type instantiations, #241 for encrypting under the storage key, #242 for filling in remote generation).

This should help for development. But I wonder if we can have some kind of flag that will prevent us from merging broken cryptography to main inadvertently.

dams/src/crypto/arbitrary_secret.rs
```diff
@@ -0,0 +1,116 @@
...
+            .with_bytes(key_id.clone())
+            .with_str("client-generated");
+        let secret = Secret(generic::Secret::generate(rng, 32, context.clone()));
+
+        Ok((
+            secret.clone(),
+            Encrypted::encrypt(rng, &self.0, secret, &context)?,
+        ))
+    }
+}
```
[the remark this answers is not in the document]
marsella: That's a good point. That also makes it more natural when generate is expanded to support multiple types -- we can make an enum that lists the kinds of secrets that can be locally generated and stored, and then the only difference in the generate protocol will be a match that calls `create_and_encrypt` for the correct type. We could have done this anyway, but I like the explicitness of having the method be on the created type, rather than the storage key.

dams/src/crypto/signing_key.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    /// Create a new `SigningKeyPair`. This must be run by the server.
+    pub fn remote_generate(
+        rng: &mut (impl CryptoRng + RngCore),
+        user_id: &UserId,
+        key_id: &KeyId,
+    ) -> Self {
+        SigningKeyPair
+    }
+
```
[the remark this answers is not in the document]
marsella: Yep, with associated data indicating that this is server generated.

dams/src/crypto/signing_key.rs
marsella: This is just the domain separator for serialization. I'm not sure we can be more specific for a tag that'll be on every `SigningKeyPair`. This is independent of the per-key context / associated data, which will probably include specifics like user id and key id.

marsella:
> Is the idea that the implementations for arbitrary secrets and signing keys will be disjoint? Are the choices we are making here ones that will ease implementation efforts and modularity of the code base? How about when we extend to add additional secret types?
>
> For example, the protocols for generation and retrieval will be identical aside from the underlying key type/cryptography. Similarly, we will have overlapping notions of custody/shared control via an external policy engine.

I see two options for re-using the existing generate / retrieve protocols. Maybe @picklenerd and/or @vidyaap can weigh in on these.
1. Define an enum that lists out types that can be generated locally and stored. Update generate and retrieve to take an enum variant. At the appropriate point in each protocol, match on the enum and call the appropriate generation / decryption method. E.g.
```rust
fn generate( ..., secret_type: Storables) 
{
    // do stuff
    match secret_type {
        Storables::ArbitrarySecret => Secret::create_and_encrypt(...),
        Storables::SigningKey => SigningKeyPair::create_and_encrypt(...),
    }
    // do other stuff
}
```
2. Define a trait for "things that can be stored" that has a create-and-encrypt method and a generate remotely method. Define another trait for the encrypted version of those things that has a decrypt method. Update generate and retrieve to have a generic type parameter `T`, where `T` must satisfy the first trait and `Encrypted<T>` must satisfy the second. At the appropriate point in the protocol, call the trait method.
```rust
fn generate<T>(...) 
where
    T: Storable,
    Encrypted<T>: EncryptedStorable,
{
    // do stuff
    (secret, encrypted_secret) = T::create_and_encrypt(...);
    // do other stuff
}

dams/src/crypto/signing_key.rs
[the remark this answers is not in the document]
indomitableSwan: I was thinking about this on my initial review and deleted it, but going to bring it back up because of your retrospective feedback: should we consider just... not having this helper?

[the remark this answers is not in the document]
marsella: I still want to have this helper because it will set the associated data correctly, which saves the caller from having to type in the correct context ("server generated") in the correct order, and also ideally prevents them from pretending they are generating it locally when they are actually a server.

indomitableSwan (approved):
LGTM. See also minor point in unresolved conversation on ECDSA domain separator.

dams/src/crypto/signing_key.rs
marsella: I'm not sure what the options are for type of ECDSA signature? I updated to
> ECDSA signing key pair over curve secp256k1

but I also put a note in #235 to revisit this.

picklenerd:
> Define a trait for "things that can be stored" that has a create-and-encrypt method and a generate remotely method. Define another trait for the encrypted version of those things that has a decrypt method. Update generate and retrieve to have a generic type parameter T, where T must satisfy the first trait and Encrypted<T> must satisfy the second. At the appropriate point in the protocol, call the trait method.

I think this is the better option. It's the most extensible and will probably be much easier to maintain over time.

If we end up in a situation where we need custom APIs for specific customers, this approach will be very easy to implement whereas the enum approach would probably be a huge headache.

PR #244, opened 2022-09-19
224 add docker

picklenerd:
This PR adds Docker support to the key server. Docker compose is used to run MongoDB and the key server.

The integration tests were refactored to run as a normal executable and can now point to the dockerized server instead of running the server internally.

A few other changes were made to improve the ergonomics of running things:

- `cargo make` support was added with several tasks for generating certificates, starting the server, and starting the tests.
- The server CLI was simplified to reflect how it is actually used. This makes it easier to run locally.
- The `dev` folder was split to have separate configs for running locally and running in Docker. 
- A simple health check method was added to the RPC definition to make it easier for tests to wait for the server to start.
- Removed config-relative paths for certs and keys.

I also replaced `structopt` with `clap`. `structopt` was an extension of `clap` but was integrated into `clap` in version 3.


picklenerd:
The `dev/docker/e2e-Client.toml` is a bit of a hack due to the way certificates are handled. `Client.toml` is configured for a client running in Docker so it points to `"https://key_server:1113"`. The e2e client is running locally so it has to point to `"https://localhost:1113"` since there's no DNS record for `key_server` outside of Docker.

Since the certs are loaded from the same directory as the config, I just made an extra client config in here. I want to remove the feature that automatically sets the config path to the certs and make them explicit paths, but this PR is already big enough.

picklenerd:
I would recommend pulling this branch and running the commands in the README to make sure the workflow feels good and that it actually works on your machine.

README.md
```diff
@@ -28,10 +28,14 @@ Refer to the [current design specification](https://github.com/boltlabs-inc/key-
 
 ### Dependencies:
 
-- A recent version of [stable Rust](https://www.rust-lang.org/) to build the Lock Keeper project. We have tested with 1.59.0.
+- A recent version of [stable Rust](https://www.rust-lang.org/) to build the Lock Keeper project. Version 1.64 is the minimum required version.
 - OpenSSL. You should be able to install this using your package manager of choice.
-- [MongoDB](https://www.mongodb.com/try/download/community) is required to run `lock-keeper-key-server`. This includes running the integration tests.
 - `protoc` is required to build .proto files. It can be installed using `brew` for MacOS or `apt install` for Linux. Further instructions [here](https://grpc.io/docs/protoc-installation/).
+- [cargo-make](https://github.com/sagiegurari/cargo-make) can be installed with `cargo install --cargo-make`.
```
marsella, line 34: This command didn't work for me with a `--`. Just
```cargo install cargo-make```

File: README.md [hunk 2]
```diff
@@ -49,32 +53,40 @@ cargo test --all-features --doc --verbose
 
 To run all unit and integration tests:
 
-- Start MongoDB in one terminal window (see the [MongoDB docs](https://www.mongodb.com/docs/manual/reference/configuration-options/) for default mongod.conf paths based on your OS):
+1. Start the server with:
+```bash
+cargo make start-server
```
marsella, line 58: We might want a warning here that this will be slow the first time it's run.

Makefile.toml
```diff
@@ -0,0 +1,59 @@
...
+
+# Shorthand task to destroy the docker containers for a server.
+[tasks.teardown]
+alias = "teardown-server"
+
+# Start a key server and monitor its output.
+[tasks.start-server]
+command = "docker"
+args = ["compose", "up", "--build", "mongodb", "key_server"]
+dependencies = ["certs"]
```
marsella, line 22: This failed for me because I need to run it with sudo (as do all linux users, I think). Running `cargo make` with sudo didn't help. Changing this entry so the command is `sudo` and the first argument is `docker` did work, but I'm not sure that's the best solution.

README.md
```diff
@@ -28,10 +28,14 @@ Refer to the [current design specification](https://github.com/boltlabs-inc/key-
 
 ### Dependencies:
 
-- A recent version of [stable Rust](https://www.rust-lang.org/) to build the Lock Keeper project. We have tested with 1.59.0.
+- A recent version of [stable Rust](https://www.rust-lang.org/) to build the Lock Keeper project. Version 1.64 is the minimum required version.
 - OpenSSL. You should be able to install this using your package manager of choice.
-- [MongoDB](https://www.mongodb.com/try/download/community) is required to run `lock-keeper-key-server`. This includes running the integration tests.
 - `protoc` is required to build .proto files. It can be installed using `brew` for MacOS or `apt install` for Linux. Further instructions [here](https://grpc.io/docs/protoc-installation/).
+- [cargo-make](https://github.com/sagiegurari/cargo-make) can be installed with `cargo install --cargo-make`.
+- [Docker](https://www.docker.com/).
```
marsella, line 35: Starting the server initially failed for me because I only had `docker` and not `docker compose` downloaded. Maybe be more explicit that you need both.

marsella (commented):
I took a first pass trying to run this and have a few comments. 

It failed for me on the first mongo command with the error
```
Attaching to key-mgmt-key_server-1, mongodb
Error response from daemon: driver failed programming external connectivity on endpoint mongodb (c8151575a47572ea45a01c205ab6565db87fbf1aaadc36521b2904de186a18dc): Error starting userland proxy: listen tcp4 0.0.0.0:27017: bind: address already in use
```
Do you have advice on shutting down existing mongodb instances? I tried `mongod --shutdown` but it said no server running.

Makefile.toml
picklenerd: Do you need sudo any time you use docker compose? You shouldn't need it at all for Docker but I seem to remember having to changing some group settings in Linux at my last job. It might also be the `certs` task.

README.md
vidyaap: Is it worth pointing people to [Docker Desktop](https://www.docker.com/products/docker-desktop/) instead? I believe that gives you both and I think it's easier to interact with Docker using it.

Makefile.toml
picklenerd: I added a note with a link to this page: https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user. Let me know if those steps work.

README.md
```diff
@@ -49,32 +54,40 @@ cargo test --all-features --doc --verbose
...
 ```bash
-mongod --config {path_to_mongod.conf}
+cargo make e2e
 ```
 
-- Open another terminal window, navigate to this repo and run:
+The server will be running in the background so you can continue to run integration tests without starting the server again. To stop the server, run:
 ```bash
-cargo test --all-features --all-targets
+cargo make stop-server
```
vidyaap, line 70: This didn't run the server as a background process for me; was it supposed to? Otherwise, it might be worth it to specify that it doesn't.

README.md
picklenerd: I think this is only an issue on Linux so I added a note about it. If anyone has this problem on other OSes we can update this.

docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+version: '3.8'
+
+services:
+  mongodb:
+    image: mongo
+    container_name: mongodb
+    restart: always
+    # environment:
+    #   MONGO_INITDB_ROOT_USERNAME: $MONGODB_USERNAME
+    #   MONGO_INITDB_ROOT_PASSWORD: $MONGODB_PASSWORD
```
vidyaap, line 10: Do we need the environment variables here either?

README.md
vidyaap: Oops, wrong line, I meant line 61

docker-compose.yml
picklenerd: I left these here in case we ever do Docker deployments to a real server but I don't think we'd use docker compose for that so I do think these can go.

README.md
picklenerd: Oh yeah I updated the defaults and added some shorthand commands but forgot to update the docs.

picklenerd:
> Is it worth pointing people to [Docker Desktop](https://www.docker.com/products/docker-desktop/) instead? I believe that gives you both and I think it's easier to interact with Docker using it.

I'm not actually sure if that would be better. It's the only option for Mac and Windows but on Linux my experience is that people prefer to run Docker Engine.

README.md
```diff
@@ -49,32 +59,44 @@ cargo test --all-features --doc --verbose
...
+```bash
+cargo make start-server
 ```
 
-We follow test-driven development practices and the test suite should be a close mapping to the functionality we currently implement at any given stage of development.
 
 ## Running the server locally
 
-To run the server locally, make sure MongoDB is running as above. Then, generate an SSL cert locally using the provided script in the `dev/` directory:
+To run the server locally, make sure MongoDB is running as above. Then run:
```
vidyaap, line 84: Nit, but should we put the instructions for running MongoDB locally back in?

vidyaap (commented):
Thanks, this looks good to me, and the commands all run as expected on my machine! I just left one nit, and also wasn't sure if we should wait for @marsella to confirm that it runs on her machine too?

README.md
picklenerd: I left the command out but added the link back along with a new link for how to run it in Docker.

picklenerd:
> Thanks, this looks good to me, and the commands all run as expected on my machine! I just left one nit, and also wasn't sure if we should wait for @marsella to confirm that it runs on her machine too?

I definitely want to wait for @marsella since she's using Linux.

Makefile.toml
marsella: I always use `sudo` with Docker because I didn't want to set up that extra root-privilege group... Is there a way to add OS-specific conditions to this file?

picklenerd: Yes but I'd rather not since it would force people to enter their password even if they have the group set up. I think we could add some new tasks that work that way though. Would you be willing to make a separate PR that has new tasks defined with whatever works on your system?

marsella: I will try running through the instructions with the `docker` group enabled for now.

marsella, line 22: On this command, I got the error
```
Error response from daemon: driver failed programming external connectivity on endpoint mongodb (bd0d6f163e9c9e3d93ac36aa1faa4edd12a49c4223ab03cf0f50131553ebeff7): Error starting userland proxy: listen tcp4 0.0.0.0:27017: bind: address already in use
```
until I manually killed my existing, locally-running mongo instance ([stackoverflow post here](https://stackoverflow.com/questions/55887414/docker-compose-error-cannot-start-service-mongo-driver-failed-programming-exter), but the `systemctl` process didn't work for me -- I just found the pid with `ps aux | grep mongod` and manually killed it). 

This was fine for me but might be an issue if someone has local mongodb instances they want to maintain while also running our service.

marsella (commented):
Works fine with the docker group enabled. One note about compatibility with people who have a local mongodb server they don't want to kill.

picklenerd:
> Do you have advice on shutting down existing mongodb instances? I tried mongod --shutdown but it said no server running.

I think this is going to be system-dependent. For me it was just `brew services stop mongodb-community@6.0`.

Makefile.toml
picklenerd: I think this could go one of two ways:

1. We change our MongoDB port to something else.
2. Not our problem. This is what life is like when you try to run things locally so you probably already know how to do things manually.

I like option 2 because we keep using standard MongoDB ports which means we probably won't have to worry about conflicts. If we ever get a real ops person, I'd feel more comfortable getting fancy with server configuration.

I'm totally fine with option 1 if you disagree though.

jwantz: Just a quick note related to your ops mention... Many NetOps teams select their own ports when deploying production services, which they often select from the private (unassigned, non-standard) range. Others try to go with the defaults. It depends upon the deployment model and preferences of the individual network team. So, throwing in my two cents... if we allow the port to be configured/customized and we document where to make that change, then I don't have a preference on which option you select.

marsella: I don't have a strong preference, especially now while we don't have a particularly wide range of testers. If we start getting reports from stakeholders that they're seeing this error, we'll know what to look for, but I'm fine to not make any changes at this point.

marsella (approved):
Thank you for your patience in getting this reviewed!

PR #248, opened 2022-09-21
232 improve logging

### LINKED ISSUE #232: Improve logging interface

vidyaap:
Closes #232 

My tonic changes are based on [this post](https://dev.to/anshulgoyal15/a-beginners-guide-to-grpc-with-rust-3c7o). Linking here in case it's useful in the future for handling sessions/other authentication related things on the server.

Main changes:
- Add the account name to every request's metadata
- Remove `LogIdentifier` and use account name for all logs
- Pull KeyId from operation results
- Change "audit log" language to "audit event"
- Remove log extension trait + function
- Log beginning and end of each operation

Questions:
- I'd like to make a constant for the string "account_name". Where in the dams crate can I put that (or do we have enough constants that I can make a constants module there)?
- ~I don't like the way I'm handling KeyIds here because we can only log them in the successful case here, but putting KeyIds in the request metadata doesn't seem appropriate. Thoughts on how to get the KeyId in the error case?~ 
    It's now in the server context as per @picklenerd's suggestion!

### DIFF
--- a/lock-keeper-client/src/client.rs
+++ b/lock-keeper-client/src/client.rs
@@ -17,6 +17,7 @@ use rand::{rngs::StdRng, SeedableRng};
 use std::{str::FromStr, sync::Arc};
 use tokio::sync::{mpsc, Mutex};
 use tokio_stream::wrappers::ReceiverStream;
+use tonic::{metadata::MetadataValue, Request};
 use tracing::error;
 
 // TODO: password security, e.g. memory management, etc... #54
@@ -56,6 +57,7 @@ impl Password {
 pub struct LockKeeperClient {
     session_key: OpaqueSessionKey,
     config: Config,
+    account_name: AccountName,
     user_id: UserId,
     tonic_client: LockKeeperRpcClient<LockKeeperRpcClientInner>,
     pub(crate) rng: Arc<Mutex<StdRng>>,
@@ -78,9 +80,14 @@ pub(crate) struct AuthenticateResult {
 
 #[allow(unused)]
 impl LockKeeperClient {
-    // Get [`UserId`] for the authenticated client.
+    /// Get [`UserId`] for the authenticated client.
     pub fn user_id(&self) -> &UserId {
         &self.user_id
+    }
+
+    /// Get [`AccountName`] for the authenticated client.
+    pub fn account_name(&self) -> &AccountName {
+        &self.account_name
     }
 
     pub fn tonic_client(&self) -> LockKeeperRpcClient<LockKeeperRpcClientInner> {
@@ -138,7 +145,7 @@ impl LockKeeperClient {
         let mut rng = StdRng::from_entropy();
 
         let mut client_channel =
-            Self::create_channel(&mut client, ClientAction::Authenticate).await?;
+            Self::create_channel(&mut client, ClientAction::Authenticate, account_name).await?;
         let result =
             Self::handle_authentication(client_channel, &mut rng, account_name, password).await;
         match result {
@@ -153,6 +160,7 @@ impl LockKeeperClient {
                     config: config.clone(),
                     tonic_client: client,
                     rng: Arc::new(Mutex::new(rng)),
+                    account_name: account_name.clone(),
                     user_id,
                     export_key,
                 };
@@ -180,7 +188,8 @@ impl LockKeeperClient {
         let mut rng = StdRng::from_entropy();
         let server_location = config.server_location()?;
         let mut client = Self::connect(config).await?;
-        let mut client_channel = Self::create_channel(&mut client, ClientAction::Register).await?;
+        let mut client_channel =
+            Self::create_channel(&mut client, ClientAction::Register, account_name).await?;
         let result =
             Self::handle_registration(client_channel, &mut rng, account_name, password).await;
         match result {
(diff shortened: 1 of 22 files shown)


lock-keeper-key-server/src/database/audit_event.rs
```diff
@@ -0,0 +1,83 @@
...
+                    tracing::error!("{}", e);
+                }
+                // Log action
+                context
+                    .db
+                    .create_audit_event(&context.account_name, None, action, Outcome::Failed)
+                    .await?;
+                // Give the client a moment to receive the error before dropping the channel
+                thread::sleep(Duration::from_millis(100));
+                Ok(())
```
picklenerd, line 79: This should't be in here. This is part of normal logging. The audit events shouldn't ever require a channel.

picklenerd:
> I'd like to make a constant for the string "account_name". Where in the dams crate can I put that (or do we have enough constants that I can make a constants module there)?

I think we can make a constants file. The string constants in `defaults.rs` could go in there too.

lock-keeper-client/src/client.rs
```diff
@@ -208,11 +220,18 @@ impl LockKeeperClient {
     pub(crate) async fn create_channel(
         client: &mut LockKeeperRpcClient<LockKeeperRpcClientInner>,
         action: ClientAction,
+        account_name: &AccountName,
```
picklenerd, line 223: I think this method could take `&self` instead since the account name is already stored in it.

lock-keeper-client/src/client.rs
```diff
@@ -208,11 +220,18 @@ impl LockKeeperClient {
...
     ) -> Result<ClientChannel, LockKeeperClientError> {
         // Create channel to send messages to server after connection is established via
         // RPC
         let (tx, rx) = mpsc::channel(2);
-        let stream = ReceiverStream::new(rx);
+        let mut stream = Request::new(ReceiverStream::new(rx));
+
+        let account_name_val = MetadataValue::try_from(account_name.to_string())
+            .map_err(|_| LockKeeperClientError::InvalidAccount)?;
```
picklenerd, line 231:
```suggestion
        let account_name_val = MetadataValue::try_from(account_name.to_string())?;
```

This tonic error should be added to the error type. This should work: 
```rust
    #[error(transparent)]
    TonicMetadata(#[from] tonic::metadata::errors::InvalidMetadataValue),
```

lock-keeper-key-server/src/server.rs
```diff
@@ -3,20 +3,25 @@ mod service;
...
+    ClientAction,
 };
 
+use lock_keeper::{crypto::KeyId, user::AccountName};
 use rand::{rngs::StdRng, SeedableRng};
 use std::sync::Arc;
 use tokio::sync::Mutex;
 use tonic::{Request, Response, Status};
 
+pub(crate) struct OperationResult(pub(crate) Option<KeyId>);
```
picklenerd, line 23: Do we still need this if we're adding it to the context now?

lock-keeper-key-server/src/server.rs
```diff
@@ -42,12 +47,27 @@ impl LockKeeperKeyServer {
         })
     }
 
-    pub fn context(&self) -> Context {
-        Context {
+    pub fn context(
+        &self,
+        request: &Request<tonic::Streaming<Message>>,
```
picklenerd, line 52: I like that we're getting the request here. I think this will give us a lot of flexibility in the future

lock-keeper-key-server/src/server/operation.rs
```diff
@@ -31,20 +34,44 @@ pub(crate) trait Operation: Sized + Send + 'static {
...
+                &context.account_name,
+                &context.key_id,
+                &context.action,
+                status,
+            )
+            .await;
+        if let Err(e) = audit_event {
+            let _ = Self::handle_error(channel, e);
+        };
+    }
```
picklenerd, line 76: These are great. This will make it really easy to add steps to both of these processes.

I think they can be free functions outside of the `Operation` definition. I think it's important to keep trait definitions as small as possible.

lock-keeper-client/src/client.rs
vidyaap: I think it only doesn't because of register and authenticate. Is there a way to have it take `&self` anyway?

lock-keeper-key-server/src/server.rs
vidyaap: Oops, forgot to remove!

lock-keeper-client/src/client.rs
picklenerd: Oh right I always forget that those are weird. Nevermind!

picklenerd (approved):
I think it would be good to make that "account_name" constant before merging but this looks great otherwise.

vidyaap:
> I think it would be good to make that "account_name" constant before merging but this looks great otherwise.

Sorry, missed your first comment about this! I added "account_name" and "user_id" to defaults for now, because I think there will be some changes when we refactor the lock-keeper crate anyway.

PR #249, opened 2022-09-22
Improve extensibility for stored-secret types

### LINKED ISSUE #242: Add remote generation methods

marsella:
Closes #242 (implementing remote generation for arbitrary secrets) and addresses comments discussed in #243. Specifically, adds a trait for "things that can be generated and stored securely at the server" to reduce repeated code.

I am not convinced that `Storable` is a good name for this type, but I'm wary of making another thing called `Secret` -- we already have two `Secret`s and a `StoredSecret` in the DB module. Any ideas?

We might be able to abstract this even more once I implement `create_and_encrypt` for `SigningKeyPair`s. I suspect that this function will be identical for key pairs and secrets modulo their generate functions. If that's true, I might add a default implementation to the trait so that in the future, all we have to do is define how to generate new secret stored types. (but there may be something weird with keeping that generate function private, so TBD).

EDIT: We might not always be able to maintain a single decrypt function for every stored type, especially if we want to, e.g. validate associated data or something.

I only need review from one of @picklenerd and @vidyaap.


indomitableSwan:
> Closes #242 (implementing remote generation for arbitrary secrets) and addresses comments discussed in #243. Specifically, adds a trait for "things that can be generated and stored securely at the server" to reduce repeated code.
> 
> I am not convinced that `Storable` is a good name for this type, but I'm wary of making another thing called `Secret` -- we already have two `Secret`s and a `StoredSecret` in the DB module. Any ideas?
> 
> We might be able to abstract this even more once I implement `create_and_encrypt` for `SigningKeyPair`s. I suspect that this function will be identical for key pairs and secrets modulo their generate functions. If that's true, I might add a default implementation to the trait so that in the future, all we have to do is define how to generate new secret stored types. (but there may be something weird with keeping that generate function private, so TBD).
> 
> EDIT: We might not always be able to maintain a single decrypt function for every stored type, especially if we want to, e.g. validate associated data or something.
> 
> I only need review from one of @picklenerd and @vidyaap.

As an aside, I question the naming of `StoredSecret` for the DB. Since we have an `EncryptedSecret` type already, `StoredEncryptedSecret` would seem more plausible, or, from a semantic perspective, perhaps `ClientCiphertext` or `StoredClientCiphertext`.

More to the point, this PR seems like it's begging for a bit more up front design thinking. Specifically, I think the fact that we already have two `Secret`s is indicative that we are still struggling with our abstractions here and we will probably want to reconsider our naming at the same time.

I recommend we hold off on merging this PR:
- Now is a good time for us (the cryptographers) to take a step back and dedicate some design time (both protocol and software) to more complicated key types.
- My understanding is that we already have enough for the developers to move forward on protocol implementations for signing keys, so taking the design time now would not be blocking anyone. Am I understanding that correctly? If we are merging this now because it makes @vidyaap or @picklenerd's life significantly easier in the short-term, this would change my perspective on this PR.

marsella:
> As an aside, I question the naming of StoredSecret for the DB. Since we have an EncryptedSecret type already, StoredEncryptedSecret would seem more plausible, or, from a semantic perspective, perhaps ClientCiphertext or StoredClientCiphertext.

Yeah, I think we'll want to change that when we implement the generate-and-store flow with signing keys. I'm not sure how well MongoDB would handle trait objects vs types (e.g. can we have a list of `Storables` or do we have to maintain a separate vector for each type of secret we can hold?), but I'm guessing the abstraction will change either way.

> More to the point, this PR seems like it's begging for a bit more up front design thinking. Specifically, I think the fact that we already have two Secrets is indicative that we are still struggling with our abstractions here and we will probably want to reconsider our naming at the same time.
>
> I recommend we hold off on merging this PR:
>
> Now is a good time for us (the cryptographers) to take a step back and dedicate some design time (both protocol and software) to more complicated key types.
> My understanding is that we already have enough for the developers to move forward on protocol implementations for signing keys, so taking the design time now would not be blocking anyone. Am I understanding that correctly? If we are merging this now because it makes @vidyaap or @picklenerd's life significantly easier in the short-term, this would change my perspective on this PR.

I agree that this is a good moment to do design thinking, but this might block the developer roadmap a bit, depending on what they start with. Updating the existing generate/retrieve flows to work with multiple different secret types will require this PR to be in place, because they need to know what abstraction they're dealing with. But I don't think it would block the continued audit log updates, the testing refactor, or implementation of signing-key-specific flows (e.g. requesting a secret).

indomitableSwan:
> > As an aside, I question the naming of StoredSecret for the DB. Since we have an EncryptedSecret type already, StoredEncryptedSecret would seem more plausible, or, from a semantic perspective, perhaps ClientCiphertext or StoredClientCiphertext.
> 
> Yeah, I think we'll want to change that when we implement the generate-and-store flow with signing keys. I'm not sure how well MongoDB would handle trait objects vs types (e.g. can we have a list of `Storables` or do we have to maintain a separate vector for each type of secret we can hold?), but I'm guessing the abstraction will change either way.
> 
Flagging @jwantz on this point, given that we may want to revisit our choice of MongoDB anyway.

> > More to the point, this PR seems like it's begging for a bit more up front design thinking. Specifically, I think the fact that we already have two Secrets is indicative that we are still struggling with our abstractions here and we will probably want to reconsider our naming at the same time.
> > I recommend we hold off on merging this PR:
> > Now is a good time for us (the cryptographers) to take a step back and dedicate some design time (both protocol and software) to more complicated key types.
> > My understanding is that we already have enough for the developers to move forward on protocol implementations for signing keys, so taking the design time now would not be blocking anyone. Am I understanding that correctly? If we are merging this now because it makes @vidyaap or @picklenerd's life significantly easier in the short-term, this would change my perspective on this PR.
> 
> I agree that this is a good moment to do design thinking, but this might block the developer roadmap a bit, depending on what they start with. Updating the existing generate/retrieve flows to work with multiple different secret types will require this PR to be in place, because they need to know what abstraction they're dealing with. But I don't think it would block the continued audit log updates, the testing refactor, or implementation of signing-key-specific flows (e.g. requesting a secret).

In the last line: requesting a signature, you mean?

To clarify, would not merging this PR actually be blocking development (or otherwise make development more difficult now), or just imply there will be a refactor that impacts protocol implementations in the future? My concern here is that I am not at all confident that this PR contains a durable abstraction, i.e., I expect we will have such a refactor in the future anyway.

picklenerd:
>  I'm not sure how well MongoDB would handle trait objects vs types (e.g. can we have a list of Storables or do we have to maintain a separate vector for each type of secret we can hold?), but I'm guessing the abstraction will change either way.

We'll probably have to tag them in the DB but that can be done with an associated constant in the trait definition. That tag could be added to `StoredSecret` so that it's in the DB record with the data. Having a generic way to get a type from a tagged record will still be a little tricky.

I agree that it's worth taking a step back now to think about what secrets stored in the DB will look like. Maybe we can come up with a generic set of secret types that can cover every use case or maybe we need this more flexible solution where we can store anything in the DB. 

If we can't sort it out before this PR becomes a blocker, I don't think there's any harm in merging this and refactoring later. I think this is already a valuable step forward in the design.

As far as naming goes, the trait should be named based on what the methods do and not how we intend for it to be used. I actually kind of like `Storable` since the methods create things that are acceptable to store.

lock-keeper/src/crypto.rs
```diff
@@ -263,6 +263,47 @@ impl KeyId {
...
+
+    /// Create a new `Storable`. This is part of the flow to remotely generate
+    /// a new secret.
+    ///
+    /// This must be run by the server.
+    fn remote_generate(
+        rng: &mut (impl CryptoRng + RngCore),
+        user_id: &UserId,
+        key_id: &KeyId,
+    ) -> Self;
```
picklenerd, line 288: This kind of bugs me since it returns an unencrypted object but still calls itself `Storable`. It's still going to be encrypted in some way before it's stored isn't it? If so, I feel like this method should go somewhere else and the output of that somewhere else can then be turned into something `Storable`.

lock-keeper/src/crypto.rs
```diff
@@ -263,6 +263,47 @@ impl KeyId {
...
+impl<T> Encrypted<T>
+where
+    T: Storable + TryFrom<Vec<u8>>,
+    CryptoError: From<<T as TryFrom<Vec<u8>>>::Error>,
+    Vec<u8>: From<T>,
+{
+    /// Decrypt a `Storable`. This should be run as part of the subprotocol to
+    /// retrieve an encrypted secret from the server.
+    ///
+    /// This must be run by the client.
```
picklenerd, line 300: Maybe this should be a free function in the client crate so that it can't be misused.

indomitableSwan:
> > I'm not sure how well MongoDB would handle trait objects vs types (e.g. can we have a list of Storables or do we have to maintain a separate vector for each type of secret we can hold?), but I'm guessing the abstraction will change either way.
> 
> We'll probably have to tag them in the DB but that can be done with an associated constant in the trait definition. That tag could be added to `StoredSecret` so that it's in the DB record with the data. Having a generic way to get a type from a tagged record will still be a little tricky.
> 
> I agree that it's worth taking a step back now to think about what secrets stored in the DB will look like. Maybe we can come up with a generic set of secret types that can cover every use case or maybe we need this more flexible solution where we can store anything in the DB.
> 
Makes sense to me.

> If we can't sort it out before this PR becomes a blocker, I don't think there's any harm in merging this and refactoring later. I think this is already a valuable step forward in the design.
> 
Here is where I disagree. I think this complicates the software design before taking the time to think through what abstractions we need. Unraveling that can sometimes be more of a headache than taking some more time up front, and I think there is also a risk of clouding our thinking for future refactors.

What I'm still unclear on: to what extent is not merging this PR a blocker?

> As far as naming goes, the trait should be named based on what the methods do and not how we intend for it to be used. I actually kind of like `Storable` since the methods create things that are acceptable to store.

Yes, that makes sense. `Storable` seems more on point than `StoredSecret`, which to me doesn't describe what the method does or how we intend to use it....

picklenerd:
> What I'm still unclear on: to what extent is not merging this PR a blocker?

I don't think this will be a blocker until we start the next phase. Once we start development and need to store different types of secrets, I think we'll need this PR to avoid writing duplicate code for different types of stored information.

vidyaap:
> Yes, that makes sense. Storable seems more on point than StoredSecret, which to me doesn't describe what the method does or how we intend to use it....

Just throwing in that `StoredSecret` is just a wrapper struct around an `Encrypted<Secret>` and some metadata (key ID, flag for whether it was retrieved, etc.), that I used to make storing and looking up keys in the DB easier. I don't love the name, but I'm not sure what a concise way to say the above is. Does `StoredEncryptedSecret` make it clearer?

indomitableSwan:
> > Yes, that makes sense. Storable seems more on point than StoredSecret, which to me doesn't describe what the method does or how we intend to use it....
> 
> Just throwing in that `StoredSecret` is just a wrapper struct around an `Encrypted<Secret>` and some metadata (key ID, flag for whether it was retrieved, etc.), that I used to make storing and looking up keys in the DB easier. I don't love the name, but I'm not sure what a concise way to say the above is. 
>
> Does `StoredEncryptedSecret` make it clearer?

Maybe? Naming is hard. I think the missing bit for me is some indication that the key server does not (ever) have access to the key material itself in this case, i.e., they do not control the encryption key that created the `Encrypted<Secret>`. Which is different from the use case for a remote-generated secret, in which case the key server *does* have access to the generated secret, at least at generation time. 

The common use case for signing keys will be remote generation and usage, where the key servers will _sometimes_ have access to key material (that is, at generation and when creating a signature).

Looking ahead (past this next phase that adds basic ECDSA signing key support for EVM), the way keys are generated will change, and involve different storage needs at the server-side than what we have now.

marsella:
NB for future: this should probably expand to include the `import_and_encrypt` method added in #277. 

Closing this because it's out of date. We might revisit when we get to client-side signing key generation and retrieval.

PR #250, opened 2022-09-22
Updated Cargo manifests to use 1.64 workspace definitions

picklenerd:
Rust 1.64 added the ability to specify some workspace-level values that can be shared between all of the crates in a workspace.

This PR adds the version number and all shared dependencies to the workspace `Cargo.toml` so that they can be updated in a single place.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -21,6 +21,11 @@ jobs:
         uses: arduino/setup-protoc@v1
         with:
           version: '3.x'
+      
+      # Ensure Rust is the latest version
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
 
       - name: Cache
         uses: actions/cache@v2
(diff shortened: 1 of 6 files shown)


picklenerd:
Gotta figure out how to update the Rust version running the E2E tests.

marsella (approved):
This seems very useful to keep things in sync! I see that some workspace dependencies are just dev dependencies in some of the crates. Will this bloat the size of our executables?

picklenerd:
> This seems very useful to keep things in sync! I see that some workspace dependencies are just dev dependencies in some of the crates. Will this bloat the size of our executables?

I don't think the workspace dependencies actually add the dependency to each create. You still have to import them in each crate with `name.workspace = true`. All it really does it centralize the version numbers.

I don't think any code from unused dependencies would end up in the executable either way though.

picklenerd:
This new feature doesn't seem to be recognized by the VS Code extension so it highlights the files in red even though it compiles fine. I'm going to hold off on merging while I figure out how to solve that or wait for it to solve itself.

picklenerd:
Turns out it was just my TOML extension.

Issue #48, opened 2022-06-29
Evaluate the contents of the transaction approval request


vidyaap:
Right now the transaction approval request (TAR) consists of a transaction, an ID that we assign to the transaction, and a number of IDs indicating the user, asset, and key that are associated with the transaction.  However, there may be more or less information required here, e.g.:
- Do we need to keep the asset ID?
- What user information is required?
- How do we identify the transaction originator?

This issue can be closed when:
- [ ] We have accepted and documented a list of fields to make up the TAR
- [ ] We have created an issue to reflect these fields in the implementation

Issue #47, opened 2022-06-29
Update scope of transaction approval requests


marsella:
Right now, `TransactionApprovalRequest`s have a corresponding ID, which must be unique among requests. This is an immutable, read-only field on the TAR. 

In practice, we don't want the calling application (or transaction originating application) to be in charge of choosing the ID. Goals: avoid duplicated IDs, avoid potential attacks from similar/same IDs on different TARs. 

To close this issue 
- [ ] determine who the correct party is to choose the TAR (the mighty block app? our client code? the key servers via consensus? some combination of these?)
- [ ] define the flow for converting TARs into TARs with IDs assigned by the correct party (and telling the calling app what the ID is, probably).
- [ ] implement the solution

indomitableSwan:
To reduce the attack surface and increase misuse resistance, choosing the TAR ID should be done by our client code as part of a transaction approval request session.

vidyaap:
I changed the title of this issue because I think ID assignment is part of a larger scoping issue (scoping in the sense of “scope of a variable in a program”). A `TransactionApprovalRequest` needs to have an ID, and we want to be the ones to assign the ID.  In that case, it makes sense for this to be an internal type that we don’t expose to the calling application, and instead construct ourselves after asking for each of the fields as a separate parameter in the API functions.

It may be worth making a separate ID-less version of a TAR if it’s convenient for the calling app devs to group this data together.  But that seems unlikely since the calling app would already have its own way of storing the data we need to make a TAR.

indomitableSwan:
Closing as an issue that was written prematurely. See https://github.com/boltlabs-inc/key-mgmt-spec/issues/136.

Issue #43, opened 2022-06-28
Move SignatureScheme trait to crypto module


vidyaap:
The `SignatureScheme` trait is a trait to indicate objects representing a signature scheme, and would hold signing operations common to different signature schemes. Right now, this trait is defined in the blockchain module, but some refactoring needs to be done to:
- [ ] move the trait definition to a crypto library
- [ ] define signing operations as trait methods
- [ ] ensure that structs representing specific signature schemes (e.g. the existing `ECDSA` struct in the blockchain module) implement this trait

marsella:
The goals of this issue might change as we develop the crypto module. For example, we might 

An alternate approach to this: we may define a `Signer` trait over keys, rather than signature scheme objects (like `ECDSA`). In that case, this issue will have to find a way to link between types that implement the `SignatureScheme` trait and types that implement `Signer`. The `SignatureScheme` would indicate what _type_ of `Signer`s are appropriate for a blockchain, but not the actual keys used.

indomitableSwan:
@marsella This issue can be closed as outdated, right?

marsella:
I think if #252 ends up moving this trait out of the repo, then yes. Otherwise we should probably move it to the repo, and maybe just remove it.

indomitableSwan:
Closing because we moved this trait out of the repo in #252. I think the correct time to decide where this trait should live and what it should do is when/if we re-integrate it.

Issue #27, opened 2022-06-15
Design and implement representation of signing authorities


marsella:
Any key with delegated signing authority needs to have a field to describe who the delegated party is. In a simplified instantiation, this party might be the same for all keys, or selected from a list of approved delegated parties (assigned at system setup). In a more complex version, this might be an arbitrary party.

`Delegated` and `Passive` types need to specify who the party is. This issue is complete when

- [ ] There is a type that holds all the information needed to interact with a delegated party
- [ ] There is a way to specify a default value (e.g. in a config file) for the simplified version of the system (where the Service Provider is always the delegated party)
- [ ] There are additional notes in this issue about how to extend the representation to more complex cases with multiple delegated parties

indomitableSwan:
Closing as premature.

PR #253, opened 2022-09-26
Move unimplemented API definitions to new repo

picklenerd:
This PR moved the unimplemented API functions out of the `lock-keeper-client` crate and into a markdown file in the crate root called `api.md`. The entire file is wrapped in a Rust code block so it should be easily viewable from Github.

I updated the documentation at the top of `api.md` to reflect the content of the file.

I'm open to suggestion for better locations for this file. I like that it's front and center here but I think we'll end up with a pretty bloated root directory at some point. This could be a problem we solve later.


indomitableSwan:
My feeling is that we want to move this API out of the repo entirely. It seems more appropriate to document internally and have this repo reflect our current, iterative development process without the vestiges of our original component-based approach.

picklenerd:
> My feeling is that we want to move this API out of the repo entirely. It seems more appropriate to document internally and have this repo reflect our current, iterative development process without the vestiges of our original component-based approach.

I like that idea too but it would have to go on confluence or a different repo since we can't use gists. Do you have a preference for where it ends up?

marsella (commented):
The other things we created at the same time as this API (and which I think are similarly outdated) are `blockchain.rs`, `transaction.rs`, `policy_engine.rs`, and `keys.rs`. Those files mostly contain type definitions and extra documentation for the types used here, and won't make sense in this repo without this API.

I don't have a strong preference re: moving this out of the repo entirely. We could put it in our client's confluence pages along with the issues I flagged offline (#26, #27, #47, #43, #48) that are all about modifications to these methods and types. We could make a separate repo and port these files and issues over there so they are contained and referenceable but not in the way of the rest of our development.

indomitableSwan:
> The other things we created at the same time as this API (and which I think are similarly outdated) are `blockchain.rs`, `transaction.rs`, `policy_engine.rs`, and `keys.rs`. Those files mostly contain type definitions and extra documentation for the types used here, and won't make sense in this repo without this API.
> 
> I don't have a strong preference re: moving this out of the repo entirely. We could put it in our client's confluence pages along with the issues I flagged offline (#26, #27, #47, #43, #48) that are all about modifications to these methods and types. We could make a separate repo and port these files and issues over there so they are contained and referenceable but not in the way of the rest of our development.
> 
> Edit: sorry, I see those are the two options you already suggested. I guess I have a mild preference for a separate repo because it's all code (albeit, not code we're going to run). Github has the option of archiving repos, which puts an explicit warning that the software is not being developed. That might be a good option.

I think Marcella's suggestion of a separate repo makes sense.

picklenerd:
Moved here: https://github.com/boltlabs-inc/tentative-lock-keeper-api

indomitableSwan (changes requested):
This is a good start, but as [Marcella pointed out](https://github.com/boltlabs-inc/key-mgmt/pull/253#pullrequestreview-1127307303), there is other code in this repository that similarly needs to be moved out to the new repo.

picklenerd:
I've updated the other repo with those files and removed them in this PR.

indomitableSwan (approved):
Thanks! LGTM.

Issue #251, opened 2022-09-26
Create CI task to publish rustdoc for all crates in key-mgmt workspace


picklenerd:
Code documentation should be easily viewable from the repository without having to build it locally. I think it's possible to publish them to github.io which would probably be the easiest solution.

- [ ] Add CI task to build docs and publish them to github.io.
- [ ] Add links to README for each of the crates in the workspace.

PR #255, opened 2022-09-26
Publish docs to Github pages #251

picklenerd:
This PR adds a CI task to build documentation and publish it to Github Pages.

This is a little tricky to see the results of since it depends on things being merged to `develop` or `main` but I did test this using this PR branch and it works.

I manually published the current docs for `develop` and `main` which you can see here: https://boltlabs-inc.github.io/key-mgmt/

This index is ugly but the docs for each repo are directly linked in the README now so no one should ever see it.

There is a new protected branch called `published-documentation` where all of these doc files live. We shouldn't ever have to work with it directly but you might see it when browsing branches.

The GitHub Pages settings can be accessed from the settings page for this repo: https://github.com/boltlabs-inc/key-mgmt/settings/pages

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -1,0 +1,66 @@
+# Builds documentation and publishes it to GitHub Pages via the 
+# `published-documentation` branch of the `key-mgmt` repo.
+name: Publish Docs
+
+on:
+  push:
+    branches: [ main, develop ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
+      - name: Cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,66 @@
...
+          git config --global user.email "CI"
+          git config --global user.name "CI"
+
+      - name: Switch to doc branch and publish
+        run: |
+          git fetch
+          git checkout published-documentation
+          rm -rf "docs/${{ steps.vars.outputs.branch }}"
+          mkdir -p "docs/${{ steps.vars.outputs.branch }}"
+          yes | cp -rf index.html docs/index.html
```
marsella, line 62: TIL the `yes` command. How useful!

marsella (approved):
Looks good to me.

Is this intended for external consumption or just for us? If the latter, is it useful to add `--document-private-items`?

picklenerd:
> Is this intended for external consumption or just for us? If the latter, is it useful to add --document-private-items?

I think it just for us at the moment. I do think that would be useful.

PR #256, opened 2022-09-27
74 retrieve logs

### LINKED ISSUE #74: Implement the "retrieve audit logs" flow
The [retrieve audit logs](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/system-functionalities.md#retrieve-audit-logs) flow is a multi-message protocol between client and server.

1. The client sends a request to the server with a user ID and type of logs requested.
2. The client sends a retrieval request (type not defined at time of writing), including a user ID and key ID.
3. The key server checks the request, retrieves the logs, and sends to the client. The key server also logs that this audit log request happened.
4. The client returns the logs to the calling app.

- [x] Define client and server message types and serialization into / from tonic messages.
- [x] Implement client-side protocol
- [x] Implement server-side protocol
- [x] Integration tests

vidyaap:
Closes #74 

This PR implements the audit event retrieval flow, and also changes the serialization library from `bincode` to `serde_json`.

Question: The spec says that we need to pass the UserId in the request, but the audit events are now stored and looked up by AccountName, and that is passed in the request metadata. Should the UserId be passed in the message anyway? Is there any validation we'd want to add with the UserId (e.g. making sure the AccountName in the metadata matches that of the UserId)?

### DIFF
--- a/lock-keeper-client/src/api.rs
+++ b/lock-keeper-client/src/api.rs
@@ -43,7 +43,7 @@ impl LockKeeperClient {
     pub async fn retrieve_audit_event_log(
         &self,
         event_type: EventType,
-        options: Option<AuditEventOptions>,
+        options: AuditEventOptions,
     ) -> Result<Vec<AuditEvent>, LockKeeperClientError> {
         let mut client_channel = Self::create_channel(
             &mut self.tonic_client(),
--- a/lock-keeper-client/src/api.rs
+++ b/lock-keeper-client/src/api.rs
@@ -8,17 +8,53 @@
 pub(crate) mod authenticate;
 pub(crate) mod create_storage_key;
 pub(crate) mod register;
+pub(crate) mod retrieve_audit_events;
 
 pub mod arbitrary_secrets;
 
-use crate::LockKeeperClientError;
+use crate::{LockKeeperClient, LockKeeperClientError};
 use lock_keeper::{
+    audit_event::{AuditEvent, AuditEventOptions, EventType},
     blockchain::Blockchain,
     crypto::KeyId,
     keys::{KeyInfo, UsePermission, UseRestriction, UserPolicySpecification},
     transaction::{TransactionApprovalRequest, TransactionSignature},
     user::UserId,
+    ClientAction,
 };
+
+impl LockKeeperClient {
+    /// Retrieve the log of audit events from the key server for the
+    /// authenticated asset owner; optionally, filter for audit events
+    /// associated with the specified [`KeyId`].
+    ///
+    /// The log of audit events includes context
+    /// about any action requested and/or taken on the digital asset key,
+    /// including which action was requested and by whom, the date, details
+    /// about approval or rejection from each key server, the policy engine,
+    /// and each asset fiduciary (if relevant), and any other relevant
+    /// details.
+    ///
+    /// The [`UserId`] must match the asset owner authenticated in the
+    /// [`crate::LockKeeperClient`], and if specified, the [`KeyId`] must
+    /// correspond to a key owned by the [`UserId`].
+    ///
+    /// Output: if successful, returns a [`String`] representation of the logs.
+    pub async fn retrieve_audit_event_log(
+        &self,
+        event_type: EventType,
+        options: Option<AuditEventOptions>,
+    ) -> Result<Vec<AuditEvent>, LockKeeperClientError> {
+        let mut client_channel = Self::create_channel(
+            &mut self.tonic_client(),
+            ClientAction::RetrieveAuditEvents,
+            self.account_name(),
+        )
+        .await?;
+        self.handle_retrieve_audit_events(&mut client_channel, event_type, options)
+            .await
+    }
+}
 
 /// Generate a new, distributed digital asset key with the given use
 /// parameters for the [`UserId`], and compatible with the specified blockchain.
@@ -112,28 +148,5 @@ pub fn retrieve_public_key_by_id(
     user_id: UserId,
     key_id: &KeyId,
 ) -> Result<KeyInfo, LockKeeperClientError> {
-    todo!()
-}
-
-/// Retrieve the log of audit events from the key server for a specified asset
-/// owner; optionally, filter for audit events associated with the specified
-/// [`KeyId`].
-///
-/// The log of audit events includes context
-/// about any action requested and/or taken on the digital asset key, including
-/// which action was requested and by whom, the date, details about approval or
-/// rejection from each key server, the policy engine, and each asset fiduciary
-/// (if relevant), and any other relevant details.
-///
-/// The [`UserId`] must match the asset owner authenticated in the
-/// [`crate::LockKeeperClient`], and if specified, the [`KeyId`] must correspond
-/// to a key owned by the [`UserId`].
-///
-/// Output: if successful, returns a [`String`] representation of the logs.
-#[allow(unused)]
-pub fn retrieve_audit_event_log(
-    user_id: UserId,
-    key_id: Option<&KeyId>,
-) -> Result<String, LockKeeperClientError> {
     todo!()
 }
--- a/lock-keeper-client/src/api/retrieve_audit_events.rs
+++ b/lock-keeper-client/src/api/retrieve_audit_events.rs
@@ -1,0 +1,27 @@
+use crate::{api::AuditEventOptions, LockKeeperClient, LockKeeperClientError};
+use lock_keeper::{
+    audit_event::{AuditEvent, EventType},
+    channel::ClientChannel,
+    types::retrieve_audit_events::{client, server},
+};
+
+impl LockKeeperClient {
+    pub(crate) async fn handle_retrieve_audit_events(
+        &self,
+        channel: &mut ClientChannel,
+        event_type: EventType,
+        options: Option<AuditEventOptions>,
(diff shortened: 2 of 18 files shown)


lock-keeper-tests/tests/end_to_end.rs
```diff
@@ -375,6 +391,43 @@ impl Test {
...
+
+        // Get audit event log
+        let audit_event_log = lock_keeper_client
+            .retrieve_audit_event_log(EventType::All, None)
+            .await?;
+        // Get the fourth last event, the last 3 are for retrieving audit logs and
+        // authenticating
+        let fourth_last = audit_event_log
+            .len()
+            .checked_sub(4)
```
vidyaap, line 419: This feels easily breakable, but I want a way to avoid the last few logs which would be about authenticating and starting the audit log retrieval.

picklenerd:
> changes the serialization library from bincode to serde_json

Just wanted to add a note that this change came about because `bincode` wasn't able to deserialize the log messages because it doesn't support serde's `deserialize_any` function. Instead of writing some hacks to work around this, we decided to switch to a library that we know won't ever have this problem.

Having JSON encoded-messages will make also troubleshooting much easier so this is a nice change either way. It will likely make the messages slightly larger but that's an issue we can deal with later. This is a non-breaking change since messages are only in this format during transit.

lock-keeper-client/src/api/retrieve_audit_events.rs
```diff
@@ -0,0 +1,27 @@
...
+    channel::ClientChannel,
+    types::retrieve_audit_events::{client, server},
+};
+
+impl LockKeeperClient {
+    pub(crate) async fn handle_retrieve_audit_events(
+        &self,
+        channel: &mut ClientChannel,
+        event_type: EventType,
+        options: Option<AuditEventOptions>,
```
picklenerd, line 13: This should not be wrapped in an `Option` since all of its fields are already `Option` types. We can derive `Default` on `AuditEventOptions` to handle situations where we don't want to set any options in code.

lock-keeper-key-server/src/database/audit_event.rs
```diff
@@ -27,4 +34,231 @@ impl Database {
...
+        &self,
+        account_name: &AccountName,
+        event_type: EventType,
+        options: Option<AuditEventOptions>,
+    ) -> Result<Vec<AuditEvent>, LockKeeperServerError> {
+        let actions = event_type.into_client_actions();
+        let mut query = doc! { ACTOR: account_name.to_string(), ACTION: {"$in": mongodb::bson::to_bson(&actions)?} };
+        if let Some(options) = options {
+            query = construct_query_with_options(options, query)?;
+        }
```
picklenerd, line 50: Also doesn't need to be wrapped in an `Option`. You can remove the `if let Some(options` block and just pass the options object along.

lock-keeper/src/audit_event.rs
```diff
@@ -54,3 +73,38 @@ impl Display for AuditEvent {
...
+            Self::SystemOnly => system_only,
+            Self::KeyOnly => ClientAction::iter()
+                .filter(|x| !system_only.contains(x))
+                .collect::<Vec<_>>(),
+        }
+    }
+}
+
+/// Optional parameters to filter [`AuditEvent`]s by
+#[derive(Debug, Serialize, Deserialize)]
```
picklenerd, line 105:
```suggestion
#[derive(Debug, Default, Serialize, Deserialize)]
```

lock-keeper/src/types/retrieve_audit_events.rs
```diff
@@ -0,0 +1,29 @@
...
+        audit_event::{AuditEventOptions, EventType},
+        impl_message_conversion,
+    };
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Debug, Deserialize, Serialize)]
+    /// Query specific set of audit event logs
+    pub struct Request {
+        pub event_type: EventType,
+        pub options: Option<AuditEventOptions>,
```
picklenerd, line 12:
```suggestion
        pub options: AuditEventOptions,
```

lock-keeper-client/src/api.rs
```diff
@@ -8,18 +8,54 @@
...
+    ///
+    /// The [`UserId`] must match the asset owner authenticated in the
+    /// [`crate::LockKeeperClient`], and if specified, the [`KeyId`] must
+    /// correspond to a key owned by the [`UserId`].
+    ///
+    /// Output: if successful, returns a [`String`] representation of the logs.
+    pub async fn retrieve_audit_event_log(
+        &self,
+        event_type: EventType,
+        options: Option<AuditEventOptions>,
```
picklenerd, line 46:
```suggestion
        options: AuditEventOptions,
```

lock-keeper-client/src/api/retrieve_audit_events.rs
vidyaap: Oh that makes sense; I just took this construction from the mongodb crate's optional parameters. So if the caller didn't want to use any of the filters, would they have to manually pass the default in?

picklenerd: Yep. I think this is actually preferable since the code will say `AuditEventOptions::default()` instead of `None` which makes it more obvious what's happening.

lock-keeper-tests/tests/end_to_end.rs
picklenerd: Could you `.filter()` those out by event type instead?

vidyaap: I think so? What I want is to know that the last event in the log for this user that is not re-authenticating or getting the audit logs has the correct action and status. I'm not sure if filtering by the expected action and checking the status of the latest one of those is the same.

lock-keeper/src/audit_event.rs
```diff
@@ -45,6 +46,24 @@ impl AuditEvent {
...
+
+    pub fn date(&self) -> DateTime {
+        self.date
+    }
+
+    pub fn status(&self) -> EventStatus {
+        self.status.clone()
+    }
+}
+
```
picklenerd, line 66: It's generally not recommended to call `clone` in getters since the caller can do this themself if they need to. `ClientAction` and `EventStatus` can derive `Copy` since they're just basic enums. 

`key_id()` can return `Option<&KeyId>` which you can get with `self.secret_id.as_ref()`. The caller can use `.cloned()` on the result if they need an owned `KeyId`.

lock-keeper-tests/tests/end_to_end.rs
picklenerd: I meant filter **out** the actions you don't want so that you can just grab the last action.

picklenerd: Per Slack, my suggestion wouldn't work if you're actually looking for one of those two actions.

lock-keeper-key-server/src/database/audit_event.rs
```diff
@@ -23,8 +30,242 @@ impl Database {
...
+        // RNG for this test
+        let mut rng = rand::thread_rng();
+
+        // Add a user
+        let (user_id, account_name) = create_user("key_id_filter_test", &mut rng, &db).await?;
+        // Create random audit events
+        let key_ids = create_random_audit_events(&account_name, &user_id, &mut rng, &db).await?;
+        // Retrieve audits for just one key
+        let options = AuditEventOptions {
+            key_ids: Some(key_ids[0..5].to_vec()),
```
picklenerd, line 201: Why is this limited to the first 5? Also, this will panic if someone changes `NUM_LOGS` to be less than 5. Not a big deal for tests but this function returns a result so it should probably use `.get()`.

lock-keeper-key-server/src/database/audit_event.rs
```diff
@@ -23,8 +30,242 @@ impl Database {
...
+        // starting
+        check_after_date(after_date_audit, after_date);
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn before_date_filter_works() -> Result<(), LockKeeperServerError> {
+        // Use timestamp as comparison date
+        let before_date = DateTime::now();
```
picklenerd, line 247: Could this do some stuff before the `before_date` so that there are results the check against?

picklenerd (commented):
Sorry this took so long for me to finish reviewing. I have a couple questions left about some tests but this looks good to me otherwise.

lock-keeper-key-server/src/database/audit_event.rs
vidyaap: This was just to request some of the keys logs but not all of them. I changed it to sample randomly from the key IDs from line 198 instead (haven't pushed yet).

lock-keeper-key-server/src/database/audit_event.rs
vidyaap: Yep, I changed the after_date one too, so that it excludes some. They both make the timestamp after the user is created now.

picklenerd (approved):
LGTM

vidyaap:
Thanks! 
@marsella or @indomitableSwan could you confirm whether the UserId needs to be present here, given the context from the first comment?

vidyaap:
> ... could you confirm whether the UserId needs to be present here, given the context from the first comment?

Added boltlabs-inc/key-mgmt-spec#132 to address the discrepancy over whether UserIds need to be present in the request.

PR #261, opened 2022-09-30
Create an interactive client for troubleshooting #254

### LINKED ISSUE #254: Interactive Client
The demo repo has proven incredibly useful for testing and troubleshooting but the user interface is a bit clunky outside of a demonstration environment.

It would be useful to have a client that can be used to perform quick operations in succession. It would also be useful if that client lived in the `key-mgmt` repo so that it could easily run with `LockKeeperClient` code that is being actively developed.

The client should run as an interactive terminal application that keeps some state to simplify user actions. The initial implementation can be a simple parser that inputs commands and performs actions with a `LockKeeperClient` while keeping state in memory.

In the future, we can add support to parse the same commands from files which could be used to define end to end tests.

For the first iteration, this application should:

- [x] Create a binary application within the `lock-keeper-client` crate
- [x] Write a parser for all supported operations such as `register [user-name] [password]`
- [x] Allow the user to specify a name for generated keys so that they can retrieve a key by name
- [x] Number unnamed keys sequentially starting with 1
- [x] Call `LockKeeperClient` functions based on parsed input

picklenerd:
Closes #254


jakinyele:
@picklenerd I'd love to test this when ready!

picklenerd:
This is mostly functional now but the local storage isn't persisting. It also needs documentation.

picklenerd:
Ayo had a great idea to hex encode all of these bytes. I'm working on an update.

lkic/src/command.rs
```diff
@@ -0,0 +1,281 @@
...
+            Command::Register {
+                account_name,
+                password,
+            } => {
+                let account_name = account_name.parse()?;
+                let password = password.parse()?;
+
+                LockKeeperClient::register(&account_name, &password, &state.config).await?;
+
+                println!("Logged in to {account_name}");
```
vidyaap, line 58: nit: I don't think we should say "logged in" here, just because in practice, someone would have to call `authenticated_client` before they can actually call anything else.

lkic/src/cli.rs
```diff
@@ -0,0 +1,14 @@
...
+use std::path::PathBuf;
+
+#[derive(Debug, Parser)]
+pub struct Cli {
+    /// Location of client config file.
+    #[clap(long, default_value = "dev/local/Client.toml")]
+    pub config: PathBuf,
+    /// Directory where persistent storage files will be saved.
+    #[clap(long, default_value = "dev/lkic_data")]
+    pub storage_path: PathBuf,
```
vidyaap, line 13: I had to manually make the "lkic_data" folder in the dev folder before `cargo make lkic` would run. Am I supposed to do that or is something else supposed to generate that folder?

picklenerd: No it's supposed to create the folder automatically. I'll take a look.

picklenerd: Updated the `load` function to also create the directory. This should work now.

lkic/README.md
```diff
@@ -0,0 +1,132 @@
...
+## Usage
+
+Start the server with `cargo make start`, then run the client with `cargo make lkic`.
+
+Note that this tool is built using the current code from the `lock-keeper-client` crate. This makes it useful for testing changes to the client
+but it also means that it will break alongside the client crate.
+
+
+## Commands
+To see this list in the application, type `help` in the `lkic` prompt.
```
vidyaap, line 16: Should the `list` command be included here too?

lkic/src/command.rs
vidyaap: Ah okay, that makes sense. I think I'm just getting hung up on the details because spec-wise, just registering without authenticating shouldn't give you a client. It's maybe not worth it to stick to that here if it's just for testing. (unless we can test that part of the functionality by doing so?)

lkic/README.md
picklenerd: Yes! I added it here and to the help text that's printed in the app. Thanks!

lkic/src/command.rs
picklenerd: Oh yeah this definitely isn't a 1:1 mapping to the spec. I think it would be detrimental to the user experience to require authentication after registration. One of the big ideas here coming from the demo app is that you only have to input your username and password once.

vidyaap (approved):
I didn't see any other comments on the location of this crate, so this looks good to me!

PR #262, opened 2022-10-04
Add ECDSA signing

### LINKED ISSUE #235: Implement ECDSA signing

marsella:
Closes #235 

TODO
- The interface for converting to / from `Vec<u8>` for `SigningKeyPair`s (as well as other types not addressed in this PR) is quite ugly. Per discussion with @picklenerd, @gatoWololo, and @vidyaap, we should try to write a generic little parser to read a length, then read that many bytes and return `(len, bytes, remaining_bytes)`, and apply this to `SigningKeyPair`, `StorageKey`,  `AssociatedData`, `EncryptionKey`, and `Secret`. On second thought, maybe this should be its own issue.

- That same interface has a ton of error handling that all ends up returning the same error. Try wrapping this in a lambda, throwing up the errors with `?`, then converting to `CryptoError::ConversionError` exactly once.

- Review types that will actually get passed as messages to `sign`/`verify`. Can they support the type bound (`AsRef<[u8]>`)?


indomitableSwan:
> Closes #235
> 
> TODO
> 
> * The interface for converting to / from `Vec<u8>` for `SigningKeyPair`s (as well as other types not addressed in this PR) is quite ugly. Per discussion with @picklenerd, @gatoWololo, and @vidyaap, we should try to write a generic little parser to read a length, then read that many bytes and return `(len, bytes, remaining_bytes)`, and apply this to `SigningKeyPair`, `StorageKey`,  `AssociatedData`, `EncryptionKey`, and `Secret`. On second thought, maybe this should be its own issue.
> * That same interface has a ton of error handling that all ends up returning the same error. Try wrapping this in a lambda, throwing up the errors with `?`, then converting to `CryptoError::ConversionError` exactly once.
> * Review types that will actually get passed as messages to `sign`/`verify`. Can they support the type bound (`AsRef<[u8]>`)?
> 
> Cryptography notes
> 
> * I'm using the `sign_with_rng` method because the plain `sign` method produces non-randomized signatures, and I don't quite understand why that's safe. Is there some psudo-Fiat-Shamir way to get `k` without an RNG?

Yes, although I wouldn't call the technique used Fiat-Shamir (since Fiat-Shamir turns an interactive protocol into a non-interactive one). See [RFC 6979](https://datatracker.ietf.org/doc/html/rfc6979). There are many advantages to using deterministic ECDSA; historically, I believe many attacks on ECDSA implementations were a consequence of failing to choose the nonce appropriately. I advise switching to the plain `sign` method.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -332,15 +448,20 @@ mod test {
...
+        // Key material must be the right size
+        let not_enough_key_material: [u8; 12] = rng.gen();
+        let short_import: Import = not_enough_key_material.as_ref().into();
+        assert!(short_import.into_signing_key(&user_id, &key_id).is_err());
 
-        // TODO #235: test any additional constraints on the key material.
+        let too_much_key_material: Vec<u8> =
+            std::iter::repeat_with(|| rng.gen()).take(64).collect();
+        let long_import: Import = too_much_key_material.as_slice().into();
+        assert!(long_import.into_signing_key(&user_id, &key_id).is_err());
```
marsella, line 464: @indomitableSwan Are there other properties off the top of your head that we should check must The goal here is to check that key material has to be a "suitable" / canonical key, but I wasn't sure if there's a canonical expectation for formatting it (e.g. should it fail if it's larger than the modulus, or does it just get modded).

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -1,25 +1,34 @@
...
 use rand::{CryptoRng, RngCore};
 use serde::{Deserialize, Serialize};
 use std::marker::PhantomData;
 
 use super::{generic::AssociatedData, CryptoError, Encrypted, KeyId, StorageKey};
 
+/// Things that can be signed must implement the `Signable` trait.
+pub trait Signable: AsRef<[u8]> {}
+
+impl<T> Signable for T where T: AsRef<[u8]> {}
```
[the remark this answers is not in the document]
picklenerd: If it's a marker trait, I don't think we should do this blanket implementation for `AsRef<[u8]>`. We should specifically mark the types we want.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -164,7 +181,10 @@ impl Import {
     ///
     /// This is part of the flow to send an imported key in cleartext to the key
     /// server and must be called by the server.
-    #[allow(unused)]
+    ///
+    /// This will fail if `material` is not a scalar value formatted in big
+    /// endian. See [k256 documentation](https://docs.rs/k256/latest/k256/ecdsa/struct.SigningKey.html#method.from_bytes)
+    /// for details.
```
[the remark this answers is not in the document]
indomitableSwan: I think the additional check makes sense.

lock-keeper/src/crypto/signing_key.rs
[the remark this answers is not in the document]
indomitableSwan: @marsella let's discuss this again.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -1,54 +1,70 @@
 use crate::{types::database::user::UserId, LockKeeperError};
+use k256::{
+    ecdsa::{self, SigningKey, VerifyingKey},
+    schnorr::signature::{Signer, Verifier},
```
indomitableSwan, line 4: Err, why are we using Schnorr?

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -1,54 +1,70 @@
...
 #[allow(unused)]
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct SigningPublicKey;
+pub struct SigningPublicKey(VerifyingKey);
 
 /// Temporary type to represent a remotely generated encrypted
-/// [`SigningKeyPair`]
+/// [`SigningKeyPair`].
+///
+/// This should only be "decrypted" by the server.
```
indomitableSwan, line 36: "Should" or "can"? This documentation is a bit weird.

indomitableSwan: Do we have to remember to fix this later?

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -1,54 +1,70 @@
...
+    type Error = LockKeeperError;
+    fn try_from(key_pair: PlaceholderEncryptedSigningKeyPair) -> Result<Self, Self::Error> {
+        Ok(Self {
+            signing_key: SigningKey::from_bytes(key_pair.signing_key.as_slice())
+                .map_err(|_| CryptoError::ConversionError)?,
             context: key_pair.context,
-        }
+        })
     }
 }
```
indomitableSwan, line 61: @marsella let's discuss.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -59,20 +75,23 @@ impl SigningKeyPair {
...
+    #[cfg(test)]
+    fn public_key(&self) -> SigningPublicKey {
+        SigningPublicKey(self.signing_key.verifying_key())
     }
 
+    /// Retrieve the context associated with the signing key.
+    #[cfg(test)]
     fn context(&self) -> &AssociatedData {
         &self.context
     }
```
indomitableSwan, line 89: Don't we want getters for both the public key and the context?

lock-keeper/src/crypto/signing_key.rs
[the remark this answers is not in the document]
indomitableSwan: We do have an expected canonical key format for import, don't we? I'm not seeing this here. What am I missing?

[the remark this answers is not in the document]
indomitableSwan: Note to self: discuss this function with @marsella

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -183,45 +210,54 @@ impl Import {
...
+            context,
+        })
     }
 }
 
 /// Raw material for an exported signing key.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Export {
     pub key_material: Vec<u8>,
+    pub context: Vec<u8>,
```
indomitableSwan, line 227: Same question about canonical formats here, too.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -183,45 +210,54 @@ impl Import {
...
+        let signing_key = SigningKey::from_bytes(self.key_material.as_slice())
+            .map_err(|_| CryptoError::ConversionError)?;
+        let context = self.context.try_into()?;
         Ok(SigningKeyPair {
-            context: associated_data,
+            signing_key,
+            context,
         })
     }
 }
```
indomitableSwan, line 251: note to self: discuss with @marsella

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -243,38 +279,52 @@ impl TryFrom<Vec<u8>> for SigningKeyPair {
...
+            SigningKey::from_bytes(signing_key_bytes).map_err(|_| CryptoError::ConversionError)?;
+
         // AssociatedData `try_into` handles length prepending
-        let context_offset = separator_offset;
+        let context_offset = signing_key_end;
         let context_bytes = value
             .get(context_offset..)
             .ok_or(CryptoError::ConversionError)?
             .to_vec();
         let context: AssociatedData = context_bytes.try_into()?;
```
indomitableSwan, line 300: note to self: discuss with @marsella

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -296,6 +346,31 @@ impl Encrypted<SigningKeyPair> {
 mod test {
```
indomitableSwan, line 346: What about checking for correct formatting in conversions?

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -311,6 +386,49 @@ mod test {
...
+        let import: Import = raw_bytes.as_slice().into();
+
+        let user_id = UserId::new(&mut rng)?;
+        let key_id = KeyId::generate(&mut rng, &user_id)?;
+        let output_key: SigningKeyPair = import.into_signing_key(&user_id, &key_id)?;
+
+        // Make sure the key material matches. The associated data doesn't have to match
+        // because it's manually incorrect in the original key.
+        assert_eq!(key.signing_key, output_key.signing_key);
+        Ok(())
```
indomitableSwan, line 417: note to self: discuss with @marsella

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -331,28 +449,77 @@ mod test {
...
             .map(|len| -> Vec<u8> { std::iter::repeat_with(|| rng.gen()).take(len).collect() })
-            .map(|msg| signing_key.sign(&msg))
-            .all(
-                |sig| trivial_signature == sig && sig.verify(public_key, &generic_message).is_ok()
-            ));
+            .map(|msg| (signing_key.sign(&msg), msg))
+            .all(|(sig, msg)| sig.verify(&public_key, &msg).is_ok()));
+    }
+
+    #[test]
```
indomitableSwan, line 466: This test has a strange name. Do you mean 'verifying' here?

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -331,28 +449,77 @@ mod test {
...
+        let message = b"signatures won't verify with a bad message".to_vec();
+        let sig = signing_key.sign(&message);
+
+        let bad_msg = b"this is obviously not the same message".to_vec();
+        assert!(sig.verify(&public_key, &bad_msg).is_err());
+        assert!(sig.verify(&public_key, &message).is_ok());
+    }
+
+    #[test]
+    fn signing_requires_correct_public_key() {
```
indomitableSwan, line 481: Same comment here: you're talking about verifying signatures, not signing.

indomitableSwan (commented):
Question:
- Do we already have issues to track the need to import and export arbitrary secrets?
- The way we're building out the cryptography still seems a bit heavy-handed to me, in that we have more abstraction layers between the cryptography and the rest of the software than I believe is warranted by the functionalities we are implementing here. I am still concerned that this code is more complex (and therefore error-prone) than it needs to be.

lock-keeper/src/crypto/signing_key.rs
indomitableSwan: Please point me to the issue that addresses the need to fix this later.

lock-keeper/src/crypto/signing_key.rs
indomitableSwan: Also: spot check dependency to make sure our dependency is doing proper input validation.

lock-keeper/src/crypto/signing_key.rs
indomitableSwan: discussion comment: might want to do test of associated data correctness here, too.

lock-keeper/src/crypto/signing_key.rs
marsella: @picklenerd Do you know which issue this is?

lock-keeper/src/crypto/signing_key.rs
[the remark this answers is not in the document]
marsella: Added the check and made #306 to define a canonical format.

lock-keeper/src/crypto/signing_key.rs
marsella: We talked about trying to write tests that change the length byte and fail if it's not the right value. I added a note about this to #292 to add it across the board. I'll play around a bit and try to add it for signing keys, but I suspect the easiest thing will be to hardcode the location of the length byte in the test, which seems fragile. Open to dev ideas on this.

lock-keeper/src/crypto/signing_key.rs
marsella: Added AD check here and updated the "labelled" test to make sure all generated key types have exactly only the correct generation type.

lock-keeper/src/crypto/signing_key.rs
marsella: See #306

lock-keeper/src/crypto/signing_key.rs
marsella: Fixed name.

lock-keeper/src/crypto/signing_key.rs
marsella: Fixed name.

lock-keeper/src/crypto/signing_key.rs
picklenerd: I don't think there's an issue for replacing this type with something properly encrypted. This was always sort of up in the air while we figure out how to actually do it.

I think the comment was just referring to the use of "should" implying that anyone can decrypt this but only the key server actually should. This is one of those cases where the decryption code should be in the server crate if possible. Then we can definitively say that it can only be decrypted by the server.

lock-keeper/src/crypto/signing_key.rs
marsella: See #306 for canonical formatting.

The dependency
- checks that it's non-zero (twice)
- checks that it's not larger than the curve max byte size
- checks that the value is not larger than the curve modulus

The first two checks are here:
https://docs.rs/elliptic-curve/latest/src/elliptic_curve/secret_key.rs.html#148-163

The last is here:
https://docs.rs/elliptic-curve/latest/src/elliptic_curve/scalar/core.rs.html#77-84

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -89,11 +101,15 @@ impl SigningKeyPair {
     /// by the client. In this flow, the key server will only receive an
     /// [`Encrypted<SigningKeyPair>`], not the cleartext.
     ///
+    /// `key_material` should be a scalar value formatted in big endian. See
+    /// [k256 documentation](https://docs.rs/k256/latest/k256/ecdsa/struct.SigningKey.html#method.from_bytes)
+    /// for details.
+    ///
```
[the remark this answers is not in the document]
marsella: See #306 to improve this documentation.

lock-keeper/src/crypto/signing_key.rs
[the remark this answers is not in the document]
marsella: I implemented this just for `Vec<u8>` and with a `#[cfg(test)]` flag. We can now add types as we figure out what to sign. This may cause a problem in the end-to-end tests when we implement signing. Can make the impl public at that time or maybe, say, make a custom `TestMessage` type to sign that impls this.

lock-keeper/src/crypto/signing_key.rs
marsella: Hmm. I think I'll make an epic about this and specifically flag this bit of code with a TODO. It would be very bad to ship a product that doesn't actually encrypt internal storage. In my mind, the to-do list is pretty straightforward here, it's more a question of when; I would suggest in the next release. E.g.
- write up something in the spec with a concrete decision about encryption scheme for server storage (likely: the encryption scheme we already implemented)
- decide how to store server-side encryption keys (this seems like the most open question)
- Add crypto API: define a server encryption key (wrapper around existing encryption key) and a function to encrypt + decrypt `SigningKeyPair`s with it.
- call the crypto API when we add / remove keys from storage, and when the server initializes (to create its signing key).

marsella: Made #307 and put a reference to it here.

lock-keeper/src/crypto/signing_key.rs
indomitableSwan: Nothing comes to mind immediately on this.

indomitableSwan (approved):
LGTM.

PR #263, opened 2022-10-05
148 export

### LINKED ISSUE #148: Implement export key flows for LockKeeperClient
**Background:** 
   - The export protocol is loosely incorporated into the pre-existing "retrieve a secret" functionality. Refer to the [relevant epic](https://app.zenhub.com/workspaces/bolt-team-60886be9e01f13000e361d92/issues/boltlabs-inc/key-mgmt/66). 
   - Refer to the spec for the export context under [Retrieve a Secret](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/system-functionalities.md#retrieve-a-secret)

**Requirements:**

   - Add a new public function to the local client (formerly called the DamsClient) API that allows the caller to export a key
   - We would prefer to have a separate API, even if this API calls the retrieve API under the hood
   - API function name: TBD (example: `export_key`)
   - Inputs: KeyId, DeleteOnExport (boolean)
1. DeleteOnExport: In the initial implementation of the API, exported keys will not be deleted from the Key Server
   - We are proposing to expose this option in the API for future enhancements
   - The default value for DeleteOnExport is false, meaning that the exported key will not be deleted from the Key Server
   - If the caller attempts to set DeleteOnExport to true in the Passive Keys PoC, then an error shall be returned, because this feature is not yet implemented
1. Client Action: Add a new ClientAction type to send a message between LockKeeperClient and Key Server to request a key export.
1. New Server Feature: Implement feature that will process the export request as follows:
   - retrieve the key from secure storage
   - return exported key to the caller
       - (In progress, I'm reading the spec) Where is the key decrypted?
   - import a key
   - export a key
   - compare exported key to the originally imported key

**Design Thoughts:** TBD: Jay is working on a diagram, which will be linked to Confluence here.

vidyaap:
Closes #148 

This just creates a client function called `export_key` which calls retrieve internally. The only server-side change is adding the action name to the tonic metadata to handle action names that don't have their own protobuf definitions (e.g. the way Export doesn't).

@jwantz Now that I'm looking at it, I'm wondering if it makes sense to put the delete parameter in at all, if we're leaving it unimplemented. I'd rather not have the parameter there and add it back in once we decide how to handle deletion. Thoughts?


lock-keeper-client/src/api.rs
```diff
@@ -12,18 +12,50 @@ pub(crate) mod retrieve_audit_events;
...
-    ClientAction,
+    ClientAction, RetrieveContext,
 };
 
 impl LockKeeperClient {
+    /// Export key material from the key servers. If `delete` is true, the key
+    /// will be deleted from the key server.
+    ///
+    /// Output: If successful, returns the requested key material in byte form.
+    pub async fn export_key(
```
picklenerd, line 31: I think this should go in the main client definition based on the current code layout. Alternatively, we could put all of the public API functions for the client in here.

jwantz:
> Closes #148
> 
> @jwantz Now that I'm looking at it, I'm wondering if it makes sense to put the delete parameter in at all, if we're leaving it unimplemented. I'd rather not have the parameter there and add it back in once we decide how to handle deletion. Thoughts?

@vidyaap: I think I agree with you. My only concern is for backward compatibility when we make the change to include the additional parameter. As long as we're thinking ahead, so that we don't break any pre-existing code that integrates with our library, then all is OK.

lock-keeper-client/src/api.rs
vidyaap: I kind of like the API functions in their own module, but I can probably get rid of the `arbitrary_secrets` submodule. I think originally, I kept register and authenticate in `client` because they were like the "constructors" for the client, but I can move those too if that makes more sense.

picklenerd: Yeah I like that too. I think it makes sense to move all of the public functions here even though register and authenticate are a little weird. It makes this module a centralized location for the public interface.

indomitableSwan:
> 

@vidyaap makes a good point, but we aren't expecting the API to be stable after this phase, right? I think we need to settler on a repo-wide strategy/plan for moving from an unstable to stable API in the future. i.e., I agree that it makes sense to drop the deletion parameter for now.

indomitableSwan:
> > Closes #148
> > @jwantz Now that I'm looking at it, I'm wondering if it makes sense to put the delete parameter in at all, if we're leaving it unimplemented. I'd rather not have the parameter there and add it back in once we decide how to handle deletion. Thoughts?
> 
> @vidyaap: I think I agree with you. My only concern is for backward compatibility when we make the change to include the additional parameter. As long as we're thinking ahead, so that we don't break any pre-existing code that integrates with our library, then all is OK.

@vidyaap makes a good point, but we aren't expecting the API to be stable after this phase, right? I think we need to settler on a repo-wide strategy/plan for moving from an unstable to stable API in the future. i.e., I agree that it makes sense to drop the deletion parameter for now.

lock-keeper-tests/src/end_to_end.rs
```diff
@@ -255,6 +279,39 @@ impl Test {
...
+                        &self.account_name,
+                        &self.password,
+                        config,
+                    )
+                    .await?;
+
+                    // Get KeyId from state and run export
+                    let key_id_json = self.state.get(GENERATED_ID)?;
+                    let key_id: KeyId = serde_json::from_value(key_id_json.clone())?;
+                    let res = lock_keeper_client.export_key(&key_id, false).await?;
```
marsella, line 294: Thought: Same comment I've had before on tests - this only lets us test the "happy" case, where we're exporting  key we already generated. Is there an existing issue about the test suite where we can add this as a case that we'd like to expand on?

marsella (commented):
This looks good to me.

marsella (commented):
Does anything need to change in the existing retrieve method to remove the ability to do an export at that point?

marsella:
> Does that answer the question?

Yes, thank you!

vidyaap:
@jwantz @indomitableSwan I think it's okay for now if the change is breaking in the future, depending on how far out we're intending to push spec-ing the deletion. What do you think of Marcella's enum suggestion?

jwantz:
> @jwantz @indomitableSwan I think it's okay for now if the change is breaking in the future, depending on how far out we're intending to push spec-ing the deletion. What do you think of Marcella's enum suggestion?

I like Marcella's suggestion. It allows us to be non-breaking in the future and to also enhance the capabilities for what might happen after an export. That being said, I do not have a compelling argument to simply remove the additional parameter from this phase, since we are not committing to lock down our remote API at the conclusion of this phase. I know I am not strictly answering your question, and that is because I don't have strong enough feelings. If I had to pick, I would personally go with Marcella's suggestion.

indomitableSwan:
> @jwantz @indomitableSwan I think it's okay for now if the change is breaking in the future, depending on how far out we're intending to push spec-ing the deletion. What do you think of Marcella's enum suggestion?

The enum suggestion seems fine to me. I'm leaning towards removing the deletion option though. I'd rather we try to narrow in on the underlying use cases/requirements here. For example, I could see us settling on `export_key` _always_ deleting the key material, and having a different external-facing API for the use case in which you want to say, backup your key without removing it from the key server.

lock-keeper-client/src/client.rs
```diff
@@ -244,15 +244,19 @@ impl LockKeeperClient {
         let (tx, rx) = mpsc::channel(2);
         let mut stream = Request::new(ReceiverStream::new(rx));
 
+        // Set AccountName and Action in metadata
         let account_name_val = MetadataValue::try_from(account_name.to_string())?;
+        let action_val = MetadataValue::try_from(format!("{:?}", action))?;
         let _ = stream.metadata_mut().insert(ACCOUNT_NAME, account_name_val);
+        let _ = stream.metadata_mut().insert(ACTION, action_val);
```
picklenerd, line 179: I like that this is included in the requests now but I feel like we might be able to extract it from the request itself. Maybe from the request path or something. I don't think it's in the `tonic` request type but it might be in the `hyper` request type.

This definitely doesn't need to be part of this PR if it's not an easy change.

vidyaap:
Re: delete parameter, since there are no strong feelings against removing the parameter, I think I'm going to go with that, especially if there is a possibility that deleting the key won't be optional later.

lock-keeper-key-server/src/server.rs
```diff
@@ -99,7 +121,7 @@ impl LockKeeperRpc for LockKeeperKeyServer {
         &self,
         request: Request<tonic::Streaming<Message>>,
     ) -> Result<Response<Self::RegisterStream>, Status> {
-        let context = self.context(&request, ClientAction::Register)?;
```
picklenerd, line 102: 🎉

lock-keeper-client/src/client.rs
vidyaap: Oh that sounds interesting! Not sure the path would help though, since the issue was that we're still calling retrieve even though the action is export. I do like how the hyper request type looks though - more explicit types for headers instead of the string key-values we use for metadata here.

lock-keeper-client/src/api.rs
```diff
@@ -12,18 +12,50 @@ pub(crate) mod retrieve_audit_events;
...
+        &self,
+        key_id: &KeyId,
+        delete: bool,
```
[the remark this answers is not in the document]
vidyaap: Resolving based on comment thread!

lock-keeper-client/src/client.rs
picklenerd: Yeah it's actual HTTP request so it might come in handy down the line. I'll resolve this. I think we can look at the hyper headers down the line when we have more service layers.

lock-keeper-tests/src/end_to_end.rs
marsella: Can you access public lock-keeper methods here? You might be able to use the normal constructor `KeyId::generate(rng, user_id)`.

marsella: Yes, just make a fresh one. You can use `rand::thread_rng()` or `rand::StdRng::from_entropy()`.

vidyaap: Added! Fyi, it ended up being a little clunky because of the way returning the errors from that match block works

picklenerd (approved):
LGTM. We're definitely starting to stretch our test harness here. I'm going to make an epic where we can start to collect our thoughts.

lock-keeper-tests/src/end_to_end.rs
marsella: Great, thank you! Looks good to me, thanks for also adding the retrieve test.

PR #267, opened 2022-10-06
Add #[ignore] to database unit tests #266

### LINKED ISSUE #266: Add #[ignore] to database unit tests
It's currently not possible to run `cargo test` without MongoDB running due to the database unit tests in `lock_keeper_key_server/database`. Unit tests should be running without any other pre-conditions.

We can add an `#[ignore]` block to these tests so that they don't run by default.

This is a short-term quick fix. Longer-term, we should find a way to move these tests into `lock-keeper-tests` without exposing private functions unnecessarily.

picklenerd:
Closes #266

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -60,7 +60,7 @@ jobs:
         run: cargo test --all-features --lib --bins --workspace
 
       - name: Run all doctests
-        run: cargo test --all-features --doc --verbose --workspace
+        run: cargo test --all-features --doc --verbose --workspace -- --ignored
 
       - name: Build documentation
         run: RUSTDOCFLAGS="-Dwarnings" cargo doc --all-features --no-deps --verbose --workspace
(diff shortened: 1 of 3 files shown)


gatoWololo (approved):
Looks good!

As far as a long term solution: What about creating a `mongodb-tests` feature? And we can have all these unit tests live under a module which is only compiled when `mongodb-tests` feature is enabled?

picklenerd:
> Looks good!
> 
> As far as a long term solution: What about creating a `mongodb-tests` feature? And we can have all these unit tests live under a module which is only compiled when `mongodb-tests` feature is enabled?

I don't think these tests are in the spirit of unit tests so I think I'd still prefer it if we moved them to the integration tests.

gatoWololo:
> I don't think these tests are in the spirit of unit tests so I think I'd still prefer it if we moved them to the integration tests.

Makes sense. I saw #268 and I agree.

PR #269, opened 2022-10-06
Added cargo make tasks for CI #265

### LINKED ISSUE #265: Create CI task in Makefile.toml
We can define all our CI tasks in the `cargo make` definition which would allow us to easily run all of the CI checks locally before pushing. The CI script can be updated to run the single CI task which will keep local testing in sync with CI testing.

Most of these tasks will be straightforward. Formatting will require an extra line to make sure it runs on nightly

```rust
[tasks.format]
toolchain = "nightly"
command = "cargo"
args = ["fmt"]
```

- [ ] Add new tasks to `Makefile.toml` for each check that we want to run in CI
- [ ] Add a new task called `[tasks.ci]` to `Makefile.toml` that has all of the other tasks as dependencies
- [ ] Update `.github/workflows/rust.yml` to call `cargo make ci` (see `e2e-test.yml` for a cargo make example)

picklenerd:
This PR adds some new tasks to `Makefile.toml` for use in CI. CI now calls `cargo make ci` to run all checks. This also allows developers to run `cargo make ci` locally before pushing to Github so that they can feel confident that their commits will pass CI checks.

`cargo make ci` currently requires MongoDB to be running on your local machine due the the database unit tests.

Closes #265

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -47,20 +47,11 @@ jobs:
         with:
           version: '3.x'
 
-      - name: Format check
-        run: cargo +nightly fmt --all -- --check
+      - name: Install cargo-make
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: --debug cargo-make
 
-      - name: Clippy lint
-        run: cargo clippy --all-features --all-targets --verbose --workspace -- -Dwarnings
-
-      - name: Build
-        run: cargo build --all-features --all-targets --verbose --workspace
-
-      - name: Run all unit + library tests (not doc, integration, benchmark, or example tests)
-        run: cargo test --all-features --lib --bins --workspace
-
-      - name: Run all doctests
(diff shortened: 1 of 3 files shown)


vidyaap (approved):
Thanks, this is super helpful! Looks good to me, I just had a question: some of these used to be run with the verbose flag, and I was curious why we don't have that anymore? (they do still seem to provide the relevant errors when they fail though, even without verbose)

picklenerd:
> Thanks, this is super helpful! Looks good to me, I just had a question: some of these used to be run with the verbose flag, and I was curious why we don't have that anymore? (they do still seem to provide the relevant errors when they fail though, even without verbose)

I removed it because I've never found it useful for general cases like CI. It adds a massive amount of noise to the CI logs when you're just trying to see what went wrong. I think it's much more useful when you're troubleshooting locally.

We can always add it back in the future if we think it will be useful.

PR #273, opened 2022-10-06
Implement remote signing key generation #187

### LINKED ISSUE #187: Implement remote generation of signing keys for LockKeeperClient
**Background:** In the Arbitrary Secrets PoC, we provide the capability to locally generate (and encrypt) a key on the client, followed by sending it to the Key Server for secure storage. Our intention is to add a new capability to the Key Server, such that the key generation (and encryption) occurs remotely on the server. Refer to this PR on the Key Management spec for more details: https://github.com/boltlabs-inc/key-mgmt-spec/pull/111.

**Requirements:**

   - Add a new public function to the local client (formerly called the DamsClient) API that allows the caller to perform a remote generation of a signing key
   - API function name: TBD (example: `generate_signing_key`)
   - Outputs: KeyId, LockKeeperClientError
1. Client Action: Add a new ClientAction type to send a message between LockKeeperClient and Key Server to request a remote key generation.
1. New Server Feature: Implement feature that will process the remote key generation request as follows:
   - generate a signing key pair
       - This secret MUST be randomly generated according to the uniform distribution (refer to spec)
       - This secret should default to 256-bits (refer to spec)
   - encrypt the key pair
   - store in secure storage
   - return key_id and public_key to the caller
   - generate, encrypt and store key
   - retrieve and decrypt key
   - compare retrieved key to the originally generated key

**Design Thoughts:** Refer to design diagram in Confluence (associated with Remote API Design).

**Future Thoughts (not part of this ticket):**  We may want to add this capability to the Demo App and/or interactive client. If so, then a new ticket should be created and associated with this issue.

picklenerd:
Implements remote generation of signing keys. We don't have an encryption scheme defined for remotely generated keys yet so these are currently stored in the database as unencrypted bytes.

This PR also includes some cleanup items:
- Added more `strum` derives to `ClientAction` for conversion to/from strings
- Added operation tracing to server

Closes #187


lock-keeper/src/types/database/secrets.rs
```diff
@@ -0,0 +1,42 @@
...
+            retrieved: false,
+        }
+    }
+}
+
+/// Wrapper around a server generated secret
+/// WARNING: This type is temporary. Do not use in production!
+#[derive(Debug, Deserialize, Serialize)]
+pub struct ServerGeneratedSecret {
+    pub secret: Vec<u8>,
```
picklenerd, line 29: @indomitableSwan  https://www.youtube.com/watch?v=3tmd-ClpJxA

This is awful and wrong but I don't know of a good way to encrypt this with what we have right now without having the encryption key stored in the same database table.

lock-keeper-key-server/src/database/secrets.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// TODO: This function temporarily stores an unencrypted key pair.
+    /// WARNING: Do not use in production!
+    pub async fn add_remote_secret(
+        &self,
+        user_id: &UserId,
+        secret: SigningKeyPair,
+        key_id: KeyId,
+    ) -> Result<(), LockKeeperServerError> {
+        let collection = self.inner.collection::<User>(constants::USERS);
+        let stored_secret = ServerGeneratedSecret::new(secret.into(), key_id);
```
picklenerd, line 76: @indomitableSwan @marsella This is where the storage type for the `SigningKeyPair`. It takes the raw `SigningKeyPair` bytes and stores them directly for now.

vidyaap: FYI, I think my import code will help will this! I can make a draft PR in a second so you can see

vidyaap: Or actually, it's Marcella's `Import` type and related code.

lock-keeper/src/types/database/secrets.rs
indomitableSwan: > https://www.youtube.com/watch?v=3tmd-ClpJxA

Hmm. Yes. This is why I had been thinking that leaving things unencrypted for now would not be the worst idea: just as broken, only more evidently so.

picklenerd: Ok cool. We can leave this for now and come back to it later! Resolving.

lock-keeper-key-server/src/database/secrets.rs
picklenerd: Ok I updated the code to use the same types from your PR.

lock-keeper-key-server/src/operations/remote_generate.rs
```diff
@@ -0,0 +1,46 @@
...
+pub struct RemoteGenerate;
+
+#[async_trait]
+impl Operation for RemoteGenerate {
+    async fn operation(
+        self,
+        channel: &mut ServerChannel,
+        context: &mut Context,
+    ) -> Result<(), LockKeeperServerError> {
+        let request: client::RequestRemoteGenerate = channel.receive().await?;
```
marsella, line 23: todo: I expected a check here that the `request.user_id` matches the account from (I think) the `Context` -- e.g. to prevent user A authenticating as themselves, but then creating a key for user B.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -11,7 +11,7 @@ use super::{generic::AssociatedData, CryptoError, Encrypted, KeyId, StorageKey};
 ///
 /// This can be generated locally by the client or remotely by the server.
 #[allow(unused)]
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
```
marsella, line 14: todo: I'd like documentation on this type to the effect of "it's serializable for the database, but don't send it in the clear over a network". Or just the latter half if you think it's more reasonable.

lock-keeper/src/types/database/secrets.rs
```diff
@@ -1,18 +1,25 @@
...
+use crate::crypto::{Encrypted, KeyId, Secret, SigningKeyPair};
 use serde::{Deserialize, Serialize};
 
+/// Holds user's stored secrets of all types
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct StoredSecrets {
+    pub arbitrary_secrets: Vec<StoredEncryptedSecret>,
+    pub signing_keys: Vec<StoredEncryptedSigningKeyPair>,
+    pub server_generated_signing_keys: Vec<StoredSigningKeyPair>,
+}
```
marsella, line 12: thought: This structure seems like every time we add a new type of stored secret, we have to update our retrieve code to look through another list. I know there's a plan to discuss better database structure, but is it a worthwhile intermediate step to pull out the key ids from this so it's easier to search on them? Like

```rust
// don't use this name
pub struct AnyKindOfSecret {
    key_id: KeyId
    retrieved: bool,
    secret: StoredKind,
}
enum StoredKind {
    EncryptedSecret(Encrypted<Secret>),
    EncryptedSigningKeyPair(Encrypted<SigningKeyPair>),
    SigningKeyPair(SigningKeyPair),
}
```

lock-keeper-tests/src/end_to_end.rs
```diff
@@ -249,6 +249,23 @@ async fn tests() -> Vec<Test> {
                 ),
             ],
         ),
+        Test::new(
```
marsella, line 252: todo: Can you add a test to retrieve a remote generated secret? Is that a separate issue?

marsella (changes requested):
There's a missing server-side check that the user id matches the user making the request. My other comments are less important.

This was very easy to review -- all the refactoring that you all have done has been very successful in making this super clearly delineated.

lock-keeper-key-server/src/operations/remote_generate.rs
picklenerd: I can add that.

I think these requests containing user_ids will be redundant once we have sessions. If the user is authenticated and has a session key that decrypts the message correctly, then we know their user_id and we know that they're the one that authenticated.

lock-keeper/src/crypto/signing_key.rs
picklenerd: I think we could do bson conversions instead of serialize/deserialize. I'll investigate.

lock-keeper-key-server/src/operations/remote_generate.rs
vidyaap: Should that check be added in the other requests too, until we have sessions? If so, I'm happy to make those changes to all the endpoints at once, if that's easier?

lock-keeper/src/types/database/secrets.rs
picklenerd: We avoided the enum because we don't want to add error handling for situations where you need an `EncryptedSecret` but you have to make sure it's not the other variants. Having it this way means that the types are always correct for the secret you're trying to retrieve. If the database data gets corrupted and the data for a given key id changes to a different type, the whole user will fail to load instead of having a silent failure that doesn't show up until the key is retrieved.

This struct is temporary until we have a more permanent solution for how we represent secrets in the database.

picklenerd: Alternatively we could remove the guarantee that the caller gets what they asked for and force them to check the type.

vidyaap: Also wanted to throw in there that because of the types, retrieving each type of secret has its own function right now, so we can specify which one of the lists to look through in each.

lock-keeper-tests/src/end_to_end.rs
picklenerd: I don't think there's an issue for that since I don't think retrieving them is part of our API. We'll be able to test this more thoroughly when we do the remote signing issue but this feature is a prerequisite for that issue.

lock-keeper-key-server/src/operations/remote_generate.rs
picklenerd: If it's missing from all of them I think we should just deal with them once we have our basic sessions working later in this phase.

lock-keeper/src/crypto/signing_key.rs
picklenerd: On second though,  I think it would be a better idea to have a temporary type that represents an encrypted version of this. I really don't want to implement any conversions on this type that would allow it to be used improperly.

picklenerd: Ok I updated this with a really simple type that has the same data, is convertible from a `SigningPublicKey`, and implements serialize/deserialize. It has a very long and ugly name so that it stands out.

lock-keeper-key-server/src/operations/remote_generate.rs
marsella: Jordan: are you suggesting the messages with user ids would not have user ids in them any more? Like, you would send a unit struct of `RequestRemoteGenerate` and the server would pull the user id from the authentication?
If not, then this is a separate issue -- you can have an authenticated channel for user A and still send a user id over it that corresponds to user B.

Vidya: yes, any request that starts with sending a user id should make sure that user id matches the authenticated user (with some caveat for the future of passive requests where the requesting client will be a Service Provider making requests on behalf of a user. But I don't think we need to build for that now).

picklenerd: Yeah I think they'd have to have some sort of marker in them to indicate the message type but they wouldn't need to send a user id since this will be available in the session. If the authenticated user_id is the only valid user_id, then having the user send their user_id doesn't do anything.

marsella: Okay, that sounds good to me. Is this plan explicitly documented in an issue yet?

lock-keeper/src/crypto/signing_key.rs
marsella: Thanks, I like this plan!

lock-keeper/src/types/database/secrets.rs
marsella: Okay, cool. That sounds good to me. 

> Alternatively we could remove the guarantee that the caller gets what they asked for and force them to check the type.

This seems more mistake-prone. What you have planned makes sense to me.

lock-keeper-tests/src/end_to_end.rs
marsella: Retrieval of any kind of secret is definitely in the spec. Do you mean that it's not part of this phase?

vidyaap: I don't think retrieval of signing keys is in this phase, but export is, and the way I wrote it is just a wrapper around retrieval (same as arbitrary secrets). I'm not sure about the accuracy of the retrieval I made for signing keys because we didn't talk about including it in the phase, but pending fixes there, it should be enough to add the test you're talking about.

lock-keeper-key-server/src/operations/remote_generate.rs
picklenerd: https://app.zenhub.com/workspaces/bolt-team-60886be9e01f13000e361d92/issues/boltlabs-inc/key-mgmt/289

marsella: Thank you!

PR #277, opened 2022-10-11
add import methods for secret types

### LINKED ISSUE #134: Implement public cryptography for generate-and-store flow

marsella:
Closes boltlabs-inc/key-mgmt-spec#134: adds an import-and-encrypt method for signing keys and arbitrary secrets.

I added a new item to #235 to implement this correctly for signing keys. I also added context to the (previously trivial) signing keys because I wanted to test that the context is set correctly.  This adds a bit of boilerplate code that is actually duplicated in #262.

### DIFF
--- a/lock-keeper/src/crypto.rs
+++ b/lock-keeper/src/crypto.rs
@@ -268,6 +268,41 @@ impl KeyId {
     }
 }
 
+/// The set of types that can be imported into the system, either to the key
+/// server only or as a local import with remote backup.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum Importable {
+    ArbitrarySecret,
+    SigningKey,
+}
+
+/// Raw material for imported key types.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Import {
+    material: Vec<u8>,
+    key_type: Importable,
+}
+
+impl Import {
+    /// Wrap key material as an import.
+    pub fn new(material: &[u8], key_type: Importable) -> Self {
+        Self {
+            material: material.into(),
+            key_type,
+        }
+    }
+
+    /// Retrieve the key type.
+    pub fn key_type(&self) -> Importable {
+        self.key_type
+    }
+
+    /// Extract the key material.
+    pub fn into_material(self) -> Vec<u8> {
+        self.material
+    }
+}
+
 #[cfg(test)]
 mod test {
     use std::collections::HashSet;
--- a/lock-keeper/src/crypto.rs
+++ b/lock-keeper/src/crypto.rs
@@ -268,41 +268,6 @@ impl KeyId {
     }
 }
 
-/// The set of types that can be imported into the system, either to the key
-/// server only or as a local import with remote backup.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
-pub enum Importable {
-    ArbitrarySecret,
-    SigningKey,
-}
-
-/// Raw material for imported key types.
-#[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct Import {
-    material: Vec<u8>,
-    key_type: Importable,
-}
-
-impl Import {
-    /// Wrap key material as an import.
-    pub fn new(material: &[u8], key_type: Importable) -> Self {
-        Self {
-            material: material.into(),
-            key_type,
-        }
-    }
-
-    /// Retrieve the key type.
-    pub fn key_type(&self) -> Importable {
-        self.key_type
-    }
-
-    /// Extract the key material.
-    pub fn into_material(self) -> Vec<u8> {
-        self.material
-    }
-}
-
 #[cfg(test)]
 mod test {
     use std::collections::HashSet;
--- a/lock-keeper/src/crypto/arbitrary_secret.rs
+++ b/lock-keeper/src/crypto/arbitrary_secret.rs
@@ -79,18 +79,25 @@ impl Secret {
             .with_bytes(key_id.clone())
             .with_str("imported key");
 
-        let secret = Secret(generic::Secret::try_from(secret_material.to_vec())?);
+        let secret = Secret(generic::Secret::from_parts(secret_material, &context));
 
         Ok((
             secret.clone(),
             Encrypted::encrypt(rng, &storage_key.0, secret, &context)?,
         ))
     }
+
+    /// Retrieve the context for the secret.
+    #[allow(unused)]
+    fn context(&self) -> &AssociatedData {
+        self.0.context()
+    }
 }
 
 #[cfg(test)]
 mod test {
     use super::*;
+    use rand::Rng;
 
     #[test]
     fn secret_to_vec_u8_conversion_works() -> Result<(), LockKeeperError> {
@@ -128,6 +135,45 @@ mod test {
         // Decrypt the secret
         let decrypted_secret = encrypted_secret.decrypt_secret(storage_key)?;
         assert_eq!(decrypted_secret, secret);
+
+        Ok(())
+    }
+
+    #[test]
+    fn imported_keys_are_labelled() -> Result<(), LockKeeperError> {
+        let mut rng = rand::thread_rng();
+        let storage_key = StorageKey::generate(&mut rng);
+
+        let user_id = UserId::new(&mut rng)?;
+        let key_id = KeyId::generate(&mut rng, &user_id)?;
+
+        // Convenient, inefficient method to check whether one vector contains another.
+        let contains = |container: Vec<u8>, subset: Vec<u8>| -> bool {
+            container.windows(subset.len()).any(|c| c == subset)
+        };
+
+        // Create and encrypt a secret -- not imported.
+        let (secret, _) = Secret::create_and_encrypt(&mut rng, &storage_key, &user_id, &key_id)?;
+
+        assert!(!contains(
+            secret.context().to_owned().into(),
+            "imported".as_bytes().into()
+        ));
+
+        // Use the imported creation function
+        let secret_material: [u8; 32] = rng.gen();
+        let (imported_secret, _) = Secret::import_and_encrypt(
+            &secret_material,
+            &mut rng,
+            &storage_key,
+            &user_id,
+            &key_id,
+        )?;
+
+        assert!(contains(
+            imported_secret.context().to_owned().into(),
+            "imported".as_bytes().into()
+        ));
 
         Ok(())
     }
--- a/lock-keeper/src/crypto/arbitrary_secret.rs
+++ b/lock-keeper/src/crypto/arbitrary_secret.rs
@@ -5,7 +5,7 @@ use std::convert::TryFrom;
 
 use crate::crypto::{
     generic::{self, AssociatedData, CryptoError},
-    Encrypted, Import, Importable, KeyId, StorageKey,
+    Encrypted, KeyId, StorageKey,
 };
 
 /// An arbitrary secret.
@@ -98,28 +98,6 @@ impl Secret {
     }
 }
 
-impl Import {
-    /// Convert an [`Import`] into a [`Secret`] with appropriate context.
-    ///
-    /// This is part of the import to key server only flow and must be called by
-    /// the server.
-    pub fn into_secret(self, user_id: &UserId, key_id: &KeyId) -> Result<Secret, CryptoError> {
-        if self.key_type() != Importable::ArbitrarySecret {
-            Err(CryptoError::ConversionError)
-        } else {
-            let context = AssociatedData::new()
-                .with_bytes(user_id.clone())
-                .with_bytes(key_id.clone())
-                .with_str("imported key");
-
-            Ok(Secret(generic::Secret::from_parts(
-                &self.into_material(),
-                &context,
-            )))
-        }
-    }
-}
-
 #[cfg(test)]
 mod test {
     use super::*;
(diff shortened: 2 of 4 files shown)


indomitableSwan:
> Closes [boltlabs-inc/key-mgmt-spec#134](https://github.com/boltlabs-inc/key-mgmt-spec/issues/134): adds an import-and-encrypt method for signing keys and arbitrary secrets.
> 
1. Does this mean that no changes to the spec are required to close [boltlabs-inc/key-mgmt-spec#134](https://github.com/boltlabs-inc/key-mgmt-spec/issues/134)?
2. Can you clarify what your goal is for this functionality? i.e., are you encrypting client-side? Thinking about what we need in terms of a full set of features, I think we probably have another bug in the spec here. That is, I think it was incorrect to say the import functionality is identical for signing keys as what we already have implemented, because we ultimately want to support:
    1. Importing and encrypting client-side (under a key known only to the client), then storing at server. [@jwantz We do want this, but we should write it up as a stretch goal.]
    2. Importing and securely sending the key to the server, which then encrypts and stores in (potentially untrusted) storage. 
 
As an aside, let's think about this functionality for a moment: 
- Importing an arbitrary secret and securely sending the arbitrary secret to the server, which then encrypts and stores in (potentially untrusted) storage.

I believe this functionality has limited utility. What are your thoughts?

> I added a new item to #235 to implement this correctly for signing keys. I also added context to the (previously trivial) signing keys because I wanted to test that the context is set correctly. This adds a bit of boilerplate code that is actually duplicated in #262.

lock-keeper/src/crypto/arbitrary_secret.rs
```diff
@@ -62,11 +62,42 @@ impl Secret {
             Encrypted::encrypt(rng, &storage_key.0, secret, &context)?,
         ))
     }
+
+    /// Create a `Secret` from an imported key and encrypt it for
+    /// storage at a server.
```
indomitableSwan, line 67: Clarify that this is encrypted under a key known only to the client.

lock-keeper/src/crypto/arbitrary_secret.rs
```diff
@@ -107,4 +138,43 @@ mod test {
 
         Ok(())
     }
+
+    #[test]
```
indomitableSwan, line 142: This test looks OK, but why don't we have unit tests for the new `import_and_encrypt` method?

lock-keeper/src/crypto/generic.rs
```diff
@@ -252,6 +252,19 @@ impl Secret {
             context,
         }
     }
+
+    /// Create a new secret from its constituent parts.
+    /// This is unchecked; use with care.
+    pub(super) fn from_parts(secret_material: &[u8], context: &AssociatedData) -> Self {
```
indomitableSwan, line 258: What validation should we add? We should create an issue for this if we aren't doing this here.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -51,7 +59,38 @@ impl SigningKeyPair {
...
-        SigningKeyPair
+        let context = AssociatedData::new()
+            .with_bytes(user_id.clone())
+            .with_bytes(key_id.clone())
+            .with_str("server-generated");
+        SigningKeyPair { context }
+    }
+
+    /// Create a `SigningKeyPair` from an imported key and encrypt it for
+    /// storage at a server.
```
indomitableSwan, line 70: Clarify that this encryption is under a key known only to the client.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -194,4 +240,52 @@ mod test {
                 |sig| trivial_signature == sig && sig.verify(public_key, &generic_message).is_ok()
             ));
     }
+
+    #[test]
```
indomitableSwan, line 244: Same question here about test coverage as for arbitrary secrets. Don't we want unit tests for the new `import_and_encrypt` method?

indomitableSwan (changes requested):
Aside from test coverage and some minor documentation nits, what is implemented here looks OK, but incomplete.

That is, we need additional functionality that imports the key without encrypting it.

Note: I didn't check the build or docs yet; my dev environment needs some more setup atm.

lock-keeper/src/crypto/generic.rs
marsella: I couldn't think of any validation to do, which is why it's unchecked. We allow any-length arbitrary secrets and I'm not sure there are specific restrictions on the associated data that we would want to enforce at this level (e.g. the requirement that an imported key is labeled imported is enforced at function that calls this one, not within this).

marsella:
Thanks for the comments.

> 1. Does this mean that no changes to the spec are required to close https://github.com/boltlabs-inc/key-mgmt-spec/issues/134?

As far as I can tell, yes. Both import methods specify that imported keys are imported, and in the case where the server stores an imported key from the client, it also marks it as imported (e.g. doesn't trust the client to set context correctly).

> 2. Can you clarify what your goal is for this functionality? i.e., are you encrypting client-side?

Yeah, I tried to do this in a hurry and didn't do it carefully enough. The initial commit only included methods for client-side import + encryption, which can then be sent to the server and stored. I added a second commit with a new type (`Import`) to send to the server with (unencrypted) imported key material and a method to convert it server-side to the appropriate type (`Secret`/`SigningKeyPair`) with correct context. I also improved documentation to clarify the difference between these two import flows.

I won't take out the client-side stuff since we want to use it eventually; I trust @vidyaap to ignore it while implementing the "import to key server only" flow.

> Importing an arbitrary secret and securely sending the arbitrary secret to the server, which then encrypts (under a key known only to the server) and stores in (potentially untrusted) storage.
>
> I believe this functionality has limited utility. What are your thoughts?

I agree. With an arbitrary secret, there isn't a good reason to allow the server to actually see it, if they can't do anything useful with it. The only useful thing you can do with an arbitrary secret is to retrieve it, which doesn't require the server to know the secret. In future cases where we're dealing with shared custody (e.g. you want to import an arbitrary secret and then get permission to retrieve it in the future), it's still of limited value because the other custody partners would have to trust the client to have gotten rid of their original value. And actually, the server still wouldn't need to _know_ the secret.

I can take that bit of crypto out, or we can clarify in the spec that the remote-only storage is for secret types that can be used for operations beyond just retrieval.

> Aside from test coverage and some minor documentation nits...

I added unit tests for the two imported key generation methods: the `import_and_encrypt` that you saw in the first review, and the `into_secret`/`into_signing_key` methods I added now for the server-only import flow. There are some TODOs in the signing key tests about adding more checks once that type is fleshed out properly in #235; I wrote the tests so that they'll fail once we add key material, so I won't miss fixing them.

vidyaap:
> I won't take out the client-side stuff since we want to use it eventually; I trust @vidyaap to ignore it while implementing the "import to key server only" flow.

Does this mean I should only be doing the "import to key server only" flow for now, or just that I should know that the two types of import use different functionality?

lock-keeper/src/crypto/signing_key.rs
marsella: Yeah, we definitely do.

marsella:
> Does this mean I should only be doing the "import to key server only" flow for now, or just that I should know that the two types of import use different functionality?

Based on @indomitableSwan's first comment, I think the former. This phase prioritizes "import to key server only," and has as a stretch goal "local import with remote backup". I guess you can evaluate whether they are similar enough to do at once; if not, just do the key server version.

indomitableSwan:
> > I won't take out the client-side stuff since we want to use it eventually; I trust @vidyaap to ignore it while implementing the "import to key server only" flow.
> 
> Does this mean I should only be doing the "import to key server only" flow for now, or just that I should know that the two types of import use different functionality?

Yes, the "import to key server only" flow is what you should be implementing now. We should fix the import epic/ticket description to split up the [two import flows](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/system-functionalities.md#import-a-secret) into different tickets. We will eventually want both.

indomitableSwan (changes requested):
- What are you thoughts on leaving crypto in that we don't have a good reason to enable? Particularly since in this case, "importing to key server only" involves sending sensitive information over the wire, I'm leaning towards removal (and opening up an appropriate bug issue in the spec).
- A few more nits to make documentation clearer.
- Thanks for adding tests!

lock-keeper/src/crypto.rs
```diff
@@ -268,6 +268,41 @@ impl KeyId {
     }
 }
 
+/// The set of types that can be imported into the system, either to the key
+/// server only or as a local import with remote backup.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum Importable {
+    ArbitrarySecret,
+    SigningKey,
+}
```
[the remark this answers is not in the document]
marsella: Right now, this enum is actually only used in the "import to key server only" flow. Since I'm going to remove that flow entirely for arbitrary secrets, I can also remove this entire abstraction.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -115,25 +121,68 @@ impl SigningKeyPair {
     }
 }
 
+impl Import {
+    /// Convert an [`Import`] into a [`SigningKeyPair`] with appropriate
+    /// context.
+    ///
+    /// This is part of the import to key server only flow and must be called by
+    /// the server.
```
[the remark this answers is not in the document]
marsella: Removing this function instead.

lock-keeper/src/crypto/arbitrary_secret.rs
```diff
@@ -62,11 +62,46 @@ impl Secret {
...
         ))
     }
+
+    /// Create a `Secret` from an imported key and encrypt it for
+    /// storage at a server, under a key known only to the client.
+    ///
+    /// This is part of the local import with remote backup flow and must be run
+    /// by the client.
+    pub fn import_and_encrypt(
+        secret_material: &[u8],
```
marsella, line 72: @vidyaap I said I would change this to a `Vec` but got a clippy error when I tried to pass a reference to a `Vec`. You can feel free to still use vectors everywhere and just pass by reference to this function.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -51,22 +60,66 @@ impl SigningKeyPair {
...
+    /// storage at a server, under a key known only to the client.
+    ///
+    /// This is part of the local import with remote backup flow and must be run
+    /// by the client. In this flow, the key server will only recieve an
+    /// [`Encrypted<SigningKeyPair>`], not the cleartext.
     ///
-    /// This must be run by the client. It takes the following steps:
+    /// This function takes the following steps:
+    /// 1. Format the `key_material` as a signing key
+    /// 2. Encrypt it under the [`StorageKey`], using an AEAD scheme
```
marsella, line 79: @indomitableSwan edited this a little bit to more obviously echo the documentation for `create_and_encrypt` since they are quite similar. It should still hit all the points you raised re: defining the flow it's for and mentioning what information gets sent to the server.

lock-keeper/src/crypto/arbitrary_secret.rs
```diff
@@ -88,12 +89,37 @@ impl Secret {
...
+    #[cfg(test)]
     fn context(&self) -> &AssociatedData {
         self.0.context()
     }
 }
 
+impl Import {
+    /// Convert an [`Import`] into a [`Secret`] with appropriate context.
+    ///
+    /// This is part of the import to key server only flow and must be called by
```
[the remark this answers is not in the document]
marsella: Fixed this.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -67,9 +69,10 @@ impl SigningKeyPair {
     }
 
     /// Create a `SigningKeyPair` from an imported key and encrypt it for
-    /// storage at a server.
+    /// storage at a server, under a key known only to the client.
     ///
-    /// This is part of the import flow and must be run by the client.
+    /// This is part of the local import with remote backup flow and must be run
+    /// by the client.
```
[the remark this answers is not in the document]
marsella: Good call, added.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -72,7 +71,12 @@ impl SigningKeyPair {
     /// storage at a server, under a key known only to the client.
     ///
     /// This is part of the local import with remote backup flow and must be run
-    /// by the client.
+    /// by the client. In this flow, the key server will only recieve an
```
indomitableSwan: typo: "recieve"

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -96,18 +100,22 @@ impl SigningKeyPair {
         ))
     }
 
-    /// Create and encrypt a new signing key. This is part of the local signing
-    /// key generation flow.
+    /// Create and encrypt a new signing key for storage at
+    /// a server, under a key known only to the client.
+    ///
+    /// This is part of the local signing key generation flow and must be run by
+    /// the client. In this flow the key server will only recieve an
```
indomitableSwan: typo: "recieve"

marsella:
I also made boltlabs-inc/key-mgmt-spec#139 to address updating the spec to remove the "import to key server only" flow for arbitrary secrets.

indomitableSwan (approved):
LGTM.

PR #278, opened 2022-10-11
Run cargo audit when dependencies change #275

### LINKED ISSUE #275: Add Rust crate audits to CI
There are tools to audit Rust crates for vulnerabilities. `cargo audit` and `cargo crev` are the two that I'm familiar with. We already have Dependabot enabled which will likely cover the same bases, but it would be useful to add one of these tools to CI just to be safe.

https://crates.io/crates/cargo-audit
https://crates.io/crates/cargo-crev

I've used `cargo audit` in the past so I'd like to do a little research on `cargo crev` to see if it's worth using instead or maybe even using both.

One downside of putting this in CI is that there might be a vulnerability in a downstream crate of one of our dependencies. If the crate we depend on doesn't update their downstream vulnerability, we have no way of fixing CI. One way to solve this would be for CI to pass but to highlight that errors occurred. This isn't ideal because we don't want to end up in a situation where we always have a bunch of warnings that we ignore, so I think we should deal with this in the future if we end up in this situation.

1. Investigate `cargo audit` and `cargo crev` and decide whether to use one or both of them.
2. Add new tasks to `Makefile.toml` for whichever tool we choose.
3. Add the new tasks to the `[tasks.ci]` task so that they run in CI

picklenerd:
We had some discussions on Slack and it turns out that we already have `cargo audit` running daily. That action also creates issues for any vulnerabilities it finds.

This PR adds a new action that runs any time dependencies change. This action will fail if it discovers any vulnerabilities but it will allow the CI workflow to continue so that we're not blocked.

Closes #275

### DIFF
--- a/.github/workflows/audit-scheduled.yml
+++ b/.github/workflows/audit-scheduled.yml
@@ -1,0 +1,14 @@
+# Runs a daily security audit of dependencies using `cargo audit`, an issue gets created for vulnerabilities detected
+name: Scheduled security audit
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *'
+jobs:
+  audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/audit-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,12 +1,14 @@
-# Runs a daily security audit of dependencies using `cargo audit`, an issue gets created for vulnerabilities detected
+# Runs `cargo audit` when dependencies change
 name: Security audit
 on:
-  workflow_dispatch:
-  schedule:
-    - cron: '0 0 * * *'
+  push:
+    paths: 
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
 jobs:
-  audit:
+  security_audit:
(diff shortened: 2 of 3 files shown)


vidyaap (commented):
This looks good to me! Could these actions can't be combined into one YAML file, or would the `continue-on-error` not work in that case?

picklenerd:
> This makes sense to me! Could these actions can't be combined into one YAML file, or would the `continue-on-error` not work in that case?

I think the different `on:` triggers require different workflows but I can double check that. 

> Also, it seems that this check blocks merging now, but should it do that if it's also creating the issue to look at?

Yeah it shouldn't be blocked merge. I think maybe `continue-on-error` only allows the single workflow file to continue running. I'll see if I can figure out a way to turn these errors into warnings.

vidyaap:
Sounds good! And I think we were able to put multiple `on:` triggers on the [zeekoe e2e tests](https://github.com/boltlabs-inc/zeekoe/blob/main/.github/workflows/e2e-test.yml) so that might work here.

picklenerd:
> Sounds good! And I think we were able to put multiple `on:` triggers on the [zeekoe e2e tests](https://github.com/boltlabs-inc/zeekoe/blob/main/.github/workflows/e2e-test.yml) so that might work here.

I think those triggers work for the whole workflow. We have different triggers for different workflows.

picklenerd:
I don't see a way to make this a warning. I'm going to close this. I think the daily checks are sufficient.

PR #282, opened 2022-10-12
Added name filtering to e2e tests

picklenerd:
Adds the ability to specify filters to the e2e tests. If filters are provided, tests will only run if their name contains the text of the filter argument. See README for an example.

This PR also includes a few housekeeping items:
- `cargo make restart` task added to quickly rebuild and run server after making code changes.
- Updated docker compose to run key server with Docker's `-it` flags for easier  attach/detach behavior.
- Configs are now loaded synchronously. 

This code was pulled from https://github.com/boltlabs-inc/key-mgmt/pull/273 to make that PR smaller.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -15,6 +15,10 @@ alias = "stop-server"
 [tasks.teardown]
 alias = "teardown-server"
 
+# Stops the server, recompiles it, and restarts it
+[tasks.restart]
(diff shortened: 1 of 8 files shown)


gatoWololo (approved):
Looks good! I couldn't find anything to comment on or complain about :)

PR #283, opened 2022-10-12
258 import signing keys

### LINKED ISSUE #258: Implement `import_key` flows for LockKeeperClient
**DRAFT:** This ticket is still being drafted

**Background:** 
   - Refer to the spec ... TBD Link Here

**Requirements:**

   - Add a new public function to the local client (formerly called the DamsClient) API that allows the caller to import a key.
   - API function name: TBD (example: `import_key`)
   - Inputs: TBD
   - Outputs: KeyId, Error
   - Actions: The client should encrypt the key with the storage_key before sending to the server
1. Client Action: Utilize the pre-existing action to store an encrypted key on the Key Server.
   - generate a key locally and hold onto it
   - import the previously generated key
   - request to export the same key
   - compare exported key to the originally generated key

**Design Thoughts:** TBD: Jay is working on a diagram, which will be linked to Confluence here.

vidyaap:
Closes #258 

Making this to reconcile differences with remote generation branch. I like the new types module structure there, so I'll port that over. 

Things that will probably be helpful for remote generation:
- the Import type - this can maybe be a more general wrapper around bytes that can be turned into a signing key?
- the different stored secret types
- the struct that holds a user's different secrets (@picklenerd I went with a struct after all because there were more than just the 2 types of keys)

Update:
The export branch is branched off this one, so this should be merged first. There is a question in the export PR (#287) about import behavior.

### DIFF
--- a/lock-keeper-client/src/api/import_signing_key.rs
+++ b/lock-keeper-client/src/api/import_signing_key.rs
@@ -11,9 +11,7 @@ impl LockKeeperClient {
         channel: &mut ClientChannel,
         key_material: Vec<u8>,
     ) -> Result<KeyId, LockKeeperClientError> {
-        let key_material = Import {
-            material: key_material,
-        };
+        let key_material = Import { key_material };
         // Send UserId and key material to server
         let request = client::Request {
             user_id: self.user_id().clone(),
--- a/lock-keeper-client/src/api/import_signing_key.rs
+++ b/lock-keeper-client/src/api/import_signing_key.rs
@@ -1,0 +1,29 @@
+use crate::{LockKeeperClient, LockKeeperClientError};
+use lock_keeper::{
+    crypto::{Import, KeyId},
+    infrastructure::channel::ClientChannel,
+    types::operations::import_signing_key::{client, server},
+};
+
+impl LockKeeperClient {
+    pub(crate) async fn handle_import_signing_key(
+        &self,
+        channel: &mut ClientChannel,
+        key_material: Vec<u8>,
+    ) -> Result<KeyId, LockKeeperClientError> {
+        let key_material = Import {
+            material: key_material,
+        };
+        // Send UserId and key material to server
+        let request = client::Request {
+            user_id: self.user_id().clone(),
+            key_material,
+        };
+        channel.send(request).await?;
+
+        // Get KeyId for imported key from server
+        let server_response: server::Response = channel.receive().await?;
+
+        Ok(server_response.key_id)
+    }
+}
--- a/lock-keeper-key-server/src/operations/import_signing_key.rs
+++ b/lock-keeper-key-server/src/operations/import_signing_key.rs
@@ -1,0 +1,49 @@
+use crate::{
+    server::{Context, Operation},
+    LockKeeperServerError,
+};
+
+use async_trait::async_trait;
+use lock_keeper::{
+    crypto::KeyId,
+    infrastructure::channel::ServerChannel,
+    types::operations::import_signing_key::{client, server},
+};
+
+#[derive(Debug)]
+pub struct ImportSigningKey;
+
+#[async_trait]
+impl Operation for ImportSigningKey {
+    async fn operation(
+        self,
+        channel: &mut ServerChannel,
+        context: &mut Context,
+    ) -> Result<(), LockKeeperServerError> {
+        // Receive UserId and key material from client
+        let request: client::Request = channel.receive().await?;
+
+        // Generate new KeyId
+        let key_id = {
+            let mut rng = context.rng.lock().await;
+            KeyId::generate(&mut *rng, &request.user_id)?
+        };
+        context.key_id = Some(key_id.clone());
+
+        // Make signing key out of bytes
+        let signing_key = request
+            .key_material
+            .into_signing_key(&request.user_id, &key_id)?;
+
+        // Check validity of ciphertext and store in DB
+        context
+            .db
+            .add_server_imported_signing_key(&request.user_id, signing_key, key_id.clone())
+            .await?;
+
+        // Serialize KeyId and send to client
+        let reply = server::Response { key_id };
+        channel.send(reply).await?;
+        Ok(())
+    }
+}
--- a/lock-keeper-tests/src/end_to_end.rs
+++ b/lock-keeper-tests/src/end_to_end.rs
@@ -24,6 +24,7 @@ const USER: &str = "user";
 const PASSWORD: &str = "password";
 const GENERATED_ID: &str = "generated_key_id";
 const GENERATED_KEY: &str = "generated_key";
+const IMPORTED_ID: &str = "imported_key_id";
 
 pub async fn end_to_end_tests(config: &Config) {
     // Run every test, printing out details if it fails
@@ -266,6 +267,23 @@ async fn tests() -> Vec<Test> {
                 ),
             ],
         ),
+        Test::new(
+            "Import a signing key",
+            vec![
+                (
+                    Register,
+                    Outcome {
+                        expected_error: None,
+                    },
+                ),
+                (
+                    ImportSigningKey,
+                    Outcome {
+                        expected_error: None,
+                    },
+                ),
+            ],
+        ),
     ]
 }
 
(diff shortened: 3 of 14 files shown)


lock-keeper/src/types/database/secrets.rs
```diff
@@ -21,3 +26,39 @@ impl StoredSecret {
...
+            retrieved: false,
+        }
+    }
+}
+
+/// Wrapper around an [`SigningKeyPair`] and its [`KeyId`]
+#[derive(Debug, Deserialize, Serialize)]
+pub struct StoredSigningKeyPair {
+    pub signing_key: SigningKeyPair,
+    pub key_id: KeyId,
```
vidyaap, line 52: @marsella What do you think of having the two types for stored signing keys here? I was wondering if it would be easier to just have the one type and update it later when we decide on how to encrypt server-generated/imported keys. But @picklenerd pointed out that having the two different types makes it clear that we are not encrypting them right now. Thoughts?

picklenerd (approved):
LGTM pending any feedback @marsella on the two types question.

lock-keeper-client/src/api/import_signing_key.rs
```diff
@@ -0,0 +1,29 @@
...
+};
+
+impl LockKeeperClient {
+    pub(crate) async fn handle_import_signing_key(
+        &self,
+        channel: &mut ClientChannel,
+        key_material: Vec<u8>,
+    ) -> Result<KeyId, LockKeeperClientError> {
+        let key_material = Import {
+            material: key_material,
```
marsella, line 15: extremely minor nit: if you want, now that `Import` is only used for signing keys, you can name this parameter `key_material`. I made a `From<&[u8]>` for this step -- if it's not useful here, you can delete it.

lock-keeper-key-server/src/operations/import_signing_key.rs
```diff
@@ -0,0 +1,49 @@
...
+        channel: &mut ServerChannel,
+        context: &mut Context,
+    ) -> Result<(), LockKeeperServerError> {
+        // Receive UserId and key material from client
+        let request: client::Request = channel.receive().await?;
+
+        // Generate new KeyId
+        let key_id = {
+            let mut rng = context.rng.lock().await;
+            KeyId::generate(&mut *rng, &request.user_id)?
```
marsella, line 29: note: Same comment here as on #273 about making sure this is the correct user id. I think that is going to end up with an issue being written about how we'll deal with it, so no action is needed here, but keep an eye on that conversation.

https://github.com/boltlabs-inc/key-mgmt/pull/273#discussion_r995050691

lock-keeper-tests/src/end_to_end.rs
```diff
@@ -400,6 +418,19 @@ impl Test {
...
+                    let lock_keeper_client = LockKeeperClient::authenticated_client(
+                        &self.account_name,
+                        &self.password,
+                        config,
+                    )
+                    .await?;
+                    let random_bytes = rand::thread_rng().gen::<[u8; 32]>().to_vec();
+                    let key_id = lock_keeper_client.import_signing_key(random_bytes).await?;
+                    // Store generated key ID and local storage object to state
+                    self.state.set(IMPORTED_ID, key_id)
```
marsella, line 432: nit: What's "local storage object"? Is that a copy/paste remnant?

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -11,7 +11,7 @@ use super::{generic::AssociatedData, CryptoError, Encrypted, KeyId, StorageKey};
 ///
 /// This can be generated locally by the client or remotely by the server.
 #[allow(unused)]
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
```
marsella, line 14: question: this is ported from #273, right? It got removed from that PR; will you make sure it disappears before you merge this one, too?

lock-keeper/src/types/database/secrets.rs
marsella: I definitely don't want to "fake encrypt" the imported keys because I think it will be harder to tell that it's an incorrect intermediate step later.

I would be open to the possibility of something like, put an enum into the `StoredEncryptedSigningKeyPair` that lets it hold either an encrypted or an unencrypted `SKP`. I think in either this case or the current status (totally separate types + lists), there will be a small handful of changes to export and retrieve when we switch to correctly-encrypted server-generated/imported keys.

lock-keeper-key-server/src/operations/import_signing_key.rs
marsella: Update - please add import to the checklist in #289.

lock-keeper-key-server/src/operations/import_signing_key.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use async_trait::async_trait;
+use lock_keeper::{
+    crypto::KeyId,
+    infrastructure::channel::ServerChannel,
+    types::operations::import_signing_key::{client, server},
+};
+
+#[derive(Debug)]
+pub struct ImportSigningKey;
```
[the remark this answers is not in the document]
vidyaap: Yeah I wasn't sure how to handle these because I don't want the names to get too long. I think they will be different calls since the messages will be different. "RemoteImportSigningKey" didn't sound correct either, but if that works, I can switch it to that.

lock-keeper/src/crypto/signing_key.rs
vidyaap: Yep, if that gets merged first, I can rebase and use the placeholder object instead of this one. And then this won't be serializable anymore.

lock-keeper/src/types/database/secrets.rs
vidyaap: I'm not sure I like the possibility of having an unencrypted SKP in a struct that says "encrypted". I'm currently using the `StoredSigningKeyPair` in the DB, and I don't actually use the encrypted one anywhere. I was wondering if it was worth keeping the encrypted one around so we can switch to that later or if I should just delete it and update the unencrypted one to be encrypted when it actually is? 
Also, does the placeholder type from #273 relate to this at all since it's a placeholder for an encrypted SKP?

lock-keeper-key-server/src/operations/import_signing_key.rs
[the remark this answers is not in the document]
marsella: Yeah, that's reasonable. I think this is fine for now. Maybe the second one can be, like, `ImportEncryptedSigningKey`.

lock-keeper/src/types/database/secrets.rs
picklenerd: `PlaceholderEncryptedSigningKeyPair` is merged now, along with a `StoredSigningKeyPair` that holds it.

lock-keeper-key-server/src/operations/import_signing_key.rs
vidyaap: Updated!

lock-keeper/src/types/database/secrets.rs
marsella: > I'm not sure I like the possibility of having an unencrypted SKP in a struct that says "encrypted". 

Yes, very good point, that was a bad idea.

Is the encrypted version supposed to be for the scenario where the client encrypts a key pair and sends it to the server, or to replace the unencrypted key pair when the server gets a key pair in the clear and then encrypts it for storage?

If the latter, you should use the `PlaceholderESKP` that @picklenerd made. If the former, I might tend towards just deleting it now and rewriting when we add that flow (was it hard to make?). Or, if you choose to keep it around, make sure you document it explicitly to describe exactly which flow it belongs to (e.g. with something like, "encrypted under a key held by the client").

vidyaap: Ah right, I forgot about the two types of encryption. I decided to leave both types in so that "StoredX" means "X" is what the client is working with/sending (docstrings are updated too). So `StoredSigningKeyPair` holds the Placeholder and means the client sent the server a SKP and the server encrypted it, and `StoredEncryptedSigningKeyPair` holds an `Encrypted<SKP>` and means that the client sent the server an encrypted SKP. Does that work?

lock-keeper-client/src/api/import_signing_key.rs
vidyaap: Did the rename, but I left the From impl in because it's used elsewhere.

lock-keeper/src/types/database/secrets.rs
marsella: Yep, that sounds good.

PR #286, opened 2022-10-13
Refactor database types into submodules

picklenerd:
This PR creates a new types module for database types and moves the secret functionality out of the `user` module into its own `secrets` module.

Moved from: https://github.com/boltlabs-inc/key-mgmt/pull/283


vidyaap (approved):
Looks good to me! Thanks for rearranging these!

PR #287, opened 2022-10-13
281 export signing keys

### LINKED ISSUE #281: Implement export for signing keys
The description of this ticket is the same as #148, but the previous ticket was implemented only for arbitrary secrets. This ticket covers the same functionality for signing keys.

vidyaap:
Closes #281 

This adds the export functionality for signing keys. The previous PR (#263) only implemented it for arbitrary secrets.

Questions for @indomitableSwan: 
- This export is modeled after the arbitrary secrets export, which in turn is based on arbitrary secrets retrieve. Since we didn't talk about implementing retrieve for signing keys yet, I wasn't sure if we would want to wrap the retrieved signing key in a `LocalStorage` like we do for arbitrary keys, so I have them being returned plain for now. Is that okay?
- The import spec doesn't say to return the imported signing key to the user, so currently I don't have a way to write an export test that checks that the key material is the same. Should I be returning the key in import or is there a different way to check this?

### DIFF
--- a/lkic/src/storage.rs
+++ b/lkic/src/storage.rs
@@ -8,7 +8,7 @@ use std::{
 };
 
 use anyhow::anyhow;
-use lock_keeper::crypto::{KeyId, PlaceholderEncryptedSigningKeyPair};
+use lock_keeper::crypto::{Export, KeyId};
 use lock_keeper_client::api::{LocalStorage, RetrieveResult};
 use serde::{Deserialize, Serialize};
 
@@ -260,10 +260,10 @@ fn local_storage_bytes(local_storage: &LocalStorage) -> anyhow::Result<Vec<u8>> 
     Ok(bytes)
 }
 
-fn signing_key_bytes(signing_key: &PlaceholderEncryptedSigningKeyPair) -> anyhow::Result<Vec<u8>> {
+fn signing_key_bytes(signing_key: &Export) -> anyhow::Result<Vec<u8>> {
     let json = serde_json::to_value(signing_key)?;
     let key = json
-        .pointer("/context")
+        .pointer("/key_material")
         .ok_or_else(|| anyhow!("Error converting SigningKey to bytes"))?
         .clone();
     let bytes: Vec<u8> = serde_json::from_value(key)?;
--- a/lkic/src/storage.rs
+++ b/lkic/src/storage.rs
@@ -8,7 +8,7 @@ use std::{
 };
 
 use anyhow::anyhow;
-use lock_keeper::crypto::KeyId;
+use lock_keeper::crypto::{KeyId, PlaceholderEncryptedSigningKeyPair};
 use lock_keeper_client::api::{LocalStorage, RetrieveResult};
 use serde::{Deserialize, Serialize};
 
@@ -214,6 +214,14 @@ impl Display for Entry {
                 })?;
                 format!("Arbitrary Key - {}", hex::encode(&bytes))
             }
+            RetrieveResult::SigningKey(signing_key) => {
+                let bytes = signing_key_bytes(signing_key).map_err(|e| {
+                    // Avoid error inception
+                    let _ = writeln!(f, "{e}");
+                    std::fmt::Error
+                })?;
+                format!("Signing Key Pair - {}", hex::encode(&bytes))
+            }
         };
 
         writeln!(f, "key_id: {}", key_id_hex)?;
@@ -247,6 +255,16 @@ fn local_storage_bytes(local_storage: &LocalStorage) -> anyhow::Result<Vec<u8>> 
     let key = json
         .pointer("/secret/material")
         .ok_or_else(|| anyhow!("Error converting LocalStorage to bytes"))?
+        .clone();
+    let bytes: Vec<u8> = serde_json::from_value(key)?;
+    Ok(bytes)
+}
+
+fn signing_key_bytes(signing_key: &PlaceholderEncryptedSigningKeyPair) -> anyhow::Result<Vec<u8>> {
+    let json = serde_json::to_value(signing_key)?;
+    let key = json
+        .pointer("/context")
+        .ok_or_else(|| anyhow!("Error converting SigningKey to bytes"))?
         .clone();
     let bytes: Vec<u8> = serde_json::from_value(key)?;
     Ok(bytes)
--- a/lock-keeper-client/src/api.rs
+++ b/lock-keeper-client/src/api.rs
@@ -147,13 +147,13 @@ impl LockKeeperClient {
         // Create channel: this will internally be a `retrieve` channel
         let mut client_channel = Self::create_channel(
             &mut self.tonic_client(),
-            ClientAction::Export,
+            ClientAction::ExportSigningKey,
             self.account_name(),
         )
         .await?;
         // Get local-only secret
         let secret = self
-            .handle_retrieve(&mut client_channel, key_id, RetrieveContext::LocalOnly)
+            .handle_retrieve_signing_key(&mut client_channel, key_id, RetrieveContext::LocalOnly)
             .await?;
         // Return secret as bytes
         match secret {
--- a/lock-keeper-client/src/api.rs
+++ b/lock-keeper-client/src/api.rs
@@ -17,7 +17,7 @@ mod retrieve_audit_events;
 use crate::{client::Password, LockKeeperClient, LockKeeperClientError};
 use lock_keeper::{
     config::client::Config,
-    crypto::{KeyId, Secret},
+    crypto::{KeyId, PlaceholderEncryptedSigningKeyPair, Secret},
     types::{
         audit_event::{AuditEvent, AuditEventOptions, EventType},
         database::user::AccountName,
@@ -34,6 +34,7 @@ use tracing::error;
 pub enum RetrieveResult {
     None,
     ArbitraryKey(LocalStorage),
+    SigningKey(PlaceholderEncryptedSigningKeyPair),
 }
 
 #[derive(Debug, Deserialize, Serialize)]
@@ -132,6 +133,33 @@ impl LockKeeperClient {
         match secret {
             RetrieveResult::None => Err(LockKeeperClientError::ExportFailed),
             RetrieveResult::ArbitraryKey(local_storage) => Ok(local_storage.secret.into()),
+            RetrieveResult::SigningKey(_) => Err(LockKeeperClientError::InvalidKeyRetrieved),
+        }
+    }
+
+    /// Export signing key pair material from the key servers.
+    ///
+    /// Output: If successful, returns the requested key material in byte form.
+    pub async fn export_signing_key(
+        &self,
+        key_id: &KeyId,
+    ) -> Result<Vec<u8>, LockKeeperClientError> {
+        // Create channel: this will internally be a `retrieve` channel
+        let mut client_channel = Self::create_channel(
+            &mut self.tonic_client(),
+            ClientAction::Export,
+            self.account_name(),
+        )
+        .await?;
+        // Get local-only secret
+        let secret = self
+            .handle_retrieve(&mut client_channel, key_id, RetrieveContext::LocalOnly)
+            .await?;
+        // Return secret as bytes
+        match secret {
+            RetrieveResult::None => Err(LockKeeperClientError::ExportFailed),
+            RetrieveResult::ArbitraryKey(local_storage) => Ok(local_storage.secret.into()),
+            RetrieveResult::SigningKey(_) => Err(LockKeeperClientError::InvalidKeyRetrieved),
         }
     }
 
@@ -217,6 +245,22 @@ impl LockKeeperClient {
         )
         .await?;
         self.handle_retrieve_audit_events(&mut client_channel, event_type, options)
+            .await
+    }
+
+    /// Retrieve a signing key pair from the key server by [`KeyId`]
+    pub async fn retrieve_signing_key(
+        &self,
+        key_id: &KeyId,
+        context: RetrieveContext,
+    ) -> Result<RetrieveResult, LockKeeperClientError> {
+        let mut client_channel = Self::create_channel(
+            &mut self.tonic_client(),
+            ClientAction::RetrieveSigningKey,
+            self.account_name(),
+        )
+        .await?;
+        self.handle_retrieve_signing_key(&mut client_channel, key_id, context)
             .await
     }
 }
--- a/lock-keeper-client/src/api.rs
+++ b/lock-keeper-client/src/api.rs
@@ -17,7 +17,7 @@ mod retrieve_audit_events;
 use crate::{client::Password, LockKeeperClient, LockKeeperClientError};
 use lock_keeper::{
     config::client::Config,
-    crypto::{KeyId, PlaceholderEncryptedSigningKeyPair, Secret},
+    crypto::{KeyId, PlaceholderEncryptedSigningKeyPair, Secret, SigningKeyPair},
     types::{
         audit_event::{AuditEvent, AuditEventOptions, EventType},
         database::user::AccountName,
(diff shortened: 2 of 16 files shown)


lock-keeper-key-server/src/database/secrets.rs
```diff
@@ -3,50 +3,99 @@
...
+        user_id: &UserId,
+        secret: SigningKeyPair,
+        key_id: KeyId,
+    ) -> Result<(), LockKeeperServerError> {
+        let collection = self.inner.collection::<User>(constants::USERS);
+        let stored_secret = StoredSigningKeyPair::new(secret, key_id);
+        let stored_secret_bson = mongodb::bson::to_bson(&stored_secret)?;
+        let filter = doc! { USER_ID: user_id };
+        let update =
+            doc! { "$push":  { "secrets.server_created_signing_keys": stored_secret_bson } };
```
vidyaap, line 50: @picklenerd Is there a better way to handle these query strings like `secrets.server_created_signing_keys`? We had the constants before, but then we'd have to format! them into the correct query anyway; is that more appropriate?

picklenerd: I think this is fine for now. It's a little easier to read than the formatting and I don't think we're any more likely to make mistakes this way.

I wouldn't mind adding some kind of query generation abstraction somewhere so stuff like this is easier. This entire update process is already [problematic](https://app.zenhub.com/workspaces/bolt-team-60886be9e01f13000e361d92/issues/boltlabs-inc/key-mgmt/284).

I don't want to spend too much time on this now because I think we need to have the database talk soon. If we move away from mongo, we'd have to get rid of whatever we come up with.

marsella:
If this needs to be merged after #283, can you set the destination branch to that one instead of `develop`? Makes it easier to read the diff, and it'll automatically switch to `develop` once #283 is merged.

lock-keeper-key-server/src/database/secrets.rs
```diff
@@ -77,11 +96,41 @@ impl Database {
         let stored_secret = StoredSigningKeyPair::new(secret, key_id);
         let stored_secret_bson = mongodb::bson::to_bson(&stored_secret)?;
         let filter = doc! { USER_ID: user_id };
-        let update = doc! { "$push": { REMOTE_SECRETS: stored_secret_bson } };
+        let update =
+            doc! { "$push": { "secrets.server_created_signing_keys": stored_secret_bson } };
```
vidyaap, line 100: @picklenerd This used to push to "remote_secrets" but that creates an extra field called "remote_secrets" on the top level of the user struct. I wasn't sure if you wanted that since we also have the "server_created_signing_keys" list but if you did I can remove the commit that changes it.

lock-keeper-client/src/api/retrieve.rs
```diff
@@ -45,4 +45,34 @@ impl LockKeeperClient {
             }
         }
     }
+
+    pub(crate) async fn handle_retrieve_signing_key(
```
marsella, line 49: suggestion: consider leaving some documentation here and on `handle_retrieve` to be explicit about which cases each one handles.

lock-keeper-key-server/src/operations/retrieve_signing_key.rs
```diff
@@ -0,0 +1,37 @@
...
+        context: &mut Context,
+    ) -> Result<(), LockKeeperServerError> {
+        // Receive UserId from client
+        let request: client::RequestSigningKey = channel.receive().await?;
+        context.key_id = Some(request.key_id.clone());
+
+        // Find secret based on key_id
+        let stored_secret = context
+            .db
+            .get_user_signing_key(&request.user_id, &request.key_id)
```
marsella, line 29: todo: Add this to the checklist in #289.

lock-keeper-key-server/src/operations/retrieve_signing_key.rs
```diff
@@ -0,0 +1,37 @@
...
+            .await?;
+
+        // Serialize KeyId and send to client
```
marsella, line 32: nit: fix comment - this is serializing the whole key, not just the id.

lock-keeper-key-server/src/database/secrets.rs
```diff
@@ -77,11 +96,41 @@ impl Database {
...
         let _ = collection
             .find_one_and_update(filter, update, None)
             .await?
             .ok_or(LockKeeperServerError::InvalidAccount)?;
         Ok(())
     }
+
+    /// Get a [`User`]'s [`StoredSigningKeyPair`] based on its [`KeyId`]
+    pub async fn get_user_signing_key(
```
marsella, line 109: todo/question: I would normally expect a test to make sure this only gets a key if it belongs to the user. But I know there was something funky going on with the db test environment. Could you either write the test and `#[ignore]` it or put a note in an issue somewhere that this function (and the `add` function) need tests?

lock-keeper-key-server/src/server.rs
```diff
@@ -179,6 +191,16 @@ impl LockKeeperRpc for LockKeeperKeyServer {
             .await?)
     }
 
+    async fn retrieve_audit_events(
```
marsella, line 194: question: You didn't actually add this method, right? You just added `retrieve_signing_key` and the diff got stressed out?

lock-keeper-tests/src/end_to_end.rs
```diff
@@ -266,6 +267,46 @@ async fn tests() -> Vec<Test> {
...
+                ),
+                (
+                    ImportSigningKey,
+                    Outcome {
+                        expected_error: None,
+                    },
+                ),
+            ],
+        ),
+        Test::new(
```
marsella, line 287: suggestion: Can you add a test that's like, register -> set a key id (maybe generate a different kind of key) -> export signing key? I think there are a couple versions of "bad key id" (e.g. wrong key type, non-existent key id, locally-generated key that we don't support exporting yet), but I'm not sure how many can be tested at this point.

Edit: Hm, I see that the ID storage for generated keys is different from imported keys in the testing framework. Maybe you can add a "set key id" function like the one you made for generated keys.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -11,7 +11,7 @@ use super::{generic::AssociatedData, CryptoError, Encrypted, KeyId, StorageKey};
 ///
 /// This can be generated locally by the client or remotely by the server.
 #[allow(unused)]
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
```
marsella, line 14: nit: If you take the approach on a previous comment to make an `Export` type to send this around, please remove the ser/de derivations here.

marsella (commented):
Sorry for long comments on this one -- I was figuring stuff out as I went. You probably already know most of the stuff in there.

My main questions stem out of the abstractions we're using, especially the temporary / intermediary things we're adding to patch holes we plan to fill in later. I think we should document such things a bit more aggressively, even if we plan to take them out, so we don't get confused and reuse them in inappropriate ways.

Edit: The diff also seemed a bit funky to me -- a lot of added `import` methods that I thought would have disappeared when you changed the base. Can you double-check before you merge that none of that code got duplicated somehow?

lock-keeper-client/src/api.rs
```diff
@@ -145,6 +177,21 @@ impl LockKeeperClient {
...
+        let mut client_channel = Self::create_channel(
+            &mut self.tonic_client(),
+            ClientAction::ImportSigningKey,
+            self.account_name(),
+        )
+        .await?;
+        self.handle_import_signing_key(&mut client_channel, key_material)
+            .await
+    }
+
```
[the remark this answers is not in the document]
vidyaap: No, I did some rebasing, but it messed up the history. I think once import is actually merged, I can make the history cleaner again.

vidyaap:
@marsella I was going to respond to individual comments related to the weird history but decided that was more confusing so - I tried rebasing with develop and then with 258-import but something went wrong and it created duplicates of the import commits anyway. I'm hoping once import is merged, I can just remove these duplicates from here and make this look cleaner.

lkic/src/storage.rs
```diff
@@ -214,6 +214,14 @@ impl Display for Entry {
                 })?;
                 format!("Arbitrary Key - {}", hex::encode(&bytes))
             }
+            RetrieveResult::SigningKey(signing_key) => {
+                let bytes = signing_key_bytes(signing_key).map_err(|e| {
+                    // Avoid error inception
+                    let _ = writeln!(f, "{e}");
+                    std::fmt::Error
+                })?;
+                format!("Signing Key Pair - {}", hex::encode(&bytes))
```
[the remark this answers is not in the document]
vidyaap: This is part of a match that's part of a Display impl. It's returning the appropriate string based on key type and only actually writes it outside of the match.

lock-keeper-tests/src/end_to_end.rs
vidyaap: Do you think we’ll have tests that need to generate and import in the same test? Otherwise I might just change the key name of the stored key ID and run "Register", "Generate", "Export" to test the case you're talking about.

lkic/src/storage.rs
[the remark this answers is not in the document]
marsella: Ah, I see. Thanks.

lock-keeper-tests/src/end_to_end.rs
marsella: I can imagine a test case like, "you can generate multiple types of keys via different endpoints, then retrieve all of them in a different order," but I think that will require more refactoring of the test harness anyway -- I'm pretty confident at some point we will want to test generating multiple keys regardless of type. I'd say it's fine to reuse the single stored key id for now.

lock-keeper-key-server/src/database/secrets.rs
picklenerd: `server_created_signing_keys` sounds better to me.

lock-keeper-client/src/api.rs
```diff
@@ -33,6 +34,7 @@ use tracing::error;
 pub enum RetrieveResult {
     None,
     ArbitraryKey(LocalStorage),
+    SigningKey(PlaceholderEncryptedSigningKeyPair),
```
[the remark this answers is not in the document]
picklenerd: > The PlaceholderEncryptedSKP is a placeholder for the server-side storage encryption scheme 

Oh right, yeah this is for stuff generated on the server that we have no good way to encrypt right now since the user isn't involved.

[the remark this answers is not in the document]
vidyaap: Gotcha, I changed it to use the `Export` type Marcella proposed. I'm not loving the conversion from it to SigningKeyPair since it just converts the "key_material" bytes into associated data and makes a SKP out of that, but I think that can't be changed until #262 is merged, right?

Also wondering if it would be better to just have one wrapper type around bytes for key material, like the `KeyMaterial` type we had at some point? Or are there different properties for the two?

lock-keeper-key-server/src/database/secrets.rs
vidyaap: Two thoughts on this:
- I need a way to create secrets on the fly to test this properly I think, but I'm not sure how to access the client's storage key in these tests.
- These feel more like mongodb tests to me, since we're just making sure that the mongodb filter works. Not saying that it's not worth checking anyway if we got the right secret back, but is there anything else we want to check besides it being the correct user ID?

picklenerd: I would wait for the other PR to be merged before adding any new mongodb tests. Maybe we could make a new ticket for DB tests involving secrets. I wanted to write some for remote generate too.

picklenerd: I added #293 for this.

lock-keeper-client/src/api.rs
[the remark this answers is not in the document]
marsella: Hm, that sounds kind of sketch. I would rather throw away the `key_material` for now and put it in properly once #262 is merged (maybe in #262 because I'm guessing this will get merged first). But that will probably break all the tests, so maybe fine... we'll get a compile error either way. I'll comment there as well, but why don't you document the correct behavior at that point so there's no confusion when we go to fix it.

At this point, I'm not sure if they have different properties beyond the readability (e.g. you know which one is for import), so having a single thing seems fine. We can always separate them later.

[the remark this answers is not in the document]
marsella: Actually, export should probably include the associated data from the key (see [3.v.b in the spec](https://github.com/boltlabs-inc/key-mgmt-spec/blob/main/system-functionalities.md#retrieve-a-secret), while import doesn't include any user-specified AD at this point.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -180,6 +188,29 @@ impl Import {
...
+        }
+    }
+}
+
+impl Export {
+    pub fn into_signing_key(self) -> Result<SigningKeyPair, LockKeeperError> {
+        let associated_data = self.key_material.try_into()?;
+        Ok(SigningKeyPair {
+            context: associated_data,
+        })
```
marsella, line 210: todo: add documentation explaining why this is wrong and when it will get fixed. I'd also change the variable name to something obnoxious so it's clear that it needs to change and that this is not  _actually_ the associated data. 
```
// TODO #235: Once the signing key pair has a way to hold key material, 
// copy the key material and associated data from `Export` directly to the 
// corresponding fields in the `SigningKeyPair`. Do not put the key material into the
// associated data field -- they have very different security properties!
```

lock-keeper-client/src/api/retrieve.rs
marsella: nit: thanks! Missing periods at the end of these sentences (here and at `handle_retrieve`).

lock-keeper-key-server/src/database/secrets.rs
marsella: Thanks!

lock-keeper-client/src/api.rs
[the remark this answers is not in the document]
vidyaap: Ah okay, that makes sense. Also, I feel a little better about the Export -> SKP conversion since the SKP -> Export conversion uses the SKP -> bytes conversion which also just converts the AD into bytes, so at least it would be the same info back and forth. But I'll still leave a comment there for it.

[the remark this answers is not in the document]
marsella: Thanks. I also updated #235 with an explicit reminder to fix it.

marsella (approved):
Looks good, thanks for the changes.

marsella:
> This export is modeled after the arbitrary secrets export, which in turn is based on arbitrary secrets retrieve. Since we didn't talk about implementing retrieve for signing keys yet, I wasn't sure if we would want to wrap the retrieved signing key in a `LocalStorage` like we do for arbitrary keys, so I have them being returned plain for now. Is that okay?

I think we will want this to be wrapped in `LocalStorage` of some kind at some point, since we'd expect the calling application to treat / store them in the same way. That is, the point of a `LocalStorage` type was to apply whatever custom formatting or encryption is necessary to prepare stuff for local storage. I think it's fine to either do this here or make a new issue to update the `LocalStorage` type to handle multiple key types.

> The import spec doesn't say to return the imported signing key to the user, so currently I don't have a way to write an export test that checks that the key material is the same. Should I be returning the key in import or is there a different way to check this?

In boltlabs-inc/key-mgmt-spec#135, I'm going to have import ("only to the key server") return the public key, but I think you're talking about the entire signing key, right? Can you save the bytes that you originally passed to import? You'll have to be a little tricky comparing raw bytes (passed to import) to a complete `SigningKeyPair` but I can point you toward other tests where I've done something similar.

vidyaap:
> ... Can you save the bytes that you originally passed to import? You'll have to be a little tricky comparing raw bytes (passed to import) to a complete SigningKeyPair but I can point you toward other tests where I've done something similar.

I don't think I'd be able to compare within this PR actually, because even if I saved the input bytes, those get thrown out when we make the SKP, so I still wouldn't have anything to compare to.  I can make issues to add this test and to update the LocalStorage type for this PR if that works?

vidyaap:
Made #297 and #299 to address the LocalStorage and the import/export comparison respectively.

indomitableSwan:
> Closes #281
> 
> This adds the export functionality for signing keys. The previous PR (#263) only implemented it for arbitrary secrets.
> 
> Questions for @indomitableSwan:
> 
> * This export is modeled after the arbitrary secrets export, which in turn is based on arbitrary secrets retrieve. Since we didn't talk about implementing retrieve for signing keys yet, I wasn't sure if we would want to wrap the retrieved signing key in a `LocalStorage` like we do for arbitrary keys, so I have them being returned plain for now. Is that okay?

I agree with Marcella's response on this.

> * The import spec doesn't say to return the imported signing key to the user, so currently I don't have a way to write an export test that checks that the key material is the same. Should I be returning the key in import or is there a different way to check this?

Commented on #299.

PR #288, opened 2022-10-13
Add docker cleanup to teardown task #285

### LINKED ISSUE #285: Delete stale docker images on server teardown
When you teardown and restart the server with docker, it leaves the old images behind. This collection of stale images quickly grows in size until you inevitably get an error message saying that Docker ran out of space. 

We can steps to the stop task to delete these stale images. We should also update the README to suggest `docker image prune` over `docker system prune` since the latter deletes the cargo cache and makes rebuild times very long.

picklenerd:
This PR adds some extra cleanup steps to the `teardown` task so that Docker is less likely to run out of space.

It also updates the `restart` task to call `teardown` instead of `stop`. This makes it take a little bit longer to run but prevents the scenario where you can only run `cargo make restart` 2-3 times before you run out of space and have to do a `docker system prune`.

Closes #285

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -17,7 +17,7 @@ alias = "teardown-server"
 
 # Stops the server, recompiles it, and restarts it
 [tasks.restart]
-dependencies = ["stop", "start"]
+dependencies = ["teardown", "start"]
 
 # Start a key server and monitor its output.
 [tasks.start-server]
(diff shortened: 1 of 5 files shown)


picklenerd:
> Thanks, this is definitely faster than before! I was wondering how long you were expecting these to take though? On my machine, it takes ~13 minutes to start for the first time and ~8.5 to restart. That still feels pretty long to me, but I was curious if that was just on my laptop.

That's definitely not right. Mine was taking a little over a minute to restart last night from this branch, although that was based on one line code changes. I'll do some tests on my machine and see how it behaves in different situations.

vidyaap (approved):
Thanks for helping figure out my build time issue! Takes about a minute to restart now, which makes more sense, so this looks good to me.

PR #291, opened 2022-10-15
Move database tests to tests crate #268

### LINKED ISSUE #268: Move database unit tests to `lock-keeper-tests` crate
The database unit tests require a MongoDB server running in the background. This is more inline with how integration tests should work. Unit tests running through `cargo test` should be able to run without anything else running.

- [ ] Add a module called `database` to `lock-keeper-tests`.
- [ ] Move all of the database tests to this new module and remove `#[tokio::test]` from them.
- [ ] Add a function to run all database tests to the new `database` module. This function should simply call all of the database test functions moved over.
- [ ] Update `main.rs` to run the database tests.

picklenerd:
This PR creates an integration testing framework for database tests and moves them out of the unit tests so that unit tests can be running without requiring a server to be running.

Tests can now be run with the flag `--test-type` to select which type of tests to run. Options are `e2e`, `integraton`, and `all`. CI and `cargo make` have been updated accordingly.

Closes #268


vidyaap (changes requested):
Thanks for setting this up! I noticed a couple of errors in the e2e tests that weren't being caught the way they were expected to, and I wasn't sure which file(s) to leave the comment on but:
- It doesn't seem to catch when we expect an error and don't get one.
- When the `assert_eq!` checks in `check_audit_logs()` fail, it panics out of the whole process. Could these be if-blocks instead?

Also, I noticed there were a number of cleanup changes to the audit log DB tests that are missing now. I also couldn't find them in their original PR, so not sure what happened to them, but I'm happy to add them back in; let me know!

picklenerd:
We discussed the e2e tests in a call today and I don't think this PR actually modifies them in any way. This regression may have come from a previous PR that got merged.

vidyaap (approved):
Looks good to me! Export was merged in the middle though, so that may have to be rebased into here.

PR #301, opened 2022-10-19
Initial session keys implementation.

gatoWololo:
This PR is now ready for review.

It looks like our existing e2e test already exercise the code path for overwriting an existing session key. Is this expected?

### DIFF
--- a/lock-keeper-key-server/src/error.rs
+++ b/lock-keeper-key-server/src/error.rs
@@ -1,3 +1,4 @@
+use crate::server::session_key_cache::SessionKeyCacheError;
 use thiserror::Error;
 use tonic::Status;
 
@@ -5,6 +6,9 @@ use tonic::Status;
 pub enum LockKeeperServerError {
     #[error("Could not get service.")]
     MissingService,
+
+    #[error("Error in session keys cache.")]
+    SessionKeyCache(#[from] SessionKeyCacheError),
 
     // Infrastructure errors
     #[error("Invalid OPAQUE directory")]
@@ -55,7 +59,12 @@ impl From<opaque_ke::errors::ProtocolError> for LockKeeperServerError {
 
 impl From<LockKeeperServerError> for Status {
     fn from(error: LockKeeperServerError) -> Status {
+        use crate::server::session_key_cache::SessionKeyCacheError::*;
+
         match error {
+            LockKeeperServerError::SessionKeyCache(ExpiredSessionKey) => {
+                Status::unauthenticated("Session key expired")
+            }
             // Errors that are safe to return to the client
             LockKeeperServerError::AccountAlreadyRegistered
             | LockKeeperServerError::InvalidAccount
@@ -71,6 +80,7 @@ impl From<LockKeeperServerError> for Status {
 
             // Errors that the client should not see
             LockKeeperServerError::MissingService
+            | LockKeeperServerError::SessionKeyCache(_)
             | LockKeeperServerError::Hyper(_)
             | LockKeeperServerError::Io(_)
             | LockKeeperServerError::InvalidOpaqueDirectory
--- a/lock-keeper-key-server/src/server.rs
+++ b/lock-keeper-key-server/src/server.rs
@@ -1,6 +1,7 @@
 pub(crate) mod opaque_storage;
 mod operation;
 mod service;
+pub(crate) mod session_key_cache;
 
 pub(crate) use operation::Operation;
 pub use service::start_lock_keeper_server;
@@ -16,11 +17,13 @@ use lock_keeper::{
     types::{database::user::AccountName, operations::ClientAction, Message, MessageStream},
 };
 
+use crate::server::session_key_cache::SessionKeyCache;
 use rand::{rngs::StdRng, SeedableRng};
 use std::sync::Arc;
 use tokio::sync::Mutex;
 use tonic::{Request, Response, Status};
 
+///
 #[allow(unused)]
 #[derive(Debug)]
 pub struct LockKeeperKeyServer {
@@ -28,6 +31,7 @@ pub struct LockKeeperKeyServer {
     db: Arc<Database>,
     service: Arc<Service>,
     rng: Arc<Mutex<StdRng>>,
+    session_key_cache: Arc<Mutex<SessionKeyCache>>,
 }
 
 impl LockKeeperKeyServer {
@@ -43,10 +47,11 @@ impl LockKeeperKeyServer {
             db: Arc::new(db),
             service,
             rng: Arc::new(Mutex::new(rng)),
+            session_key_cache: Arc::new(Mutex::new(SessionKeyCache::default())),
         })
     }
 
-    pub fn context(
+    pub(crate) fn context(
         &self,
         request: &Request<tonic::Streaming<Message>>,
     ) -> Result<Context, LockKeeperServerError> {
@@ -75,6 +80,7 @@ impl LockKeeperKeyServer {
             account_name,
             action,
             key_id: None,
+            session_key_cache: self.session_key_cache.clone(),
         })
     }
 
@@ -95,13 +101,15 @@ impl LockKeeperKeyServer {
 }
 
 #[derive(Debug)]
-pub struct Context {
+pub(crate) struct Context {
     pub db: Arc<Database>,
     pub service: Arc<Service>,
     pub rng: Arc<Mutex<StdRng>>,
     pub account_name: AccountName,
     pub action: ClientAction,
     pub key_id: Option<KeyId>,
+    /// Our user session keys are held in this cache after authentication.
+    pub session_key_cache: Arc<Mutex<SessionKeyCache>>,
 }
 
 #[tonic::async_trait]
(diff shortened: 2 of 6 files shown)


lock-keeper-key-server/src/error.rs
```diff
@@ -6,6 +7,9 @@ pub enum LockKeeperServerError {
     #[error("Could not get service.")]
     MissingService,
 
+    #[error("Error in session key cache.")]
```
gatoWololo, line 10: Is this the best way to handle errors? I made my own error enum and then created a wrapper variant `SessionKeyCache` under `LockKeeperServerError`. I worry about `LockKeeperServerError` growing unbounded if I just add my variants directly here.

lock-keeper-key-server/src/server.rs
```diff
@@ -43,10 +46,11 @@ impl LockKeeperKeyServer {
             db: Arc::new(db),
             service,
             rng: Arc::new(Mutex::new(rng)),
+            session_keys_cache: Arc::new(RwLock::new(SessionKeyCache::default())),
         })
     }
 
-    pub fn context(
+    pub(crate) fn context(
```
gatoWololo, line 53: Felt weird that the context was just `pub`, it only needs to be `pub(crate)`

lock-keeper-key-server/src/server/session_key_cache.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub fn new(expiration: Duration) -> Self {
+        Self {
+            cache: HashMap::new(),
+            expiration,
+        }
+    }
+
+    pub fn insert(
+        &mut self,
+        user: UserId,
```
gatoWololo, line 36: Inserting a `UserId` requires ownership of it, `UserId` are `Clone`. So this is fine?

lock-keeper-key-server/src/server/session_key_cache.rs
```diff
@@ -0,0 +1,103 @@
...
+        key: OpaqueSessionKey,
+    ) -> Result<(), SessionKeyCacheError> {
+        let existing = self.cache.insert(user, (Instant::now(), key));
```
gatoWololo, line 39: I just realized I am inserting the new key and then returning an error if a previous key was overwritten... What should be the correct behavior if the key already exists?

lock-keeper-key-server/src/server/session_key_cache.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+impl Default for SessionKeyCache {
+    fn default() -> Self {
+        SessionKeyCache::new(SessionKeyCache::DEFAULT_EXPIRATION_TIME)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
gatoWololo, line 83: Logic for `SessionKeyCache` still needs to be tested. No need to scrutinize it yet :sweat_smile:

lock-keeper-key-server/src/server/session_key_cache.rs
```diff
@@ -0,0 +1,103 @@
...
+#[derive(Debug)]
+pub(crate) struct SessionKeyCache {
+    cache: HashMap<UserId, (Instant, OpaqueSessionKey)>,
+    /// Time a key is considered valid for.
+    expiration: Duration,
+}
+
+#[allow(unused)]
+impl SessionKeyCache {
+    // TODO: What is a good default expiration time?
```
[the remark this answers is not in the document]
picklenerd: This is probably a question for @indomitableSwan but I'd say 60 seconds is a good start. That should give us plenty of time to complete individual e2e tests without re-authenticating.

lock-keeper-key-server/src/error.rs
```diff
@@ -6,6 +7,9 @@ pub enum LockKeeperServerError {
...
+    SessionKeyCache(SessionKeyCacheError),
```
gatoWololo, line 11: Is this the best way to handle errors? I made my own error enum and then created a wrapper variant SessionKeyCache under LockKeeperServerError. I worry about LockKeeperServerError growing unbounded if I just add my variants directly here.

lock-keeper-key-server/src/server/session_key_cache.rs
picklenerd: Yep this is how I prefer to pass values that need to be owned.

lock-keeper-key-server/src/error.rs
picklenerd: I think this should be ok as long as you add `#[from]` in front of the parameter so they convert automatically.

gatoWololo:
CI is failing due to diff in formatting. I thought I ran cargo format through the code before pushing through my IDE, but looks like I didn't do it correctly.

Next time, I'll also do a better job with highlighting the relevant code parts for my comments...

lock-keeper-key-server/src/error.rs
```diff
@@ -55,7 +59,13 @@ impl From<opaque_ke::errors::ProtocolError> for LockKeeperServerError {
 
 impl From<LockKeeperServerError> for Status {
     fn from(error: LockKeeperServerError) -> Status {
+        use crate::server::session_key_cache::SessionKeyCacheError::*;
+
         match error {
+            LockKeeperServerError::SessionKeyCache(ref reason) => match reason {
+                SessionKeyExists | MissingSessionKey => Status::internal(error.to_string()),
+                ExpiredSessionKey => Status::invalid_argument(error.to_string()),
```
picklenerd, line 67: I don't think this is an invalid argument since it's all kind of implicit based on the user id.

lock-keeper-key-server/src/server.rs
```diff
@@ -28,6 +30,7 @@ pub struct LockKeeperKeyServer {
     db: Arc<Database>,
     service: Arc<Service>,
     rng: Arc<Mutex<StdRng>>,
+    session_keys_cache: Arc<RwLock<SessionKeyCache>>,
```
picklenerd, line 33:
```suggestion
    session_key_cache: Arc<RwLock<SessionKeyCache>>,
```

picklenerd: I think this should match the type name.

lock-keeper-key-server/src/server/session_key_cache.rs
```diff
@@ -0,0 +1,103 @@
...
+        let existing = self.cache.insert(user, (Instant::now(), key));
+
+        if existing.is_some() {
+            return Err(SessionKeyCacheError::SessionKeyExists);
+        }
+
+        Ok(())
+    }
+
+    pub fn get_key(&mut self, user: UserId) -> Result<OpaqueSessionKey, SessionKeyCacheError> {
```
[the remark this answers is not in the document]
picklenerd: I think it should return a reference, assuming that doesn't cause problems with the `RwLock`. I agree that the consumer should clone it themselves.

lock-keeper/src/types/database/user.rs
```diff
@@ -31,6 +31,13 @@ impl UserId {
         Ok(Self(uuid.to_string()))
     }
 
+    // #[cfg(test)]
```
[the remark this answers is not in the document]
picklenerd: It's not ok to expose this. The type is guaranteed to be valid in the current code but wouldn't be with this method. You should be able to pass a `StdRng` if you need to generate one in the code.

lock-keeper-key-server/src/server/session_key_cache.rs
indomitableSwan: quick fly-by question that may betray the length of time spent on this: Are we imagining allowing users to be logged in on multiple devices at once?

lock-keeper-key-server/src/server/session_key_cache.rs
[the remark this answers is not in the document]
indomitableSwan: context question: who is the caller here? We didn't intend the `session_key` to be passed out for use by the calling application.

lock-keeper-key-server/src/server/session_key_cache.rs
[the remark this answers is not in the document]
indomitableSwan: I'm fine with this for now.

lock-keeper-key-server/src/server/session_key_cache.rs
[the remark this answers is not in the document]
gatoWololo: The caller is the Key Server, this is an internal interface used by the `KeyServer` struct during handling of operations.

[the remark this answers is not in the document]
gatoWololo: Current implementation returns a reference to the key.

lock-keeper-key-server/src/server/session_key_cache.rs
```diff
@@ -0,0 +1,212 @@
...
+        }
+    }
+
+    pub fn get_key(&mut self, user: &UserId) -> Result<&OpaqueSessionKey, SessionKeyCacheError> {
+        // The control flow for is purposely inside out and requires two lookups.
+        // This is necessary due to a shortcoming of the borrow checker.
+        // See: https://github.com/rust-lang/rfcs/blob/master/text/2094-nll.md#problem-case-3-conditional-control-flow-across-functions
+
+        // We could avoid this by cloning the key and returning an owned copy. The Entry
+        // Hashmap API does not help here.
```
picklenerd, line 59: Looking at this code, I think it would be better to clone it here. We know that every consumer is going to call this from within a `Mutex` so returning a reference means they're going to hold the lock for longer than we want them to. If we return an owned value, we can require short lived mutex locks in code review.

picklenerd (approved):
LGTM!

PR #302, opened 2022-10-19
Add integration tests for secrets #293

### LINKED ISSUE #293: Write database tests for secrets
Some of the new code involving secrets should have database tests.

- [x] Add test cases for all functions that add new secrets to a user a test that the User can still be deserialized from the database data.
- [x] Add a test case to ensure `get_user_signing_key` only gets a key if it belongs to the user. See https://github.com/boltlabs-inc/key-mgmt/pull/287#discussion_r996164526

vidyaap:
Closes #293 

This adds the tests for checking user serializability after adding secrets and for checking that we're getting the right secrets back from the DB.

I'm having a little trouble justifying the second test though. It feels like this should be covered by the e2e tests already. The original comment was to check that the user ID is correct, but the key itself doesn't have any user information and it didn't make sense to me to repeat the mongoDB filter code here to manually check the user ID if that's how we got the secret back in the first place.  Right now, I'm just saving the key that we initially store and then comparing to what we get back. Is this sufficient?

### DIFF
--- a/lock-keeper-tests/src/database.rs
+++ b/lock-keeper-tests/src/database.rs
@@ -1,16 +1,19 @@
 //! Database integration tests
 
 pub mod audit_event;
+pub mod secret;
 pub mod user;
 
+use generic_array::{typenum::U64, GenericArray};
 use std::{ops::Deref, str::FromStr};
 
 use lock_keeper::{
     config::server::DatabaseSpec,
+    crypto::{Encrypted, OpaqueExportKey, StorageKey},
     types::database::user::{AccountName, UserId},
 };
 use lock_keeper_key_server::database::Database;
-use rand::{rngs::StdRng, SeedableRng};
+use rand::{rngs::StdRng, Rng, SeedableRng};
 
 use crate::utils::{server_registration, tagged};
 
@@ -21,10 +24,12 @@ pub async fn run_tests() -> anyhow::Result<()> {
 
     let audit_event_results = audit_event::run_tests().await?;
     let user_results = user::run_tests().await?;
+    let secret_results = secret::run_tests().await?;
 
     // Report results after all tests finish so results show up together
     audit_event_results.report("audit event tests");
     user_results.report("user tests");
+    secret_results.report("secret tests");
 
     println!();
 
@@ -85,6 +90,34 @@ impl TestDatabase {
     pub async fn drop(self) -> anyhow::Result<()> {
         self.mongo.drop(None).await?;
         Ok(())
+    }
+
+    /// Create an export key for testing using random bytes.
+    fn create_test_export_key(rng: &mut StdRng) -> anyhow::Result<OpaqueExportKey> {
+        let mut key = [0_u8; 64];
+        rng.try_fill(&mut key)?;
+
+        // We can't create an export key directly from bytes so we convert it to a
+        // GenericArray first.
+        let key: GenericArray<u8, U64> = key.into();
+
+        Ok(key.into())
+    }
+
+    /// Create a storage key for a test user.
+    pub fn create_test_storage_key(
+        &self,
+        user_id: &UserId,
+    ) -> anyhow::Result<(Encrypted<StorageKey>, OpaqueExportKey)> {
+        let mut rng = StdRng::from_entropy();
+
+        // Create a storage key
+        let export_key = Self::create_test_export_key(&mut rng)?;
+        let storage_key = export_key
+            .clone()
+            .create_and_encrypt_storage_key(&mut rng, user_id)?;
+
+        Ok((storage_key, export_key))
     }
 
     /// Creates a test user with a randomized name and returns the id and
--- a/lock-keeper-tests/src/database/secret.rs
+++ b/lock-keeper-tests/src/database/secret.rs
@@ -1,0 +1,128 @@
+//! Integration tests for secret objects in the database
+
+use colored::Colorize;
+use lock_keeper::{
+    crypto::{Import, KeyId, Secret, SigningKeyPair, StorageKey},
+    types::database::user::UserId,
+};
+use lock_keeper_key_server::database::Database;
+use rand::{rngs::StdRng, Rng, SeedableRng};
+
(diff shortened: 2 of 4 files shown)


lock-keeper-tests/src/database.rs
```diff
@@ -87,6 +92,37 @@ impl TestDatabase {
...
+
+        // We can't create an export key directly from bytes so we convert it to a
+        // GenericArray first.
+        let key: GenericArray<u8, U64> = key.into();
+
+        key.into()
+    }
+
+    /// Creates a test user with a randomized name and returns the id and
+    /// account name.
```
picklenerd, line 109: I think this comment is a copy/paste error.

lock-keeper-tests/src/utils.rs
```diff
@@ -10,7 +10,7 @@ use opaque_ke::{
 use rand::{distributions::Alphanumeric, Rng};
 
 /// Add random text to the end of a string
-/// # Exmaple
+/// # Example
```
picklenerd, line 13: 😆

picklenerd:
> I'm having a little trouble justifying the second test though. It feels like this should be covered by the e2e tests already. The original comment was to check that the user ID is correct, but the key itself doesn't have any user information and it didn't make sense to me to repeat the mongoDB filter code here to manually check the user ID if that's how we got the secret back in the first place. Right now, I'm just saving the key that we initially store and then comparing to what we get back. Is this sufficient?

Yeah I don't really see the point of that one. I think this is one of those cases where you would do a custom DB query in the test that returns the user id along with the secret and then compare them. That might protect against some changes in the real db code but I don't think it would make me a whole lot more confident.

vidyaap:
> I think this is one of those cases where you would do a custom DB query in the test that returns the user id along with the secret and then compare them.

Do you mean: get the secrets using the get functions, get them again but with the user ID using a custom mongo query, compare the secrets?
Or should I just remove this second test entirely?

picklenerd:
> Do you mean: get the secrets using the get functions, get them again but with the user ID using a custom mongo query, compare the secrets? Or should I just remove this second test entirely?

Yeah that's what I was thinking. That would at least make this test a decent sanity check.

lock-keeper-tests/src/database.rs
vidyaap: oops 😅

vidyaap:
> Yeah that's what I was thinking. That would at least make this test a decent sanity check.

Ok I made that change so it manually pulls the secret after we get it using our DB function. I still don't know how much it adds as a test though - if the mongoDB filter doesn't work then this test wouldn't pass until there was a change in the mongoDB source code, and we are already checking that the key material matches in the generate/retrieve tests (and soon in import and export). 

I feel like we don't really need it, but wanted to run this past @marsella first too.

lock-keeper-tests/src/database.rs
```diff
@@ -87,6 +92,36 @@ impl TestDatabase {
         Ok(())
     }
 
+    /// Create an export key for testing using random bytes.
+    fn create_test_export_key(rng: &mut StdRng) -> OpaqueExportKey {
+        let mut key = [0_u8; 64];
+        rng.try_fill(&mut key)
+            .expect("Failed to generate random key");
```
picklenerd, line 99: I think this should return a result so it doesn't get swallowed up by our custom panic handler.

lock-keeper-tests/src/database.rs
```diff
@@ -87,6 +92,36 @@ impl TestDatabase {
...
+        user_id: &UserId,
+    ) -> anyhow::Result<(Encrypted<StorageKey>, OpaqueExportKey)> {
+        let mut rng = StdRng::from_entropy();
+
+        // Create a storage key
+        let export_key = Self::create_test_export_key(&mut rng);
+        let storage_key = export_key
+            .clone()
+            .create_and_encrypt_storage_key(&mut rng, user_id)
+            .unwrap();
```
picklenerd, line 120: This should return an error instead of unwrapping.

lock-keeper-tests/src/database/secret.rs
```diff
@@ -0,0 +1,214 @@
...
+    let mut rng = StdRng::from_entropy();
+
+    // Randomly generate secrets of all 3 types
+    let key_infos = create_secrets_for_random_users(&db, &mut rng, &storage_key, &user_id).await?;
+
+    // Check that each 'get' method returns secrets for the right user
+    verify_user_keys(&db, key_infos, &user_id).await?;
+
+    Ok(())
+}
```
picklenerd, line 214: In general, I prefer to have to most relevant functions/types at the top of the file and all of the helper stuff below it. I think the test cases would be the most relevant in this case.

@gatoWololo do you have a preference for this sort of this?

gatoWololo: I agree. Having consts and types first. Followed by the main functions, then imported functions.

Some people sort functions in alphabetical order, I feel indifferent to this.

vidyaap: 👍

marsella:
> Ok I made that change so it manually pulls the secret after we get it using our DB function. I still don't know how much it adds as a test though - if the mongoDB filter doesn't work then this test wouldn't pass until there was a change in the mongoDB source code, and we are already checking that the key material matches in the generate/retrieve tests (and soon in import and export).
> 
> I feel like we don't really need it, but wanted to run this past @marsella first too.

The test I was imagining was
1. Make two users.
2. User A saves a secret with `key_id`
3. User B tries to retrieve a secret with `key_id`

Step 3 fails because the key doesn't belong to user B.

This doesn't seem to be quite what you're testing (I'm looking at `get_secret_methods_verify_user_id`).

vidyaap:
Ah okay, I think I came at this from the other end - "the keys you get don't include anyone else's keys." The test you're suggesting is a lot cleaner though, so I'll push that version.

marsella:
Ah, I see. Thanks for changing, this looks good to me.

picklenerd (approved):
LGTM!

PR #304, opened 2022-10-19
Add .dockerignore file #303

### LINKED ISSUE #303: Add a .dockerignore file to the repo
Currently, when we run docker-compose to build the image for the server, it copies everything in key-mgmt to the server which makes start/restart times quite large. Adding a .dockerignore file to ignore the following files decreases the start/restart times:
- [ ] the target folder
- [ ] files generated by IDE (e.g. `.idea/` for IntelliJ)

vidyaap:
Closes #303 

This just adds the target folder and the IntelliJ-generated folder to an ignore file so the Docker restarts don't take as long.  It brings my restart time down to about 20 seconds. I'd like if folks ran this and made sure their tests still run correctly with this addition. If there are other files that would be useful to add to this, I can do that too.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+target/
+.idea/


picklenerd (approved):
Tests passed. Then tore everything down and did a full prune. Ran it again and passed. Changed a line and did a restart, tests passd.

gatoWololo:
>  It brings my restart time down to about 20 seconds

Me too! So fast.

All test pass :heavy_check_mark:

PR #305, opened 2022-10-19
Split e2e tests into modules #270

### LINKED ISSUE #270: Split end-to-end tests into modules
The e2e tests are currently implemented as a large `match` block covering each of the testing operations. The `end_to_end` module is growing quickly and this is becoming difficult to manage. This module should be split into submodules for each operation. This is a quick and easy step we can take to improve the integration tests before we have time to work on bigger changes.

- [x] Create a submodule under `end_to_end` for each testing `Operation`.
- [x] Move operation-specific code to the appropriate module.
- [x] Call into the modules from the match block in the main `end_to_end` module.

picklenerd:
This PR splits the end-to-end tests into separate modules for each operation. I also added a module for the test cases and removed the custom error type.

Closes #270

### DIFF
--- a/lock-keeper-tests/src/end_to_end.rs
+++ b/lock-keeper-tests/src/end_to_end.rs
@@ -1,5 +1,9 @@
 //! End-to-end testing framework and test definitions
 
+pub mod operations;
+pub mod test_cases;
+
+use anyhow::anyhow;
 use colored::Colorize;
 use lock_keeper::{
     config::client,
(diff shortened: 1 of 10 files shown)


lock-keeper-tests/src/end_to_end.rs
```diff
@@ -412,215 +96,65 @@ impl Test {
...
+            config,
         }
     }
 
-    async fn execute(&mut self, config: &client::Config) -> Result<(), anyhow::Error> {
+    async fn execute(mut self) -> Result<(), anyhow::Error> {
         use Operation::*;
 
-        for (op, expected_outcome) in &self.operations {
+        let operations: Vec<(Operation, Outcome)> = self.operations.drain(..).collect();
```
gatoWololo, line 106: I would do:
```rust
let operations: Vec<(Operation, Outcome)> = std::mem::take(&mut self.operations);`
```
here.

gatoWololo (approved):
I mostly see how this works now. Looks good! I'm glad this is broken up. I just have one tiny suggestion.

lock-keeper-tests/src/end_to_end.rs
picklenerd: Nice! I never use this so I didn't realize that it just replaces it with `Default`. This is a much better solution!

gatoWololo: Yeah! The other useful function is `std::mem::swap` where you specify the value instead of using `Default`. The above could be written as:

```rust
let operations: Vec<(Operation, Outcome)> = std::mem::swap(&mut self.operations, Vec::new());`
```

PR #311, opened 2022-10-24
Make LocalStorage enum to handle multiple key types #297

### LINKED ISSUE #297: Update LocalStorage type to handle multiple key types
Right now the `LocalStorage` type is used to prepare arbitrary secrets for local storage on the calling app. We will want this to be able to handle signing keys as well.

Retrieved signing keys (not exported) will need to be wrapped in this type, the way retrieved arbitrary secrets currently are.  The type is also meant to apply any custom formatting or encryption is necessary to prepare these key types for local storage.

vidyaap:
Closes #297 

Also got rid of `RetrieveResult` in the process; felt redundant.

### DIFF
--- a/lock-keeper-client/src/api.rs
+++ b/lock-keeper-client/src/api.rs
@@ -28,18 +28,10 @@ use rand::{rngs::StdRng, SeedableRng};
 use serde::{Deserialize, Serialize};
 use tracing::error;
 
-/// Ways of returning a key from the retrieval process based on usage
-/// [`RetrieveContext`]
+/// Wrapper for secrets prepared for local storage
 #[derive(Debug, Deserialize, Serialize)]
-pub enum RetrieveResult {
-    None,
-    ArbitraryKey(LocalStorage),
-    SigningKey(Export),
-}
-
-#[derive(Debug, Deserialize, Serialize)]
-pub struct LocalStorage {
-    pub secret: Secret,
+pub struct LocalStorage<T> {
+    pub material: T,
 }
 
 impl LockKeeperClient {
@@ -126,15 +118,11 @@ impl LockKeeperClient {
         )
         .await?;
         // Get local-only secret
-        let secret = self
+        let local_storage = self
             .handle_retrieve(&mut client_channel, key_id, RetrieveContext::LocalOnly)
-            .await?;
-        // Return secret as bytes
-        match secret {
-            RetrieveResult::None => Err(LockKeeperClientError::ExportFailed),
-            RetrieveResult::ArbitraryKey(local_storage) => Ok(local_storage.secret.into()),
-            RetrieveResult::SigningKey(_) => Err(LockKeeperClientError::InvalidKeyRetrieved),
-        }
+            .await?
+            .ok_or(LockKeeperClientError::ExportFailed)?;
+        Ok(local_storage.material.into())
     }
 
     /// Export signing key pair material from the key servers.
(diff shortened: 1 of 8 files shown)


lock-keeper-client/src/api.rs
```diff
@@ -28,20 +28,13 @@ use rand::{rngs::StdRng, SeedableRng};
...
-/// Ways of returning a key from the retrieval process based on usage
-/// [`RetrieveContext`]
 #[derive(Debug, Deserialize, Serialize)]
-pub enum RetrieveResult {
-    None,
-    ArbitraryKey(LocalStorage),
+#[serde(tag = "type")]
+pub enum LocalStorage {
+    ArbitraryKey(Secret),
     SigningKey(Export),
```
vidyaap, line 35: I don't think it will always be an `Export` that goes in the SigningKey variant, but we haven't really implemented retrieving signing keys yet, so this is here for now.

lock-keeper-client/src/api.rs
```diff
@@ -28,20 +28,13 @@ use rand::{rngs::StdRng, SeedableRng};
...
 use tracing::error;
 
-/// Ways of returning a key from the retrieval process based on usage
-/// [`RetrieveContext`]
 #[derive(Debug, Deserialize, Serialize)]
-pub enum RetrieveResult {
-    None,
-    ArbitraryKey(LocalStorage),
+#[serde(tag = "type")]
+pub enum LocalStorage {
```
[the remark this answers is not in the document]
vidyaap: Actually, generate does return a plain LocalStorage, but it was already specifying which variant anyway (between arbitrary and signing key).

Separately though, is that a problem with enums here in general? That we can't specify which variant can be returned from which function(s)?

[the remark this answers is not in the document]
vidyaap: Yep, so that's how it happens now; things only get wrapped in a `LocalStorage` in the client right before being returned to the caller.  But is that still something that could be tampered with? Also, not sure if it's an issue that now the return type isn't as informative to the caller in terms of what they'll get back?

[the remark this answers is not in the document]
picklenerd: We talked about this on a call and decided to try doing this with generics.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -65,6 +65,36 @@ impl TryFrom<PlaceholderEncryptedSigningKeyPair> for SigningKeyPair {
     }
 }
 
+/// Temporary type to represent a serializable
+/// [`SigningKeyPair`].
+///
+/// This is returned when retrieving a signing key (and not exporting).
+#[derive(Debug, Deserialize, Serialize)]
+pub struct PlaceholderSerializableSigningKeyPair {
+    signing_key: Vec<u8>,
```
vidyaap, line 74: Added this placeholder type for when we need serializable signing key pairs to put into local storage. This should go away when we actually implement the encryption needed for local storage.  I'm not sure we need it technically, since we don't retrieve any signing keys yet.

vidyaap (commented):
This took more changes than I thought... Specifically I had to make a new placeholder type (commented on below), and it makes it a little more annoying on the calling app side to store multiple `LocalStorage`s together (i.e. the lkic code). Do we like this better for clarity though?

lock-keeper/src/crypto/signing_key.rs
picklenerd: Are we actually going to have the ability to retrieve signing keys at some point? Maybe we should hold off on adding this until we need it since it's a temporary type.

picklenerd: I think that would be good. We can add them back later if we need them.

vidyaap: Updated!

picklenerd (approved):
LGTM!

PR #312, opened 2022-10-25
Renamed lkic to lock-keeper-client-cli

picklenerd:
I've been referring to this as "the CLI" in conversation. This PR updates the name of the interactive client crate to match this usage.

I went with `lock-keeper-client-cli` so that it's specific enough to say what it is. The `cargo make` command is `cli` for ease of use.

https://github.com/boltlabs-inc/key-mgmt/pull/311 should be merged before this PR.


gatoWololo (approved):
Looks good! This is a good change!

PR #325, opened 2022-10-28
Refactor e2e tests to use new testing harness #313

### LINKED ISSUE #313: Refactoring end-to-end tests to use new testing harness
There was a pretty nice testing harness introduced with the database tests. We can leverage this for the end-to-end tests.

- [ ] Remove the `Test` type
- [ ] Move all of the reusable functions in the `operations` submodules into a new module for helper functions. Remove state storage and `asserts`.
- [ ] Move all of  the non-reusable functions in the `operations` submodules out of the `impl Test` block. These will form the new test cases.
- [ ] Add test cases for any helper functions that also served as test cases.
- [ ] Call new test cases with the existing `run_parallel!` macro.

This ticket will also close #115.

vidyaap:
Closes #313 

Couple of comments/questions:
- @picklenerd I'm realizing this isn't exactly the way you outlined it in the issue - I kept the operations module as is, minus the Operation enum, and just added each operation's tests in their file. Let me know if you'd still prefer to move all the operation (non-test) functions to their own module instead of them importing from each other.
- There is a public "authenticate" function in the tests that I don't use anywhere except for in the authenticate tests, since the other operations already include auth as part of it. Technically this maintains the "authenticate per operation" policy, but would it be better to just use that public function once per test? Or is that maybe a future update once we have sessions?
- I could not figure out how to keep the filtering since I think it would have to go in the `run_parallel!` macro somehow? The filter option code is still there in case someone else knows how to do it, but if not I'll remove it.


picklenerd:
>  Let me know if you'd still prefer to move all the operation (non-test) functions to their own module instead of them importing from each other.

I do think the client wrapper and helper functions should be separate from the test cases. It's not too bad right now but I think the operation files will tend to cross pollinate over time and it will get messy. I think the test cases should be the only functions in their respective files so that they're easy to find and troubleshoot.

picklenerd:
> I could not figure out how to keep the filtering since I think it would have to go in the run_parallel! macro somehow? The filter option code is still there in case someone else knows how to do it, but if not I'll remove it.

Yeah I think it would have to go in the macro. We could compare the filter against the result of `stringify!($task)` which returns the function name. The macro would have to take the config object as a parameter in that case.

lock-keeper-tests/src/end_to_end/operations/authenticate.rs
```diff
@@ -1,16 +1,83 @@
...
+    )?;
+
+    Ok(result)
+}
+
+/// Authenticate to the LockKeeper key server and get a LockKeeperClient.
+pub async fn authenticate(
+    account_name: &AccountName,
+    password: &Password,
+    config: &Config,
```
picklenerd, line 31: It might be worth having a `TestState` type that holds these so that we don't have to pass them to all of the helper functions.

vidyaap:
Updated with separated test-cases and operations modules, and test state struct!

> The macro would have to take the config object as a parameter in that case.

Yeah that's the part I wasn't sure how to do, but I think I figured it out! But now I think I did the actual skipping wrong. I added it into the `run_test_case` function and usually it looks like it works, but sometimes I get an error like this one that makes me think that maybe the tests aren't being skipped at all?

```
thread 'tokio-runtime-worker' panicked at 'assertion failed: `(left == right)`
  left: `"tokio Sender error: channel closed"`,
 right: `"Invalid account"`', lock-keeper-tests/src/end_to_end/operations.rs:37:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'tokio-runtime-worker' panicked at 'assertion failed: `(left == right)`
  left: `"tokio Sender error: channel closed"`,
 right: `"Invalid account"`', lock-keeper-tests/src/end_to_end/operations.rs:37:5
thread 'tokio-runtime-worker' panicked at 'assertion failed: `(left == right)`
  left: `"tokio Sender error: channel closed"`,
 right: `"Invalid account"`', lock-keeper-tests/src/end_to_end/operations.rs:37:5
```

picklenerd (approved):
Vidya figured out how to fix the error so this looks good to me!

PR #331, opened 2022-10-31
Add decryption validation for encrypted signing keys

### LINKED ISSUE #239: Implement local method for creation + encryption of signing keys
In #234, we introduce methods to create and encrypt a signing key. These methods don't panic, but the signing key is just a unit type. 

- [x] Actually call the method to create a meaningful signing key (should be defined in #235)
- [x] Encrypt the signing key
- [x] Make sure decryption still works
- [x] Add some bad-behavior tests

note on the epic: this is actually only used in the local client version where signing keys are encrypted with the client-side storage key. Maybe we need another epic for that?

### LINKED ISSUE #241: Update create+encrypt signing key

marsella:
Closes #241.
Closes #239.

This PR adds validation of the associated data for signing keys encrypted under a storage key. Encryption under a storage key means they are generated client-side, either in a generate flow or in an import flow.

It also adds "some bad-behavior checks" as requested in #239, by testing decryption with a bunch of incorrect associated data.

See also: boltlabs-inc/key-mgmt-spec#144, which fixes the spec to work for either generated or imported keys.

### DIFF
--- a/lock-keeper-key-server/src/server/opaque_storage.rs
+++ b/lock-keeper-key-server/src/server/opaque_storage.rs
@@ -22,7 +22,7 @@ pub fn create_or_retrieve_server_key_opaque(
         Err(_) => {
             let server_setup = ServerSetup::<OpaqueCipherSuite>::new(rng);
             std::fs::create_dir_all(
-                &service
+                service
                     .opaque_server_key
                     .parent()
                     .ok_or(LockKeeperServerError::InvalidOpaqueDirectory)?,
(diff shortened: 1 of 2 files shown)


indomitableSwan (commented):
LGTM. I would like the Rust reviewer to be responsible for the final approval of this PR.

picklenerd (approved):
The Rust looks good to me! Cloning the identifying context threw me off at first but doing that any other way would be really messy and probably less performant.

PR #333, opened 2022-11-01
Check user session key at the beginning of requests #328

### LINKED ISSUE #328: Check that user has valid session key on the server
Now that we store users' session keys in a cache on the server side, we can confirm that users are authenticated in subsequent requests by checking that the server's session key cache contains a valid key for the user.

vidyaap:
Closes #328 

I also made a RequestMetadata type to avoid having to parse a bunch of types one by one from the tonic metadata.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -65,7 +65,7 @@ args = ["run", "--bin", "lock-keeper-tests", "--", "--test-type", "integration"]
 [tasks.all-tests]
 command = "cargo"
 args = ["run", "--bin", "lock-keeper-tests", "--", "--test-type", "all"]
-dependencies = ["ci-tests"]
+dependencies = ["ci-test"]
 
 # -- CI Tasks --
 [tasks.ci]
--- a/lock-keeper-key-server/src/server/session_key_cache.rs
+++ b/lock-keeper-key-server/src/server/session_key_cache.rs
@@ -3,6 +3,8 @@ use std::collections::hash_map::Entry;
 
 use std::collections::HashMap;
 
+use crate::{server::Context, LockKeeperServerError};
+use lock_keeper::types::operations::ClientAction;
 use std::time::{Duration, Instant};
 use thiserror::Error;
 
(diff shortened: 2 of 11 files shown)


Makefile.toml
```diff
@@ -84,7 +84,7 @@ args = ["clippy", "--all-features", "--all-targets", "--workspace", "--", "-Dwar
 command = "cargo"
 args = ["build", "--all-features", "--all-targets", "--workspace"]
 
-[tasks.ci-test]
+[tasks.ci-tests]
```
vidyaap, line 87: I think this name got lost somewhere? It's a dependency to "all-tests" but the dependency says "ci-tests" and the task was called "ci-tests".

lock-keeper-key-server/src/server/session_key_cache.rs
```diff
@@ -73,6 +75,21 @@ impl SessionKeyCache {
             Entry::Vacant(_) => Err(SessionKeyCacheError::MissingSessionKey),
         }
     }
+
+    pub fn check_key(&mut self, context: &Context) -> Result<(), LockKeeperServerError> {
+        let metadata = &context.metadata;
+        match metadata.action {
+            ClientAction::Authenticate | ClientAction::Register => Ok(()),
+            _ => {
```
vidyaap, line 83: Is it better to explicitly list all the other ClientAction variants here?

Makefile.toml
picklenerd: I actually changed this since this particular task calls `cargo test`. The other `ci-*` task names mirror the cargo command so I wanted to make it match. I think it would be better to update `all-tests` instead.

lock-keeper-key-server/src/server/session_key_cache.rs
picklenerd: I think this is easier to read. Maybe add a comment above authenticate and register to explain that these two are unauthenticated, and then a comment above the `_` saying that the rest are authenticated.

lock-keeper/src/types/operations.rs
```diff
@@ -33,3 +37,38 @@ pub enum ClientAction {
     RetrieveSigningKey,
     RetrieveStorageKey,
 }
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct RequestMetadata {
+    pub action: ClientAction,
```
picklenerd, line 43: Is there a reason this one is public and the others aren't?

picklenerd:
> I also made a RequestMetadata type 

I really like this!

lock-keeper/src/types/operations.rs
vidyaap: Oh thanks for commenting on this; I was going to come back to it because I didn't want to deal with comparing &ClientAction, but now I'm remembering that ClientAction derives Copy so I can make an accessor that returns by value.

picklenerd (approved):
LGTM!

PR #341, opened 2022-11-02
Refactor Client CLI (#332)

### LINKED ISSUE #332: Refactor Client CLI Code
The client CLI code has grown into a big match as we add more commands. There is considerable code duplication between commands. We want to refactor this.

We would also like to be able to reuse the existing client CLI code for a service provider CLI. This also motivates this refactor.

gatoWololo:
Closes #332 

This is ready for review!


lock-keeper-client-cli/src/commands.rs
```diff
@@ -0,0 +1,762 @@
...
+
+    /// Convert a string into a Box<dyn Command> suitable for dynamically
+    /// dispatching the execute function.
+    fn parse_to_command(s: &str) -> Result<DynCommand, anyhow::Error>
+    where
+        Self: Sized + 'static + Parse,
+    {
+        Self::from_str(s).map(|p| Box::new(p) as Box<dyn Command>)
+    }
+}
```
gatoWololo, line 62: @vidyaap This trait is doing some fancy stuff to allow for Trait object safety and dynamic dispatch. Lmk if you have question on how it works!

lock-keeper-client-cli/src/commands.rs
```diff
@@ -0,0 +1,762 @@
...
+impl<T: Command> Parse for T {
+    fn from_str(s: &str) -> Result<Self, anyhow::Error> {
+        let split: Vec<_> = s.trim().split(' ').collect();
+
+        match Self::from_slice(split.as_slice()) {
+            Some(t) => Ok(t),
+            None => bail!("Expected Format: {}", Self::format()),
+        }
+    }
+}
```
gatoWololo, line 124: Blanket implementation of `FromStr`. We use a new `Parse` trait since Rust does not allow blanket impl of traits defined in other crates </3

lock-keeper-client-cli/src/commands.rs
```diff
@@ -0,0 +1,762 @@
...
+
+struct RemoteSign {
+    name: String,
+    data: String,
+}
+struct Print {
+    name: String,
+}
+struct List {}
```
picklenerd, line 95: Is there a reason these don't implement `Debug` like the others?

lock-keeper-client-cli/src/commands.rs
```diff
@@ -0,0 +1,762 @@
...
+
+        match Self::from_slice(split.as_slice()) {
+            Some(t) => Ok(t),
+            None => bail!("Expected Format: {}", Self::format()),
+        }
+    }
+}
+
+#[async_trait]
+impl Command for Authenticate {
```
picklenerd, line 127: I think all of these are big enough to go in their own files.

lock-keeper-client-cli/src/commands.rs
```diff
@@ -0,0 +1,762 @@
...
+    where
+        Self: Sized;
+
+    /// Expected format for this command such as "register [account_name]
+    /// [password]". Used for generating help string dynamically.
+    fn format() -> &'static str
+    where
+        Self: Sized;
+
+    fn aliases() -> &'static str
```
picklenerd, line 33: We talked about this in the call but it would be nice if this returned a list so we don't have to duplicate the list between parsing and help text.

lock-keeper-client-cli/src/commands.rs
picklenerd: I never though to just make my own trait for this! The compiler error does not suggest that this would work at all IIRC.

picklenerd: You say </3, I say <3 for the solution 😆

lock-keeper-client-cli/src/state.rs
```diff
@@ -3,10 +3,12 @@
 
 use std::path::PathBuf;
 
-use crate::storage::Storage;
+use crate::storage::{Entry, Storage};
 use lock_keeper::{config::client::Config, types::database::user::AccountName};
 use lock_keeper_client::client::Password;
 
+trait CliCommand {}
```
picklenerd, line 10: Just flagging this in case you miss it.

lock-keeper-client-cli/src/commands.rs
gatoWololo: Haha I don't love the trait duplication.

lock-keeper-client-cli/src/commands.rs
gatoWololo: They can all impl debug!

lock-keeper-client-cli/Cargo.toml
```diff
@@ -18,7 +18,9 @@ tracing.workspace = true
 tracing-futures.workspace = true
 tracing-subscriber.workspace = true
 rand.workspace = true
+async-trait.workspace = true
 
 # Other dependencies
 anyhow = "1"
 hex = "0.4"
+nu-ansi-term = "0.46"
```
picklenerd, line 26: The `lock-keeper-tests` crate is using `colored` for this. It looks like they have pretty similar usage but we should pick one or the other.

lock-keeper-client-cli/src/cli_command/help.rs
```diff
@@ -0,0 +1,110 @@
...
+#[derive(Debug)]
+pub struct Help {
+    help_string: String,
+}
+
+impl Help {
+    /// Generate help string from all registered commands.
+    fn all_commands() -> String {
+        let mut help_string = String::new();
+        for c in get_functions::<HelpString>() {
```
picklenerd, line 18: What's `c` stand for? This could probably use a better name.

lock-keeper-client-cli/src/cli_command/help.rs
```diff
@@ -0,0 +1,110 @@
...
+    fn format() -> &'static str {
+        "help [command (Optional)]"
+    }
+
+    fn aliases() -> Vec<&'static str> {
+        vec!["help", "h"]
+    }
+
+    fn description() -> &'static str {
+        "Prints help text for a given command or type \"help all\" to list all commands."
```
picklenerd, line 74: My impression based on the code is that it's actually `help` by itself that lists all commands.

lock-keeper-client-cli/src/cli_command/mod.rs
```diff
@@ -0,0 +1,127 @@
+pub mod authenticate;
```
picklenerd, line 1: We've been using the `module_name.rs` pattern instead of the `mod.rs` pattern.

lock-keeper-client-cli/src/cli_command/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+        F::get_functions::<Retrieve>(),
+    ]
+}
+
+/// The GetCliCommandFunctions trait allows us to generate a vector of the same
+/// function for all our commands. See [get_functions] for use.
+pub trait GetCliCommandFunctions {
+    type Function;
+
+    fn get_functions<T: CliCommand + 'static>() -> Self::Function;
```
picklenerd, line 126:
```suggestion
    fn get_function<T: CliCommand + 'static>() -> Self::Function;
```

I think this should be singular since it only returns one function.

lock-keeper-client-cli/src/cli_command.rs
```diff
@@ -0,0 +1,149 @@
...
+pub use help::*;
+pub use import::*;
+pub use list::*;
+pub use logout::*;
+pub use print::*;
+pub use quit::*;
+pub use register::*;
+pub use remote_generate::*;
+pub use remote_sign::*;
+pub use retrieve::*;
```
picklenerd, line 29: These should be explicit.

picklenerd (approved):
LGTM!

Issue #307, opened 2022-10-20
Add server-side encrypted storage


marsella:
In the passive / remote generation flow and the import-only-to-server flow, the key server creates (or receives) keys on behalf of the client and puts them in secure storage. The parameters for server-side secure storage of keys are still relatively undefined (see boltlabs-inc/key-mgmt-spec#28), but the rudimentary guidance says
> All data stored by the server MUST be encrypted by a key known only to the server.

At time of writing, this requirement is not satisfied. When the server imports or generates keys, they are transformed into a `PlaceholderEncryptedSigningKeyPair`, which does not actually encrypt the data; it just copies it.

In lieu of completing boltlabs-inc/key-mgmt-spec#28, we should try to implement a simple solution that involves actual encryption. A reasonable first step is to reuse the generic AEAD encryption API we already built for client-side encryption. Rough steps:
1. Update the spec to document initial steps and encryption algorithm, and decide how to store (or derive) the server's encryption key.
1. Define a cryptography API. This should include a type for the server encryption key, including a generation function, and methods to encrypt + decrypt signing key pairs.
1. Update the server-side protocol implementations to use this encryption instead of the dummy `Placeholder`. Update the server setup code to generate the key, if appropriate.
1. Remove all traces of `PlaceholderEncryptedDSigningKeyPair`.

gijsvl:
Important note on this issue: `PlaceholderEncryptedSigningKeyPair` implements `ZeroizeOnDrop` as it contains the plain text key. As soon as it becomes an encrypted version of the key it doesn't need to implement that trait anymore.

PR #342, opened 2022-11-03
Use Zeroize to safely discard secrets from memory #108

gijsvl:
This should solve #108 partially, by zeroing out secrets after they have been used. To make sure they don't live on in unguarded memory.


lock-keeper/src/crypto/generic.rs
```diff
@@ -182,7 +184,10 @@ where
         enc_key: &EncryptionKey,
         object: T,
         associated_data: &AssociatedData,
-    ) -> Result<Encrypted<T>, CryptoError> {
+    ) -> Result<Encrypted<T>, CryptoError>
+    where
+        T: ZeroizeOnDrop,
```
marsella, line 189: I'm not sure we'll always want this condition -- right now we only use the AEAD encryption to encrypt keys under the storage key, but in the next phase or two, we'll also start using this underlying implementation to encrypt all messages sent between the two parties under the session key. Is it necessary to make sure the zeroize function is called?

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -79,17 +81,18 @@ pub struct SigningPublicKey(VerifyingKey);
 ///
 /// This can only be "decrypted" by the server.
 /// TODO #307: Replace this placeholder with actual encryption.
-#[derive(Clone, Debug, Deserialize, Serialize)]
+#[derive(Clone, Debug, Deserialize, Serialize, ZeroizeOnDrop)]
 pub struct PlaceholderEncryptedSigningKeyPair {
     signing_key: Vec<u8>,
```
marsella, line 86: suggestion: consider adding a note to the documentation or the issue #307 -- when this placeholder is replaced with a ciphertext, I don't think we need to zeroize it.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -286,17 +289,18 @@ impl Import {
 }
 
 /// Raw material for an exported signing key.
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, ZeroizeOnDrop)]
 pub struct Export {
```
marsella, line 293: The corresponding `Import` struct will also hold key material and should be zeroized.

marsella (commented):
Took a quick pass. I'm not as familiar with the libraries outside the crypto module, so I might have missed some wrapper struct that's relevant.

Do you plan to add documentation here or in the issue on what guarantees zeroize gets us? I want to make sure it's clear (both to us and to future security-conscious users).

lock-keeper/src/crypto/generic.rs
gijsvl: Not necessarily, but two things:
- The keys that get encrypted are copied into a bytearray to get encrypted, and we have no guarantee the bytearray will get zeroized.
- In general we encrypt a certain plaintext for a reason (because it's somewhat secret), which could justify zeroizing after use.

Another solution is providing another encrypt_and_zeroize funcion (or something similar), or a specific encrypt function for the specific keys.

gijsvl:
This PR is ready for review, but we will not merge this before this week's release.

lock-keeper-client-cli/src/state.rs
```diff
@@ -15,20 +15,20 @@ pub struct State {
     /// Local storage for key information
     pub storage: Storage,
     /// Contains the credentials of the currently logged-in user
-    pub credentials: Option<Credentials>,
+    pub client: Option<LockKeeperClient>,
```
picklenerd, line 18: Can you remove this change before you merge?

lock-keeper/src/crypto/generic.rs
marsella: Okay, we can leave for now. I think this will happen in the next release so we can see at that point whether it makes sense.

lock-keeper-client/src/client.rs
```diff
@@ -134,21 +136,18 @@ impl LockKeeperClient {
...
                     config: config.clone(),
                     tonic_client: client,
                     rng: Arc::new(Mutex::new(rng)),
                     account_name: account_name.clone(),
-                    user_id,
-                    export_key,
+                    user_id: auth_result.user_id.clone(),
+                    export_key: auth_result.export_key.clone(),
                 };
+                drop(auth_result);
```
picklenerd, line 150: Is it possible to call `auth_result.zeroize()` here? I feel like it makes the intent more explicit. I could see someone removing this drop in the future without realizing that it's intended to zeroize the struct.

lock-keeper/src/crypto/generic.rs
```diff
@@ -91,27 +92,28 @@ pub struct Encrypted<T> {
 }
 
 /// A well-formed symmetric encryption key for an AEAD scheme.
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, ZeroizeOnDrop)]
 pub(super) struct EncryptionKey {
-    key: chacha20poly1305::Key,
+    key: Box<chacha20poly1305::Key>,
```
picklenerd, line 97: What's this box for?

gijsvl: It's to make sure it actually gets zeroized and is not moved into the drop: https://benma.github.io/2020/10/16/rust-zeroize-move.html

lock-keeper-client/src/client.rs
gijsvl: I changed it

lock-keeper/src/crypto/generic.rs
picklenerd: Weird and totally logical!

picklenerd (approved):
LGTM

marsella (approved):
LGTM

PR #357, opened 2022-11-07
Create LockKeeperTestError type #337

### LINKED ISSUE #337: Create `LockKeeperTestsError` type and remove `anyhow`
`anyhow` is really convenient but it can be difficult to tell which errors are returned by the client and which errors are returned in testing code. The `lock_keeper_tests` crate should have a custom error type similar to the other crates. It should wrap any error types that pop up in the `lock_keeper_tests` crate including `LockKeeperClientError`.

- [ ] Create `LockKeeperTestsError` type.
- [ ] Remove `anyhow` dependency and all references in code.

vidyaap:
Closes #337 

Is there also a tokio error I can catch from try_join? I couldn't seem to find what it throws.

### DIFF
--- a/lock-keeper-tests/src/error.rs
+++ b/lock-keeper-tests/src/error.rs
@@ -1,0 +1,19 @@
+use thiserror::Error;
+
+pub type Result<T> = std::result::Result<T, LockKeeperTestError>;
+
+#[derive(Debug, Error)]
+pub enum LockKeeperTestError {
+    #[error("Invalid test type: {0}")]
+    InvalidTestType(String),
+    #[error("LockKeeperError: {0:?}")]
+    LockKeeper(#[from] lock_keeper::LockKeeperError),
+    #[error("LockKeeperClientError: {0:?}")]
+    LockKeeperClient(#[from] lock_keeper_client::LockKeeperClientError),
+    #[error("LockKeeperServerError: {0:?}")]
+    LockKeeperServer(#[from] lock_keeper_key_server::LockKeeperServerError),
+    #[error("MongoDB error: {0:?}")]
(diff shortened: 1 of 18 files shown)


picklenerd:
> Is there also a tokio error I can catch from try_join? I couldn't seem to find what it throws.

I'm not positive but I believe that it returns the error of the functions/futures you pass to it. I think its purpose is for fallible futures.

lock-keeper-tests/src/error.rs
```diff
@@ -0,0 +1,19 @@
...
+    #[error("LockKeeperError: {0:?}")]
+    LockKeeperError(#[from] lock_keeper::LockKeeperError),
+    #[error("LockKeeperClientError: {0:?}")]
+    LockKeeperClientError(#[from] lock_keeper_client::LockKeeperClientError),
+    #[error("LockKeeperServerError: {0:?}")]
+    LockKeeperServerError(#[from] lock_keeper_key_server::LockKeeperServerError),
+    #[error("Invalid test type: {0}")]
+    InvalidTestType(String),
+    #[error("RandError: {0:?}")]
+    RandError(#[from] rand::Error),
```
picklenerd, line 18: I usually prefer to leave the word `Error` out of variants since they're normally written as `LockKeeperTestError::Rand` so it's kind of redundant. This is totally a stylistic choice though.

vidyaap: Decided to change it! But I left "error" in the strings since it's just the inner string that prints in the `{0:?}`

picklenerd: I like it in the strings since that's the name of the inner type.

picklenerd (approved):

PR #359, opened 2022-11-09
Create a DataStore abstraction over database operations #338

### LINKED ISSUE #338: Create a `DataStore` trait to abstract over database operations
Our key server shouldn't make assumptions about the underlying database implementation. It should define an interface that database operations conform to and allow the user to implement that interface with a database of their choosing. We should still include a reference implementation for our preferred database that we continue to use internally.

- [x] Create a trait with methods for all existing database functions.
- [x] Documentation for this trait should highlight that it is not complete and that new trait methods can be added in new versions.
- [x] Implement the trait for MongoDB in a separate crate and update existing code to use this implementation.
    - [ ] It maybe be worth separating our workspace at this point. We have a set of core crates and a set of other types of crates.

vidyaap:
Closes #338 

Couple of things I'm not sure about, that I'll leave inline comments for. Also moved the key server binary into a `bin` folder separate from the key server library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Integration tests are separated into two categories, end-to-end tests and genera
 
 Start the server running in the background. This will compile the project from scratch the first time you run it so it will take a while. It should be faster for future runs.
 ```bash
-cargo make start
+cargo make start-all
 ```
 
 To run the end-to-end tests:
--- a/databases/lock-keeper-mongodb/Cargo.toml
+++ b/databases/lock-keeper-mongodb/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "lock-keeper-mongodb"
+version.workspace = true
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+# Git dependencies
+lock-keeper = { version = "0.2.0", path = "../../lock-keeper" }
+lock-keeper-key-server = { version = "0.2.0", path = "../../lock-keeper-key-server" }
+
+# Workspace dependencies
+async-trait.workspace = true
+futures.workspace = true
+opaque-ke.workspace = true
+thiserror.workspace = true
+
+# Other dependencies
+mongodb = "2.3.0"
--- a/databases/lock-keeper-mongodb/src/error.rs
+++ b/databases/lock-keeper-mongodb/src/error.rs
@@ -1,0 +1,14 @@
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("Invalid account")]
+    InvalidAccount,
+    #[error("Key ID not found")]
+    KeyNotFound,
+    // Wrapped errors
+    #[error(transparent)]
+    Bson(#[from] mongodb::bson::ser::Error),
+    #[error(transparent)]
+    MongoDb(#[from] mongodb::error::Error),
+}
--- a/lock-keeper-key-server/src/operations/authenticate.rs
+++ b/lock-keeper-key-server/src/operations/authenticate.rs
@@ -3,6 +3,7 @@ use crate::{
     server::{Context, Operation},
 };
 
+use crate::database::DataStore;
 use async_trait::async_trait;
 use lock_keeper::{
     config::opaque::OpaqueCipherSuite,
@@ -24,11 +25,11 @@ struct AuthenticateStartResult {
 pub struct Authenticate;
 
 #[async_trait]
-impl Operation for Authenticate {
+impl<DB: DataStore> Operation<DB> for Authenticate {
     async fn operation(
         self,
         channel: &mut ServerChannel,
-        context: &mut Context,
+        context: &mut Context<DB>,
     ) -> Result<(), LockKeeperServerError> {
         let result = authenticate_start(channel, context).await?;
         let session_key = authenticate_finish(channel, result.login_start_result).await?;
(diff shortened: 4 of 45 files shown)


databases/lock-keeper-mongodb/src/error.rs
```diff
@@ -0,0 +1,27 @@
...
+    #[error("Invalid account")]
+    InvalidAccount,
+    #[error("Key ID not found")]
+    KeyNotFound,
+    // Wrapped errors
+    #[error(transparent)]
+    Bson(#[from] mongodb::bson::ser::Error),
+    #[error(transparent)]
+    MongoDb(#[from] mongodb::error::Error),
+}
```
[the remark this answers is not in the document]
picklenerd: I can't think of any other way to handle this. One thing we could do is have the `DataStore` trait contain an associated error type and then define a conversion from that error type to store in an error variant like `DatabaseError(Box<dyn std::Error)`

databases/lock-keeper-mongodb/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[dependencies]
+# Git dependencies
+lock-keeper = { version = "0.2.0", path = "../../lock-keeper" }
+lock-keeper-key-server = { version = "0.2.0", path = "../../lock-keeper-key-server" }
+
+# Workspace dependencies
+async-trait.workspace = true
+futures.workspace = true
+mongodb.workspace = true
```
picklenerd, line 20: Can this be moved out of the workspace and into this crate?

lock-keeper-key-server/src/operations/authenticate.rs
```diff
@@ -57,11 +58,15 @@ async fn authenticate_start(
...
-    let (server_registration, user_id) =
-        match context.db.find_user(&start_message.account_name).await? {
-            Some(user) => user.into_parts(),
-            None => return Err(LockKeeperServerError::InvalidAccount),
-        };
+    let (server_registration, user_id) = match context
+        .db
+        .find_user(&start_message.account_name)
+        .await
+        .map_err(|e| LockKeeperServerError::Database(Box::new(e)))?
```
picklenerd, line 65: We can't do a blanket `From` implementation for this but we could add a method to `LockKeeperServerError` to make these shorter:

```rust
impl LockKeeperServerError {
    pub fn database(error: impl std::error::Error + Send + Sync + 'static) -> Self {
        Self::Database(Box::new(error))
    }
}
```

Then this line would be:

```rust
.map_err(LockKeeperServerError::database)?
```

lock-keeper-tests/src/end_to_end/test_cases/export.rs
```diff
@@ -55,7 +55,7 @@ async fn cannot_export_fake_key(config: Config) -> Result<()> {
 
     let fake_key_id = generate_fake_key_id(&state).await?;
     let bytes_res = export(&state, &fake_key_id).await;
-    compare_errors(bytes_res, LockKeeperClientError::InvalidAccount);
+    compare_errors(bytes_res, Status::internal("Internal server error"));
```
picklenerd, line 58: #366

lock-keeper-key-server/src/server/service.rs
```diff
@@ -16,9 +16,11 @@ use tonic::transport::{server::Routes, Server};
 use tracing::{error, info};
 
 /// Starts a full Lock Keeper server stack based on the given config.
-pub async fn start_lock_keeper_server(config: Config) -> Result<(), LockKeeperServerError> {
+pub async fn start_lock_keeper_server<DB: DataStore + Clone + 'static>(
+    config: Config,
+    db: DB,
+) -> Result<(), LockKeeperServerError> {
```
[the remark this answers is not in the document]
picklenerd: Yeah I think this has to be generic since you need to pass the DB.

lock-keeper-key-server/src/server/service.rs
```diff
@@ -16,14 +16,17 @@ use tonic::transport::{server::Routes, Server};
...
     tracing::info!("Starting Lock Keeper key server");
-    let db = Database::connect(&config.database).await?;
+    let db = Arc::new(db);
     // Collect the futures for the result of running each specified server
     let mut server_futures: FuturesUnordered<_> = config
         .services
         .iter()
-        .map(|service| start_service(service, &config, &db))
+        .map(|service| start_service(service, &config, db.clone()))
```
picklenerd, line 29: I don't think you need this clone anymore since there's only one service now. If you rebase, this code will be a little bit different.

lock-keeper-tests/src/end_to_end/test_cases/export.rs
vidyaap: 🙏

README.md
```diff
@@ -68,7 +68,7 @@ Integration tests are separated into two categories, end-to-end tests and genera
 
 Start the server running in the background. This will compile the project from scratch the first time you run it so it will take a while. It should be faster for future runs.
 ```bash
-cargo make start
+cargo make start-all
```
vidyaap, line 71: Running `cargo make all-tests` requires start-all, but I think the rest don't. Would you rather I added this in that section instead of saying to run start-all all the time?

picklenerd: Maybe it would be better to remove `start-all` entirely and have `start` point to `start-server-detached-all`. Running both servers seems like the most common case and I haven't seen any downside so far.

picklenerd: I made a ticket for broader makefile cleanup #368

picklenerd (commented):
LGTM. @gatoWololo can you double-check this? I want to make sure it captures everything you had in mind during all of our discussions.

README.md
vidyaap: Cool, I'll still leave this in the readme until then since we do need it for all-tests.

lock-keeper-key-server/src/operations/authenticate.rs
gatoWololo: Ugh, I tried but couldn't get this blanket impl working </3
```rust
impl<E, DB> From<E> for LockKeeperServerError
where
    DB: DataStore<Error = E>,
{
    fn from(e: <DB as DataStore>::Error) -> Self {
        LockKeeperServerError::Database(Box::new(e))
    }
}
```
So I agree with Jordan.

gatoWololo:
Code looks good to me! I'm pretty sure originally I said to not put the `'static` bound on the trait definition. But I take this back. My one suggestion would be to do:
```diff
- pub trait DataStore: Send + Sync {
+ pub trait DataStore: Send + Sync + 'static {
```
And this way, we don't need to do `DB: DataStore + 'static` everywhere.

picklenerd (approved):

PR #362, opened 2022-11-15
Add LockKeeperResponse type to client api #329

### LINKED ISSUE #329: Create a `LockKeeperResponse` type to return from the client
Client API functions currently return their result types directly. For example, the `export_signing_key` function returns a `Result<Export, LockKeeperClientError>`, so you call the function and you expect to get an `Export` back. 

In a traditional client/server system, the client gets network responses back from a server. These network responses can include headers with useful metadata about the request or response. Since our client abstracts over the network requests, we don't have a way to include useful metadata.

One piece of metadata that would immediately be useful is an ID assigned to each request. This ID could be used to filter audit events or during troubleshooting to find server logs associated with a particular request. Being able to filter audit logs by request should make it easy to solve #317 where we're bumping up against our assumptions about the ordering of audit events.

The proposed type would look something like this:

```rust
pub struct LockKeeperResponse<T> {
    pub data: T,
    pub request_id: RequestId,
    // Any future metadata
}
```

Metadata could also go into its own type just to keep things clean.

The return signature for `export_signing_key` would be updated to `Result<LockKeeperResponse<Export>, LockKeeperClientError>`. This is ugly, but we could define a type alias to make it nicer:

```rust
pub type LockKeeperResult<T> = Result<LockKeeperResponse<T>, LockKeeperClientError>
```
This would give us a return signature of `LockKeeperResult<Export>`.

One big benefit of this approach is that it would allow us to add more metadata in the future without breaking the return types of all of our API functions.

picklenerd:
This PR implements the `LockKeeperResponse` type intended to hold results and metadata. The design is a work-in-progress.

Closes #329


lock-keeper/src/error.rs
```diff
@@ -35,6 +37,8 @@ pub enum LockKeeperError {
     #[error("tokio Sender error: {}", .0)]
     TokioSender(String),
     #[error(transparent)]
+    TonicMetadata(#[from] tonic::metadata::errors::InvalidMetadataValueBytes),
```
vidyaap, line 40: Since this is here now, can we remove this variant from the client errors?

lock-keeper-key-server/src/server.rs
```diff
@@ -42,41 +40,21 @@ impl LockKeeperKeyServer {
...
     }
 
-    pub(crate) fn context(
-        &self,
-        request: &Request<tonic::Streaming<Message>>,
-    ) -> Result<Context, LockKeeperServerError> {
-        // Parse RequestMetadata
-        let metadata = Self::parse_metadata(request)?;
-
+    pub(crate) fn context(&self) -> Result<Context, LockKeeperServerError> {
```
vidyaap, line 43: Does this still have to return a result?

lock-keeper-client/src/error.rs
```diff
@@ -2,6 +2,8 @@ use lock_keeper::LockKeeperError;
 use thiserror::Error;
 use tonic::{Code, Status};
 
+pub type Result<T> = std::result::Result<T, LockKeeperClientError>;
```
vidyaap, line 5: Could we use this result type for the API functions too?

lock-keeper-client/src/api.rs
```diff
@@ -120,7 +134,13 @@ impl LockKeeperClient {
             .handle_retrieve(&mut client_channel, key_id, RetrieveContext::LocalOnly)
             .await?
             .ok_or(LockKeeperClientError::ExportFailed)?;
-        Ok(local_storage.material.into())
+
+        let response: Vec<u8> = local_storage.material.into();
+
+        Ok(LockKeeperResponse {
+            data: response,
+            metadata: client_channel.metadata,
```
vidyaap, line 142: Does it matter that the request ID being in the channel makes it mutable usually? I'm not sure if it's possible to mess with it, but just curious.

lock-keeper-client/src/error.rs
picklenerd: The ones coming out of the server?

We could but it would have to go into the protobuf definition. I like leveraging the metadata on the tonic responses for these since it keeps the API types simple, but that does hide everything else that gets returned through the metadata so I can see a strong argument against it.

Thoughts @gatoWololo?

lock-keeper-client/src/api.rs
picklenerd: That does kind bug me. I'm experimenting with a `metadata(&self) -> &M` method right now but it's causing problems with threads. If I can make it work, I think this is a better solution since there's no good reason to modify the metadata after it's created.

lock-keeper-client/src/error.rs
vidyaap: Oh I just meant the client library API functions!

picklenerd: I kinda like that it keeps the noise out of the implementation details.

That makes me think though, I could probably add a function  like `generate_response<T>(T, Channel)` that handles the noise of turning a T into a response. If we feel comfortable consuming the channel in the handlers, it wouldn't be too bad to just call it from there.

gatoWololo: Yeah, I agree with Vidya. Just to simplify the client API signatures a little bit.

picklenerd: Ok added this to the handler. The API functions for the exports are still a little messy since they do some extra processing but the rest are nice and simple.

lock-keeper-client/src/api.rs
picklenerd: Still trying to get this to work. Such a weird thing to cause threading issues...

gatoWololo: @picklenerd huh, lmk if I can give any suggestions on this threading issue you're dealing with.

picklenerd: I pushed what I have if you want to see the errors.

picklenerd: Pushed our fixes!

lock-keeper-key-server/src/server.rs
```diff
@@ -103,7 +81,7 @@ impl<DB: DataStore> LockKeeperRpc for LockKeeperKeyServer<DB> {
         &self,
         request: Request<tonic::Streaming<Message>>,
     ) -> Result<Response<Self::RegisterStream>, Status> {
-        let context = self.context(&request)?;
+        let context = self.context();
```
gatoWololo, line 84: Could we just inline this `context` local variable for all these functions?

gatoWololo (approved):
Looks good. Just pointed out one nit!

PR #363, opened 2022-11-15
Remove export_key as a type and cache master_key instead #356

### LINKED ISSUE #356: Update Export/Master key API
The current cryptography API assumes that a client will maintain access to their export key across the lifetime of their session, and use it to re-derive the master key whenever necessary.

The correct behavior is to derive the master key as soon as possible (e.g. during registration/authentication), then drop the export key. The master key should be the key held across the lifetime of a session.

- [x] Maybe take out the `OpaqueExportKey` type entirely and have a function that goes straight from opaque_ke export key type to `MasterKey`.
- [x] Update `decrypt_storage_key` function to take a master key instead of an export key
- [x] Move the `create_and_encrypt_storage_key` method to be on `MasterKey` instead of `OpaqueExportKey`
- [x] Update the `LockKeeperClient` to hold a `MasterKey` instead of export key and update client methods (side question: why does `handle_create_storage_key` not take `&self`?)
- [x] Revisit database tests that involve export keys

Notes for the future: Currently, we use the `MasterKey` directly for encrypting the storage key. If we have more client-side cryptographic operations in the future that require key material, we'll want to apply a KDF to `MasterKey` with some relevant domain separation to create separate keys for each application.

gijsvl:
Closes #356

### DIFF
--- a/lock-keeper-client/src/client.rs
+++ b/lock-keeper-client/src/client.rs
@@ -64,9 +64,9 @@ pub struct LockKeeperClient {
     config: Config,
     account_name: AccountName,
     user_id: UserId,
+    master_key: MasterKey,
     tonic_client: LockKeeperRpcClient<LockKeeperRpcClientInner>,
     pub(crate) rng: Arc<Mutex<StdRng>>,
-    pub(crate) master_key: MasterKey,
 }
 
 /// Connection type used by `LockKeeperRpcClient`.
@@ -231,7 +231,7 @@ impl LockKeeperClient {
         // Decrypt storage_key
         let storage_key = response
             .ciphertext
-            .decrypt_storage_key(self.master_key.clone(), self.user_id())?;
+            .decrypt_storage_key(self.master_key().clone(), self.user_id())?;
         Ok(storage_key)
     }
 }
--- a/lock-keeper-client/src/client.rs
+++ b/lock-keeper-client/src/client.rs
@@ -7,7 +7,7 @@ use hyper::client::HttpConnector;
 use hyper_rustls::HttpsConnector;
 use lock_keeper::{
     constants::METADATA,
-    crypto::{OpaqueExportKey, OpaqueSessionKey, StorageKey},
+    crypto::{MasterKey, OpaqueSessionKey, StorageKey},
     infrastructure::channel::ClientChannel,
     rpc::lock_keeper_rpc_client::LockKeeperRpcClient,
     types::{
@@ -66,7 +66,7 @@ pub struct LockKeeperClient {
     user_id: UserId,
     tonic_client: LockKeeperRpcClient<LockKeeperRpcClientInner>,
     pub(crate) rng: Arc<Mutex<StdRng>>,
-    pub(crate) export_key: OpaqueExportKey,
+    pub(crate) master_key: MasterKey,
 }
 
 /// Connection type used by `LockKeeperRpcClient`.
@@ -79,7 +79,7 @@ type LockKeeperRpcClientInner = hyper::Client<
 
 pub(crate) struct AuthenticateResult {
     pub(crate) session_key: OpaqueSessionKey,
-    pub(crate) export_key: OpaqueExportKey,
+    pub(crate) master_key: MasterKey,
     pub(crate) user_id: UserId,
 }
 
@@ -93,6 +93,11 @@ impl LockKeeperClient {
     /// Get [`AccountName`] for the authenticated client.
     pub fn account_name(&self) -> &AccountName {
         &self.account_name
+    }
+
+    /// Get [`MasterKey`] for the authenticated client.
+    pub fn master_key(&self) -> &MasterKey {
+        &self.master_key
     }
 
     pub(crate) fn tonic_client(&self) -> LockKeeperRpcClient<LockKeeperRpcClientInner> {
@@ -136,7 +141,7 @@ impl LockKeeperClient {
         match result {
             Ok(AuthenticateResult {
                 session_key,
-                export_key,
+                master_key,
                 user_id,
             }) => {
                 // TODO #186: receive User ID over authenticated channel (under session_key)
@@ -147,7 +152,7 @@ impl LockKeeperClient {
                     rng: Arc::new(Mutex::new(rng)),
                     account_name: account_name.clone(),
                     user_id,
-                    export_key,
+                    master_key,
                 };
                 Ok(client)
             }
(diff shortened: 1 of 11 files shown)


gijsvl:
Note: this is a breaking change for existing storage keys in a database (the encryption/decryption key changed to take domain separation into account and not using master key directly). I'm not sure if we already want to account for such breaking changes. Given that we haven't got anything in production, I would say we don't necessarily care, but just putting this note here to be sure.

lock-keeper/src/crypto.rs
```diff
@@ -130,7 +112,29 @@ impl MasterKey {
             .with_bytes(user_id.clone())
             .with_str(StorageKey::domain_separator());
 
-        Encrypted::encrypt(rng, &self.0, storage_key, &associated_data)
+        let key = self.derive_key(associated_data.clone())?;
+        Encrypted::encrypt(rng, &key, storage_key, &associated_data)
+    }
+
+    /// Derive a new key from [`MasterKey`] using the domain separator in
+    /// [`AssociatedData`]
```
marsella, line 120: A couple nits:
1. This is "using the `AssociatedData` as a domain separator" -- the current phrasing is kind of confusing to me, like there's a domain sep field in AD.
2. Add a line about what the derived key should be used for.
3. Add terminating punctuation.

marsella (commented):
A couple suggestions, mostly around improving documentation.

Please update boltlabs-inc/key-mgmt-spec#146 to be more specific about the domain separation change you made here before you merge this.

lock-keeper/src/crypto.rs
```diff
@@ -42,29 +43,27 @@ impl From<GenericArray<u8, U64>> for OpaqueSessionKey {
...
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct OpaqueExportKey(Box<[u8; 64]>);
+pub struct MasterKey(EncryptionKey);
 
-impl OpaqueExportKey {
+impl MasterKey {
     /// Derive a uniformly distributed secret [`MasterKey`] using the export key
     /// as input key material.
-    fn derive_master_key(&self) -> Result<MasterKey, CryptoError> {
+    pub fn derive_master_key(export_key: Output<Sha512>) -> Result<Self, CryptoError> {
```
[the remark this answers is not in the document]
gijsvl: Added extra documentation

lock-keeper/src/crypto/generic.rs
```diff
@@ -115,6 +115,10 @@ impl EncryptionKey {
             context,
         }
     }
+
+    pub fn into_bytes(self) -> [u8; 32] {
```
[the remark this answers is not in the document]
gijsvl: Made it `pub(super)` and added a warning.

lock-keeper/src/crypto.rs
gijsvl: Rephrased and added.

gijsvl:
Ready for re-review, I've added some extra explanation to the key-mgmt-spec issue, let me know if that is insufficient.

marsella (commented):
The updated issue looks good to me and so do your changes. Thanks!

I approve on the cryptography side. Didn't really look at this Rust-wise, so I would recommend another reviewer take a pass on that.

lock-keeper/src/crypto.rs
```diff
@@ -86,40 +90,23 @@ impl OpaqueExportKey {
...
-    /// 1. Derive a master key from the [`OpaqueExportKey`]
-    /// 2. Generate a new [`StorageKey`] to encrypt stored data with
-    /// 3. Encrypt the storage key under the master key, using an AEAD scheme
-    /// 4. Return the encrypted storage key
+    /// 1. Generate a new [`StorageKey`] to encrypt stored data with
+    /// 2. Derive the decryption key from the master key,
+    ///    using the associated data
+    /// 2. Encrypt the storage key under the encryption key,
+    ///    using an AEAD scheme
+    /// 3. Return the encrypted storage key
```
vidyaap, line 98: nit: The numbering has an extra "2" here!

lock-keeper/src/crypto.rs
```diff
@@ -42,29 +43,32 @@ impl From<GenericArray<u8, U64>> for OpaqueSessionKey {
...
+impl MasterKey {
     /// Derive a uniformly distributed secret [`MasterKey`] using the export key
     /// as input key material.
-    fn derive_master_key(&self) -> Result<MasterKey, CryptoError> {
+    ///
+    /// # Arguments
+    ///
+    /// * `export_key` - the export_key as returned by opaque-ke library,
+    /// which has type [`Output<Sha512>`]
+    pub fn derive_master_key(export_key: Output<Sha512>) -> Result<Self, CryptoError> {
```
vidyaap, line 64: We've been defaulting to returning `LockKeeperError` from public crypto functions since it has a variant for wrapping `CryptoError` already. Would you be able to update this signature too? This should also make it so we don't have to add an additional error variant in the test errors.

lock-keeper-client/src/client.rs
```diff
@@ -95,6 +95,11 @@ impl LockKeeperClient {
         &self.account_name
     }
 
+    /// Get [`MasterKey`] for the authenticated client.
+    pub fn master_key(&self) -> &MasterKey {
+        &self.master_key
+    }
```
vidyaap, line 101: I think the only function that uses the master key is in the same file - would we be able to use the field directly instead of including this accessor?

vidyaap (commented):
Overall looks good to me! Just left a few minor comments. I'm also wondering if it would be worth making a small note in the README about the breaking change you mentioned until we have a specific place where we're tracking breaking changes. (maybe @jwantz has thoughts on that part?)

lock-keeper/src/crypto.rs
gijsvl: Done, and removed the extra error from the test errors.

lock-keeper-client/src/client.rs
gijsvl: Good catch!

jwantz:
> Overall looks good to me! Just left a few minor comments. I'm also wondering if it would be worth making a small note in the README about the breaking change you mentioned until we have a specific place where we're tracking breaking changes. (maybe @jwantz has thoughts on that part?)

When we publish our next release, I think we will need to point out the breaking change in our release notes. For now, I agree with Vidya, that we can add a note to the readme that v0.3 will not be backwards compatible to v0.2 for storage keys.

jwantz:
> Note: this is a breaking change for existing storage keys in a database (the encryption/decryption key changed to take domain separation into account and not using master key directly). I'm not sure if we already want to account for such breaking changes. Given that we haven't got anything in production, I would say we don't necessarily care, but just putting this note here to be sure.

I agree that we are in the right window of time to make a breaking change. We are at the start of development for the v0.3 release, which gives us plenty of time for internal testing. Moreover, we have not deployed to production yet.

vidyaap (approved):
Thanks for updating; this looks good to me!

Issue #339, opened 2022-11-01
Define a better database schema for stored secrets


picklenerd:
The current database model for user secrets is a little messy and unsustainable. We need a better design that accounts for multiple different key types.

The simplest solution would be to store key material as a blob of bytes along with a field that indicates the type of those bytes. In MongoDB, these could just be two fields in the database model for secrets. In a relational database, these would probably go in their own table as separate columns.

This approach would require some extra error handling in the server to ensure that the content of the bytes actually matches the associated key type. It might be difficult to enforce in some situation. For example, if two types of keys are both stores as 32 arbitrary bytes, how do you know if the key type is wrong? We may need to encode the key type in the bytes alongside the key material so that we can check it.

If we end up encoding the key type in the bytes, we should consider whether we need an extra field to determine the key type. The extra field would be queryable without decoding the bytes so it may prove useful.

marsella:
> This approach would require some extra error handling in the server to ensure that the content of the bytes actually matches the associated key type.

The `From<Vec<u8>>` method implemented for each encrypt-able type should implement all the relevant validation we can do for that type. As you say, though, I think we'll need to know what the _expected_ key type is so we can call the correct validation.

PR #371, opened 2022-11-18
Newtype definitions for API return types WIP

### LINKED ISSUE #336: Provide newtype definitions for all input/output types in public API
There are a couple places where we accept or return a `Vec<u8>` in our public API. These should be given newtype wrappers that describe what the bytes represent. `KeyId` and `SignableBytes` are examples of this pattern.

- [ ] `export_key` returns a `Vec<u8>`.
- [ ] `import_signing_key` accepts a `Vec<u8>` as its `key_material` parameter.

vidyaap:
Closes #336 

Question about how to handle the Export type: I'm trying to get both `export` and `export_signing_key` to return an "Export" but converting those to bytes for comparison in the tests is looking hard to do without knowing the type that it started out as. Is it worth changing this type to a generic, the same way as LocalStorage was handled?

(the export e2e test is commented out because of this question)

### DIFF
--- a/lock-keeper/src/crypto/arbitrary_secret.rs
+++ b/lock-keeper/src/crypto/arbitrary_secret.rs
@@ -9,7 +9,7 @@ use zeroize::ZeroizeOnDrop;
 
 use crate::crypto::{
     generic::{self, AssociatedData, CryptoError},
-    Encrypted, KeyId, StorageKey,
+    Encrypted, Export, KeyId, StorageKey,
 };
 
 /// An arbitrary secret.
@@ -104,9 +104,27 @@ impl Secret {
     ///
     /// This is only used in testing right now, but it would be fine to make it
     /// public.
-    #[cfg(test)]
     fn context(&self) -> &AssociatedData {
         self.0.context()
+    }
+}
+
+impl From<Secret> for Export {
+    fn from(secret: Secret) -> Self {
+        Self {
+            key_material: secret.0.get_material().into(),
+            context: secret.context().clone().into(),
+        }
+    }
+}
+
+impl TryFrom<Export> for Secret {
+    type Error = LockKeeperError;
+
+    fn try_from(export: Export) -> Result<Self, Self::Error> {
+        let context = export.context.clone().try_into()?;
+        let inner = generic::Secret::from_parts(&export.key_material, &context);
+        Ok(Secret(inner))
     }
 }
 
--- a/lock-keeper-client-cli/src/cli_command/import.rs
+++ b/lock-keeper-client-cli/src/cli_command/import.rs
@@ -1,6 +1,7 @@
 use crate::{cli_command::CliCommand, state::State};
 use anyhow::Error;
 use async_trait::async_trait;
+use lock_keeper::crypto::Import as LkImport;
 use lock_keeper_client::LockKeeperClient;
 use rand::Rng;
 
@@ -24,8 +25,9 @@ impl CliCommand for Import {
         .into_inner();
 
         let random_bytes = rand::thread_rng().gen::<[u8; 32]>().to_vec();
+        let import = LkImport::new(random_bytes)?;
         let key_id = lock_keeper_client
-            .import_signing_key(random_bytes)
+            .import_signing_key(import)
             .await
             .map_err(|e| anyhow::anyhow!("Failed to import signing key. Error: {:?}", e))?
             .into_inner();
--- a/lock-keeper-client/src/api.rs
+++ b/lock-keeper-client/src/api.rs
@@ -19,7 +19,7 @@ use crate::{
     client::Password, config::Config, LockKeeperClient, LockKeeperClientError, LockKeeperResponse,
 };
 use lock_keeper::{
-    crypto::{Export, KeyId, Secret, Signable, Signature},
+    crypto::{Export, Import, KeyId, Secret, Signable, Signature},
     types::{
         audit_event::{AuditEvent, AuditEventOptions, EventType},
         database::user::AccountName,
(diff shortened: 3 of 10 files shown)


picklenerd:
> Question about how to handle the Export type: I'm trying to get both `export` and `export_signing_key` to return an "Export" but converting those to bytes for comparison in the tests is looking hard to do without knowing the type that it started out as. Is it worth changing this type to a generic, the same way as LocalStorage was handled?

This is a really good question that could tie in closely with the schema design issue.

In the database, we're going to have a bunch of encrypted blobs for everything that we store, like @gijsvl mentioned in the estimation meeting today. This makes it difficult to know the type of the data. Our current solution is to have different arrays for each key type, but we've already seen that this isn't scalable. The easy solution is to add a field/column in the database that tells us the type of the data.

For exported keys coming out of the client API, this isn't an issue because we can sort out the type of the data and return the correct Rust type. So this would be the generic type similar to the `LocalStorage` type. Some struct that has key data of some type along with some generic data like the context. 

If the API consumer wants to store this data, they're going to have to do the same process in reverse. Take the generic typed data and convert it into bytes along with some kind of tag or other scheme to record its type.

Instead of returning fully typed keys, we could use a single type for all key data similar to the how the `Export` type is now but with an additional field to specify the key type. Then we have a straight shot from the database, to the client, and then to the API consumer. For cases like remote signing, we can check the key type and then turn the bytes into the appropriate Rust type. For everything else we have, we don't really care what's in the blob and can just pass it along.

The consumer of our client obviously loses type safety here so this might be a bad idea. I'm not convinced that the types matter much in this case.

vidyaap:
This makes sense. I did consider the additional field for the key type in Export, but along with the type safety, I wasn’t sure if it was worth having to create something (an enum I guess? unless you meant serde tags) to track the different key types. I do agree it would be nice to not have duplicate functions just because the keys types are different.

I should say though, the specific problem I’m having here is a bit more logistical - the export type separates the key material bytes from the context bytes and local storage doesn’t so when I try to compare the bytes in the tests, they’re not going to match. To make a custom conversion from Export to bytes, I’d need to know the type of key being exported since both keys' conversion to bytes are different. 

Alternatively, I could also change the local storage or export type to do what the other does, but I’m not sure if that makes sense. Or I could do the convoluted thing of turning the export back into what it started as and turning that into bytes for comparison, but that doesn't seem right either.

picklenerd:
I think making `Export` generic is a good solution while we sort out what we want our types to look like. I just tested it locally and this seems to work. Is this what you were suggesting in the first comment?

```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct Export<T> {
    pub key_material: T,
    pub context: Vec<u8>,
}

impl From<SigningKeyPair> for Export<SigningKey> {
    fn from(key_pair: SigningKeyPair) -> Self {
        Self {
            key_material: key_pair.signing_key,
            context: key_pair.context.into(),
        }
    }
}

impl Export<SigningKey> {
    /// Convert `Export` into a [`SigningKeyPair`].
    pub fn into_signing_key(self) -> Result<SigningKeyPair, LockKeeperError> {
        let context = self.context.try_into()?;
        Ok(SigningKeyPair {
            signing_key: self.key_material,
            context,
        })
    }
}
```

vidyaap:
Oh yes, that is what I was originally asking. I thought your first reply was suggesting the extra field over the generic. I can go with the generic for this though!

picklenerd:
Yeah I was thinking we could come up with a permanent solution for this but that seems like a bit much for this PR, especially since we still need to discuss the database schema. I think your original suggestion is perfect for now.

vidyaap:
Summary of some thoughts from a call:
- We can't make Export a generic because there is nothing of the generic type in the Export struct; probably not worth putting a phantom data in there for that.
- We talked about whether LocalStorage and Export should behave more similarly. @marsella tagging you for if you have thoughts on this! (Possibly the schema that gets designed for #339 is related to this)
- For now, I'm just converting the Export back into a Secret to compare in the one test.

marsella:
> We can't make Export a generic

This is because we don't actually export our _own_ `EncryptionKey` or whatever type, right? We're exporting "the bytes"? I think this will come up when we start integrating, but it would be nice to still have those bytes be wrapped in some meaningful, well-documented type so calling applications don't have to figure out from scratch what the requirements / endianness on them are. Arbitrary secrets don't really have an external compatibility, but for an ECDSA key, we could look into parsing our internal types into, say, the corresponding RustCrypto type and having `Export` be generic over that.

However, in the meantime, I think the approach of having an extra field to specify the key type is fine.

> We talked about whether LocalStorage and Export should behave more similarly

A hint of what's to come for `LocalStorage`: we're thinking about encrypting (or at least MACing) whatever things belong in local storage under a key derived from the `MasterKey`, so that the calling application can't arbitrarily modify the stuff they're storing locally. An example of why this is worth it: when we sign something, the calling application should pass a public key to verify the signature; if LocalStorage doesn't have any integrity check, the client app can't distinguish between bad input from the calling app and a cheating server. 

I'm not sure what this means for whether it should behave "more similarly" to `Export`. My instinct is to leave it up to the calling app to figure out the keys for a given encrypted `LocalStorage` blob value; `LocalStorage` should really be a tool for the calling app to maintain some state relevant to our client app, while `Export` is for the calling app to use in their own applications. But I don't have super strong feelings; if you think it makes sense to put an extra field to specify key type in `LocalStorage` for now, that would also be okay -- we'll have to rethink this either way when we actually add that encryption/MAC. Does that help?

vidyaap:
Re: LocalStorage - Yes that helps! I agree with LocalStorage being more for internal-related context and Export being more external-related. I think the original context for this point was that LocalStorage wraps around the key type as a whole while Export separates the key material bytes from the context bytes and this is what was making the comparison difficult. But with their usage contexts being different, I feel like they shouldn't wrap around the types in the same way as each other.

lock-keeper/src/crypto/arbitrary_secret.rs
```diff
@@ -104,12 +104,29 @@ impl Secret {
...
+    }
+}
+
+impl Export {
+    /// Convert `Export` into a [`Secret`].
+    pub fn into_secret(self) -> Result<Secret, LockKeeperError> {
+        let context = self.context.clone().try_into()?;
+        let inner = generic::Secret::from_parts(&self.key_material, &context);
+        Ok(Secret(inner))
+    }
```
picklenerd, line 127: I think this can be `impl TryFrom<Export> for Secret`.

picklenerd (approved):
LGTM!

PR #373, opened 2022-11-21
Adding an encryption tunnel using session_key

### LINKED ISSUE #260: Refactor Channel types
The current channel implementation does basic serialization/deserialization of the bytes stored in our `Message` type. This provides a seamless interface for users of the channel to send or receive messages from the channel using whatever types are defined by the protocol.

For some message types, these blobs will eventually be encrypted by the session key so that they can only be read with a valid session. The channel interface should incorporate the encryption/decryption process while maintaining the same seamless interface for users of the channel.

It should be noted that not all message types will require a session key since they are processed before a session key exists. Registration and authentication are two examples. The new code should handle both situations without requiring protocols to change their logic.

## New Traits

I think the best way to implement this is to add two new traits for message types, one for authenticated messages and one for unauthenticated messages. They'll both contain two methods to convert to/from the RPC `Message` type. The methods for authenticated messages will also include a session key parameter.

Having two traits will make the channel implementation a little bit trickier. We should discuss how we want to deal with the two different message types for channels. We may want an enum to cover both cases, a third trait, or to have different channel types for each message type.

The traits might look something like this:

```rust
trait Message {
    fn from_rpc(message: RpcMessage) -> Result<Self, LockKeeperError> where Self: Sized;
    fn to_rpc(self) -> Result<RpcMessage, LockKeeperError>;
}

trait AuthenticatedMessage {
    fn from_rpc(message: RpcMessage, session_key: OpaqueSessionKey) -> Result<Self, LockKeeperError> where Self: Sized;
    fn to_rpc(self, session_key: OpaqueSessionKey) -> Result<RpcMessage, LockKeeperError>;
}
```

 We can provide implementations from one trait to the other which will block anyone from implementing both on the same struct:

```rust
impl<T: Message> AuthenticatedMessage for T {
    fn from_rpc(_message: RpcMessage, _session_key: OpaqueSessionKey) -> Result<Self, LockKeeperError> where Self: Sized {
        panic!("nope")
    }
    fn to_rpc(self, _session_key: OpaqueSessionKey) -> Result<RpcMessage, LockKeeperError> {
        panic!("nope")
    }
}

impl Message for String {
    fn from_rpc(message: RpcMessage) -> Result<Self, LockKeeperError> where Self: Sized {
        Ok("test".to_string())
    }
    fn to_rpc(self) -> Result<RpcMessage, LockKeeperError> {
        Err(LockKeeperError::InvalidMessage)
    }
}

// Error: conflicting implementations of trait `AuthenticatedMessage` for type `std::string::String`
impl AuthenticatedMessage for String {
    fn from_rpc(message: RpcMessage, session_key: OpaqueSessionKey) -> Result<Self, LockKeeperError> where Self: Sized {
        Ok("test".to_string())
    }
    fn to_rpc(self, session_key: OpaqueSessionKey) -> Result<RpcMessage, LockKeeperError> {
        Err(LockKeeperError::InvalidMessage)
    }
}
```

## Trait impls

We're currently using a simple macro to implement `TryFrom` on message types. It might be worth spending a little bit of time writing custom derive macros for these new traits so that it's completely clear from the type definition whether a message is authenticated or not.

```rust
    #[derive(Message)]
    pub struct RegisterStart {
        pub registration_request: RegistrationRequest<OpaqueCipherSuite>,
        pub account_name: AccountName,
    }

    #[derive(AuthenticatedMessage)]
    pub struct Generate {
        pub user_id: UserId,
    }
```
The macro implementation for unauthenticated messages will be the same as the current macro which just does basic serialization/deserialization.

The macro implementation for authenticated messages will do encryption or decryption first, and then do the same serialization/deserialization.

### LINKED ISSUE #327: Expand session key to support authenticated sessions
The `OpaqueSessionKey` type is currently not usable for anything. We want to use it to enforce an application-layer authenticated, encrypted channel. To do this, we need to implement AEAD encryption and decryption under the session key.

1. Outline an `encrypt` and `decrypt` function on `OpaqueSessionKey`. Check with developers about what message type it should take.
2. Implement `encrypt` and `decrypt` using the generic AEAD abstraction we already have (might have to change the representation of the key type to support this more easily).
3. Check in on whether there should be a further key derivation from the raw session key bytes to the AEAD key. If so, implement that (maybe at creation of the key).
4. Write appropriate tests.

Questions:
- boltlabs-inc/key-mgmt-spec#149 still talks about creating a MAC key (as does [this part of the spec](https://github.com/boltlabs-inc/key-mgmt-spec/blob/develop/systems-architecture.md#opening-the-application-layer-authenticated-channel)). Will we use a MAC in addition to an AEAD, or does the "authenticated" in AEAD handle that?

gijsvl:
Closes #327


gijsvl:
For the reviewers:
- @marsella I liked your idea of using custom derive macros. However, custom derive macros (or procedural macros) need to be in their own crate. Therefore, we would have to create a separate crate (possibly withing our repo) to provide these custom derive functions. We can still do this in the future, but I opted not to do this now. To avoid having to pass the session key every time in the `send` (and `receive`) function, I've changed the channel to contain an optional `session_key`, this makes it an (un)authenticated channel depending on the availability of the `session_key`. Moreover, I've added a Trait `ShouldBeAuthenticated` that needs to be added to objects to answer with true or false depending on the fact that an object need to be sent over an authenticated channel or not. This is added through the already existing macro, and I've created a new macro for authenticated objects (as you will see). Let me know if you have any concerns about this approach, or other comments...
- @picklenerd As discussed I've added an enforcement on the objects (as explained above using `ShouldBeAuthenticated`). Moreover, I've been able to use the same rng after all, so there is no issue of creating a new rng everytime. Let me know if you have any other comments.

jwantz:
@gijsvl: I believe this will also be a breaking change in the sense of backwards compatibility. That is, a v0.3 server will not be able to communicate with a v0.2 client. If you agree, then we should also add this to the readme and the release notes.

lock-keeper-client/src/client.rs
```diff
@@ -181,22 +185,56 @@ impl LockKeeperClient {
         // Server returns its own channel that is uses to send responses
         let server_response = match metadata.action() {
             ClientAction::Authenticate => client.authenticate(stream).await,
+            ClientAction::Register => client.register(stream).await,
+            _ => return Err(LockKeeperClientError::AuthenticatedChannelNeeded),
```
picklenerd, line 189: Can we list the error cases explicitly here? I want to make sure these functions show an error if a new `ClientAction` is added.

lock-keeper-client/src/client.rs
```diff
@@ -181,22 +185,56 @@ impl LockKeeperClient {
...
             ClientAction::Generate => client.generate(stream).await,
             ClientAction::ImportSigningKey => client.import_signing_key(stream).await,
-            ClientAction::Register => client.register(stream).await,
             ClientAction::RemoteGenerate => client.remote_generate(stream).await,
             ClientAction::RemoteSignBytes => client.remote_sign_bytes(stream).await,
             ClientAction::Retrieve => client.retrieve(stream).await,
             ClientAction::RetrieveAuditEvents => client.retrieve_audit_events(stream).await,
             ClientAction::RetrieveSigningKey => client.retrieve_signing_key(stream).await,
             ClientAction::RetrieveStorageKey => client.retrieve_storage_key(stream).await,
+            _ => return Err(LockKeeperClientError::UnauthenticatedChannelNeeded),
```
picklenerd, line 227: Same here

lock-keeper/src/crypto/generic.rs
```diff
@@ -26,7 +26,7 @@ pub enum CryptoError {
     #[error("Signature did not verify")]
     VerificationFailed,
 
-    #[cfg(test)]
+    // #[cfg(test)]
```
picklenerd, line 29: Just flagging this.

lock-keeper/src/lib.rs
```diff
@@ -46,6 +46,45 @@ macro_rules! impl_message_conversion {
...
+
+            impl TryFrom<$message_type> for $crate::types::Message {
+                type Error = $crate::LockKeeperError;
+
+                fn try_from(value: $message_type) -> Result<Self, Self::Error> {
+                    let content = serde_json::to_vec(&value)?;
+
+                    Ok($crate::types::Message { content })
+                }
+            }
```
picklenerd, line 81: I'm wondering if there's a way we can do this without duplicating the code. I was playing around and came up with this:

```rust
/// Common implementation for `impl_message_conversion` variants.
#[macro_export]
macro_rules! impl_message_conversion_internal {
    ($message_type:ty) => {
        impl TryFrom<$crate::types::Message> for $message_type {
            type Error = $crate::LockKeeperError;

            fn try_from(value: $crate::types::Message) -> Result<Self, Self::Error> {
                Ok(serde_json::from_slice(&value.content)?)
            }
        }

        impl TryFrom<$message_type> for $crate::types::Message {
            type Error = $crate::LockKeeperError;

            fn try_from(value: $message_type) -> Result<Self, Self::Error> {
                let content = serde_json::to_vec(&value)?;

                Ok($crate::types::Message { content })
            }
        }
    };
}

/// Generates `TryFrom` implementations to and from the `Message` type for a
/// given list of types.
#[macro_export]
macro_rules! impl_message_conversion {
    (unauthenticated, $($message_type:ty),+) => {
        $(
            $crate::impl_message_conversion_internal!{$message_type}

            impl $crate::infrastructure::channel::ShouldBeAuthenticated for $message_type {
                fn should_be_authenticated(&self) -> bool {
                    false
                }
            }
        )+
    };
    (authenticated, $($message_type:ty),+) => {
        // use $crate::impl_message_conversion_internal;
        $(
            $crate::impl_message_conversion_internal!{$message_type}

            impl $crate::infrastructure::channel::ShouldBeAuthenticated for $message_type {
                fn should_be_authenticated(&self) -> bool {
                    true
                }
            }
        )+
    };
}

```

Which looks like this in use:

```rust
    impl_message_conversion!(unauthenticated, AuthenticateStart, AuthenticateFinish);
    impl_message_conversion!(authenticated, SendUserId);
```

There's probably a better way to do this but it's late so this is what my tired brain did 😆

lock-keeper-key-server/src/server/operation.rs
```diff
@@ -29,13 +29,41 @@ pub(crate) trait Operation<DB: DataStore>: Sized + Send + 'static {
...
+                thread::sleep(Duration::from_millis(100));
+            } else {
+                audit_event(&mut channel, &context, EventStatus::Successful).await;
+            }
+        });
+
+        Ok(Response::new(ReceiverStream::new(rx)))
+    }
+
+    async fn handle_authenticated_request(
```
picklenerd, line 54: Can we break these up in a way that removes the duplicate code?

lock-keeper/src/infrastructure/channel.rs
```diff
@@ -20,47 +30,131 @@ pub type ClientChannel = Channel<Message>;
 pub struct Channel<T> {
     sender: Sender<T>,
     receiver: Streaming<Message>,
+    session_key: Option<OpaqueSessionKey>,
+    rng: Arc<Mutex<StdRng>>,
+}
+
+pub trait ShouldBeAuthenticated {
+    fn should_be_authenticated(&self) -> bool;
 }
```
picklenerd, line 39: Can we make this a hard requirement of the `Operation` trait in the server?

marsella (changes requested):
Thanks for this PR!

First off: perfectly reasonable to not do derive macros in this PR. It's already quite hefty :) and that will be fine to layer on at a later date.

Putting the session key into the channel also sound reasonable to me. It's also nice to be able to use the unauthenticated channel directly to send `Encrypted<Message>`! Architecture-wise, a channel lives only for the length of a request, right? So this will make sure that two parties can finish a single request, but it still leaves decisions about overall session management / expiration to the server-side storage. 

For the overall design, it sounds like the guarantee is, "if a channel has a session key and a given message should be authenticated, it will be authenticated." I was wondering if we can enforce something stronger: "if the channel has a session key, every message sent over it is authenticated." 

This seems hard. The goal would be misuse resistance -- to encrypt stuff by default and not require the developer to correctly flag individual protocol messages as "auth" / "noauth". Is it possible to design some type-level constraint that will cause a compile error if the developer tries to send a message that isn't explicitly "send me in the clear" over a channel that supports authentication?

I think it's worth thinking about. The current version is OK since a given protocol message doesn't _default_ to unauthenticated. The developer always has to make a decision. But it would be cool to have unauthenticated messages be the special case you have to flag.

Spec updates: I wanted to compare this to the spec but it looks like [the relevant section of the spec](https://github.com/boltlabs-inc/key-mgmt-spec/blob/develop/systems-architecture.md#opening-the-application-layer-authenticated-channel) is pretty out of date. Could you make an issue in the spec to change that to describe this updated use of an AEAD scheme? I flagged a detail or two inline that should definitely be included in the spec.

Some other clarification questions for me:
• I see some cloning of session keys. Can you explain what the best practice is for dropping these now that we have zeroize on drop? Should there be an explicit `drop` when things go out of scope or do we have confidence that the runtime will handle that?
• Where does the `Message` metadata live? It looks like it's totally separate from the `Message.content` in the proto file (that gets encrypted in the `To/From<Vec<u8>>` impl in the crypto module). Is that right? I remember a conversation with Jordan about the importance of not encrypting that metadata, and since the integration tests still work, that definitely isn't happening, but I don't follow the overall architecture.

lock-keeper/src/crypto.rs
```diff
@@ -35,11 +36,36 @@ pub use signing_key::{
...
+        self,
+        rng: &mut (impl CryptoRng + RngCore),
+        message: T,
+    ) -> Result<Encrypted<T>, CryptoError>
+    where
+        T: TryFrom<Vec<u8>>,
+        CryptoError: From<<T as TryFrom<Vec<u8>>>::Error>,
+        Vec<u8>: From<T>,
+    {
+        Encrypted::encrypt(rng, &self.0, message, &AssociatedData::new())
```
marsella, line 68: thought: I'm guessing we'll want to add some meaningful associated data here. Can you make a note to make a decision about this in the spec issue?

picklenerd:
>  Where does the Message metadata live? It looks like it's totally separate from the Message.content in the proto file (that gets encrypted in the To/From<Vec<u8>> impl in the crypto module). Is that right? I remember a conversation with Jordan about the importance of not encrypting that metadata, and since the integration tests still work, that definitely isn't happening, but I don't follow the overall architecture.

The metadata that I added is actually in the return types of the public `LockKeeperClient` API. For things that go across the network, the metadata is encoded into the gRPC requests/responses using its built-in metadata capabilities. It basically works the same way as HTTP headers. Once the request/response makes it across the network, the metadata is pulled out and placed into the channel.

Metadata is only exchanged when the channel is opened, not when messages are passed across. Once the protocol is complete and the channel is closed, the metadata sent from the server is pulled out of the channel and placed into the result returned by `LockKeeperClient`.

lock-keeper-key-server/src/server/session_key_cache.rs
```diff
@@ -79,19 +79,20 @@ impl SessionKeyCache {
...
+            }
             // The rest of the actions must be authenticated
             _ => {
                 let user_id = metadata
                     .user_id()
                     .as_ref()
                     .ok_or(LockKeeperServerError::InvalidAccount)?;
-                let server_session_key = self.get_key(user_id.clone())?;
-                Ok(())
+                Ok(self.get_key(user_id.clone())?)
```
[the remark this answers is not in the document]
picklenerd: I believe this is a side effect of the hashmap's `entry` api which is used to remove the session entry if the key is expired. Our terrible and temporary key eviction strategy is to remove expired sessions when a user tries to use the expired session. @gatoWololo does that sound right?

This cache is being replaced with something permanent during the current development phase.

lock-keeper/src/infrastructure/channel.rs
```diff
@@ -20,47 +30,131 @@ pub type ClientChannel = Channel<Message>;
 pub struct Channel<T> {
     sender: Sender<T>,
     receiver: Streaming<Message>,
+    session_key: Option<OpaqueSessionKey>,
+    rng: Arc<Mutex<StdRng>>,
```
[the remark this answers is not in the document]
picklenerd: I agree with the approach to keep it generic in the `lock-keeper` crate.

lock-keeper-key-server/src/server/session_key_cache.rs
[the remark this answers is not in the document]
gatoWololo: Yes, this is due to our use of the `Entry` API for `HashMap`. @marsella we _could_ take a reference and have the `get_key` function do the clone internally, but this allows the color to choose whether to clone or give up ownership of the `user_id` so it is a little more flexible, but less ergonomic.

We use the `Entry` API so we only have a single hashmap lookup in this function (`get_key`), we could do it without entry, but then we would need two lookups for deleting invalid keys.

lock-keeper-client/src/client.rs
gijsvl: done

lock-keeper/src/crypto/generic.rs
gijsvl: removed instead of comment

lock-keeper-client/src/client.rs
gijsvl: done

lock-keeper/src/lib.rs
gijsvl: I've just extracted the `TryFrom` part into a separate macro. I like that the calling macro is split into two separate macros to have a clear difference between authenticated and unauthenticated, I think it's less clear with that key word.

lock-keeper-key-server/src/server/operation.rs
gijsvl: Good call, I've removed the `handle_authenticated_request` in favor of one `handle_request` method. Should-be-authenticated messages over an unauthenticated channel would still fail because of the `ShouldBeAuthenticated` trait.

lock-keeper/src/infrastructure/channel.rs
```diff
@@ -20,47 +30,131 @@ pub type ClientChannel = Channel<Message>;
...
+        match self.session_key.clone() {
+            None => {
+                if message.should_be_authenticated() {
+                    return Err(LockKeeperError::ShouldBeAuthenticated);
+                }
+                Ok(message
+                    .try_into()
+                    .map_err(|_| Status::internal("Invalid message"))?)
+            }
+            Some(session_key) => {
```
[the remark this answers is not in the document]
gijsvl: I was contemplating about this, but this is exactly regarding your point in your general comment:

> For the overall design, it sounds like the guarantee is, "if a channel has a session key and a given message should be authenticated, it will be authenticated." I was wondering if we can enforce something stronger: "if the channel has a session key, every message sent over it is authenticated."

By ignoring this check, it actually ensures that even a message doesn't necessarily needs to go over an authenticated channel will still go over the authenticated channel. If I add the error, it won't.

gijsvl:
@marsella about your general question:
> • I see some cloning of session keys. Can you explain what the best practice is for dropping these now that we have zeroize on drop? Should there be an explicit drop when things go out of scope or do we have confidence that the runtime will handle that?

An explicit drop is not necessary, objects will be dropped at the end of their scope anyway and will be zeroized in the process. The reason I've added some explicit drops at some points in the zeroize PR, is to just be very thorough and explicit about it sometimes.

lock-keeper-client/src/client.rs
```diff
@@ -60,7 +60,7 @@ impl Password {
 #[derive(Debug)]
 #[allow(unused)]
 pub struct LockKeeperClient {
-    session_key: OpaqueSessionKey,
+    pub(crate) session_key: OpaqueSessionKey,
```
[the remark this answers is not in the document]
gijsvl: Yes, this should have been read-only, I forgot to revert this change.

lock-keeper-client/src/client.rs
```diff
@@ -181,22 +185,56 @@ impl LockKeeperClient {
...
             ClientAction::Retrieve => client.retrieve(stream).await,
             ClientAction::RetrieveAuditEvents => client.retrieve_audit_events(stream).await,
             ClientAction::RetrieveSigningKey => client.retrieve_signing_key(stream).await,
             ClientAction::RetrieveStorageKey => client.retrieve_storage_key(stream).await,
+            _ => return Err(LockKeeperClientError::UnauthenticatedChannelNeeded),
         }?
         .into_inner();
 
-        let mut channel = ClientChannel::create(tx, server_response);
+        let mut channel_rng = StdRng::from_entropy();
```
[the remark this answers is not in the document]
gijsvl: yes, my bad

lock-keeper/src/crypto.rs
```diff
@@ -35,11 +36,36 @@ pub use signing_key::{
...
+pub struct OpaqueSessionKey(EncryptionKey);
 
-impl From<GenericArray<u8, U64>> for OpaqueSessionKey {
-    fn from(arr: GenericArray<u8, U64>) -> Self {
-        Self(Box::new(arr.into()))
+impl TryFrom<GenericArray<u8, U64>> for OpaqueSessionKey {
+    type Error = LockKeeperError;
+
+    fn try_from(arr: GenericArray<u8, U64>) -> Result<Self, Self::Error> {
+        let context = AssociatedData::new().with_str("OPAQUE-derived Lock Keeper master key");
```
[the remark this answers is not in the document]
gijsvl: changed

lock-keeper/src/crypto.rs
```diff
@@ -35,11 +36,36 @@ pub use signing_key::{
...
+    }
+}
+
+impl OpaqueSessionKey {
+    /// Encrypt the given [`StorageKey`] under the [`MasterKey`] using an
+    /// AEAD scheme.
+    pub(crate) fn encrypt<T>(
+        self,
+        rng: &mut (impl CryptoRng + RngCore),
+        message: T,
```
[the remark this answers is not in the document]
gijsvl: I've changed the encrypt function to take a `Message`, that's a good call. However, if I pass in `content` directly, I would basically get a `Encrypted<Vec<u8>>` and I think I like it better handling an `Encrypted<Message>`. But happy to change if you think that is better.

lock-keeper/src/crypto.rs
```diff
@@ -227,6 +253,30 @@ impl Encrypted<StorageKey> {
     }
 }
 
+impl Encrypted<Message> {
+    pub fn decrypt_message(
+        self,
+        session_key: OpaqueSessionKey,
```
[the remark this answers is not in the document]
gijsvl: done

lock-keeper/src/infrastructure/channel.rs
```diff
@@ -20,47 +30,131 @@ pub type ClientChannel = Channel<Message>;
...
     /// Receive the next message on the channel and convert it to the type `R`.
+    /// If this is an authenticated channel the message gets decrypted first.
     /// If the message cannot be converted to `R`, it is assumed to be an
     /// invalid message and an error is returned.
     pub async fn receive<R: TryFrom<Message>>(&mut self) -> Result<R, LockKeeperError> {
         match self.receiver.next().await {
             Some(message) => {
                 let message = message?;
-                let result = R::try_from(message).map_err(|_| LockKeeperError::InvalidMessage)?;
+                let result = match self.session_key.clone() {
```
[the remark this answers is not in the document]
gijsvl: done

lock-keeper/src/crypto.rs
```diff
@@ -35,11 +36,36 @@ pub use signing_key::{
...
+            context,
+        )))
+    }
+}
+
+impl OpaqueSessionKey {
+    /// Encrypt the given [`StorageKey`] under the [`MasterKey`] using an
+    /// AEAD scheme.
+    pub(crate) fn encrypt<T>(
+        self,
```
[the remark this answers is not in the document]
gijsvl: done

lock-keeper/src/infrastructure/channel.rs
```diff
@@ -20,47 +30,131 @@ pub type ClientChannel = Channel<Message>;
...
 
+    /// This function encrypts the message using the session_key if this is an
+    /// authenticated channel. The object that is send needs to implement
+    /// the ShouldBeAuthenticated trait as an extra verification that the object
+    /// needs to be sent over an (un)authenticated channel.
+    async fn optional_encryption_with_session_key(
+        &mut self,
+        message: impl TryInto<Message> + ShouldBeAuthenticated,
+    ) -> Result<Message, LockKeeperError> {
+        match self.session_key.clone() {
```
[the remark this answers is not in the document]
gijsvl: done

lock-keeper/src/infrastructure/channel.rs
```diff
@@ -70,17 +164,37 @@ impl ServerChannel {
 
 impl ClientChannel {
     /// Create a client-side `Channel` that sends raw `Message` objects.
-    pub fn create(sender: Sender<Message>, receiver: Streaming<Message>) -> Self {
-        Self { sender, receiver }
+    ///
+    /// # Arguments
+    ///
+    /// * rng - a reference to a secure random number generated.
```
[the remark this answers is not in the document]
gijsvl: done

lock-keeper/src/infrastructure/channel.rs
```diff
@@ -70,17 +164,37 @@ impl ServerChannel {
...
 
     /// Send a message across the channel. This function accepts any type that
-    /// can be converted to a `Message.
-    pub async fn send(&mut self, message: impl TryInto<Message>) -> Result<(), LockKeeperError> {
-        let message = message
-            .try_into()
-            .map_err(|_| Status::internal("Invalid message"))?;
-
-        self.handle_send(message).await
+    /// can be converted to a `Message. In case a session_key is provided to the
```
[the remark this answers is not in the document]
gijsvl: done

lock-keeper/src/crypto.rs
```diff
@@ -35,11 +36,36 @@ pub use signing_key::{
...
+                .try_into()
+                .map_err(|_| LockKeeperError::Crypto(CryptoError::ConversionError))?,
+            context,
+        )))
+    }
+}
+
+impl OpaqueSessionKey {
+    /// Encrypt the given [`StorageKey`] under the [`MasterKey`] using an
+    /// AEAD scheme.
```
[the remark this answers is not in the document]
gijsvl: done

lock-keeper/src/infrastructure/channel.rs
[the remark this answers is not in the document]
gijsvl: I've made this change in one commit: https://github.com/boltlabs-inc/key-mgmt/pull/373/commits/504995c763e1a27da472c00a504e2e5da9fedc67 I'm not sure if this is much better, I'm not sure how else to easily achieve this, but maybe one of you have a better idea?

lock-keeper-client/src/api.rs
```diff
@@ -81,22 +83,35 @@ impl LockKeeperClient {
...
-        let client_channel = Self::create_channel(&mut client, &metadata).await?;
-        let result =
-            Self::handle_registration(client_channel, &mut rng, account_name, password).await;
+        let rng_arc_mutex = Arc::new(Mutex::new(rng));
+        let client_channel =
+            Self::create_channel(rng_arc_mutex.clone(), &mut client, &metadata).await?;
+        let result = {
+            let mut rng = rng_arc_mutex.lock().await;
+            Self::handle_registration(client_channel, &mut *rng, account_name, password).await
+        };
```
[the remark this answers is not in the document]
gijsvl: Not sure if this is what you mean: https://github.com/boltlabs-inc/key-mgmt/pull/373/commits/ca0a2831d5d2f205f890819e19c3c81fda097370

lock-keeper/src/crypto.rs
gijsvl: https://github.com/boltlabs-inc/key-mgmt-spec/issues/147

lock-keeper/src/crypto.rs
```diff
@@ -35,11 +36,36 @@ pub use signing_key::{
...
-impl From<GenericArray<u8, U64>> for OpaqueSessionKey {
-    fn from(arr: GenericArray<u8, U64>) -> Self {
-        Self(Box::new(arr.into()))
+impl TryFrom<GenericArray<u8, U64>> for OpaqueSessionKey {
+    type Error = LockKeeperError;
+
+    fn try_from(arr: GenericArray<u8, U64>) -> Result<Self, Self::Error> {
+        let context = AssociatedData::new().with_str("OPAQUE-derived Lock Keeper master key");
+        Ok(Self(EncryptionKey::from_bytes(
+            arr[..32]
```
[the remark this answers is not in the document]
gijsvl: https://github.com/boltlabs-inc/key-mgmt-spec/issues/147

lock-keeper-client/src/client.rs
```diff
@@ -165,33 +170,105 @@ impl LockKeeperClient {
...
+            ClientAction::RetrieveAuditEvents => {
+                return Err(LockKeeperClientError::AuthenticatedChannelNeeded)
+            }
+            ClientAction::RetrieveSigningKey => {
+                return Err(LockKeeperClientError::AuthenticatedChannelNeeded)
+            }
+            ClientAction::RetrieveStorageKey => {
+                return Err(LockKeeperClientError::AuthenticatedChannelNeeded)
+            }
+        }?;
```
picklenerd, line 222:
```suggestion
        let server_response = match metadata.action() {
            ClientAction::Authenticate => client.authenticate(stream).await,
            ClientAction::Register => client.register(stream).await,

            // These actions generate an error because they should be on an authenticated channel
            ClientAction::CreateStorageKey
            | ClientAction::Export
            | ClientAction::ExportSigningKey
            | ClientAction::Generate
            | ClientAction::ImportSigningKey
            | ClientAction::RemoteGenerate
            | ClientAction::RemoteSignBytes
            | ClientAction::Retrieve
            | ClientAction::RetrieveAuditEvents
            | ClientAction::RetrieveSigningKey
            | ClientAction::RetrieveStorageKey => {
                return Err(LockKeeperClientError::AuthenticatedChannelNeeded)
            }
        }?;
```

lock-keeper-client/src/client.rs
```diff
@@ -165,33 +170,105 @@ impl LockKeeperClient {
...
             ClientAction::RetrieveSigningKey => client.retrieve_signing_key(stream).await,
             ClientAction::RetrieveStorageKey => client.retrieve_storage_key(stream).await,
+
+            // These actions generate an error because they should be on an unauthenticated channel
+            ClientAction::Authenticate => {
+                return Err(LockKeeperClientError::UnauthenticatedChannelNeeded)
+            }
+            ClientAction::Register => {
+                return Err(LockKeeperClientError::UnauthenticatedChannelNeeded)
+            }
```
picklenerd, line 263:
```suggestion
            ClientAction::Authenticate | ClientAction::Register => {
                return Err(LockKeeperClientError::UnauthenticatedChannelNeeded)
            }
```

lock-keeper-client/src/client.rs
picklenerd: Added a formatting suggestion to reduce duplication.

lock-keeper/src/crypto/generic.rs
```diff
@@ -7,7 +7,7 @@ use serde::{Deserialize, Serialize};
 use std::{iter, marker::PhantomData};
 use thiserror::Error;
 
-#[cfg(test)]
+// #[cfg(test)]
```
picklenerd, line 10: This one can go too.

lock-keeper/src/infrastructure/channel.rs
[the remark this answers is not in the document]
picklenerd: This is gross but I'm not sure there's a better way to handle it aside from passing the RNG with every `send` and `receive`, which I definitely think is worse.

lock-keeper/src/crypto.rs
```diff
@@ -228,6 +249,30 @@ impl Encrypted<StorageKey> {
...
+    }
+}
+
+impl From<Message> for Vec<u8> {
+    fn from(message: Message) -> Self {
+        message.content
+    }
+}
+
+impl_message_conversion!(Encrypted<Message>);
```
picklenerd, line 274: I'm not a fan of this approach. The `Encrypted` wrapper is supposed to guarantee that the thing inside is encrypted. This allows us to easily produce unencrypted data that claims to be encrypted. This may not cause problems in the current code but I'm worried that this could cause problems in the future.

The code that creates an `Encrypted<Message>` should have its own implementation that explicitly states what it's doing and what it guarantees. I don't think it should use the existing macro or any `std` traits.

@marsella

lock-keeper-client/src/api.rs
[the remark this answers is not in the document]
marsella: I mean that the RNG is created on L88 and goes out of scope at L130 and I don't think this function ever spawns multiple threads that all have access to the same RNG. I thought we might not need the `Arc<Mutex<Rng>>` -- could just pass the `Rng` directly without having any issues with multiple threads accessing it simultaneously.

But it looks like the methods we call _do_ require the `Arc<Mutex<>>`.

lock-keeper/src/infrastructure/channel.rs
```diff
@@ -22,33 +29,97 @@ pub type ClientChannel = Channel<Message, ResponseMetadata>;
...
+    /// During authentication a channel can be upgraded to become an
+    /// authenticated channel. Such that future messages are sent while
+    /// being encrypted by the session_key.
+    pub fn try_upgrade_to_authenticated(
+        &mut self,
+        session_key: OpaqueSessionKey,
+    ) -> Result<(), LockKeeperError> {
+        if self.session_key.is_some() {
+            return Err(AlreadyAuthenticated);
+        }
```
picklenerd, line 118: What situation does this protect against?

lock-keeper-key-server/src/server/session_key_cache.rs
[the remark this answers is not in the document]
marsella: Nah, this is fine. I was just wondering. Thanks!

lock-keeper/src/crypto.rs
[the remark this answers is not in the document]
marsella: Oh yeah, that's a good call.

lock-keeper/src/infrastructure/channel.rs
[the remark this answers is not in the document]
marsella: Ah, right. That's nice. Can you add an inline comment here to that effect? Like 
```
// If the channel has a session key, _always_ encrypt a message sent over it!
```

lock-keeper/src/crypto.rs
marsella: I think this is okay. We're reusing the AEAD encryption scheme we already implemented to do encryption under `SessionKey`s. 

E.g. we take a `Message`, encrypt it under a session key to get an `Encrypted<Message>`, and then send that blob over the network. So this impl is guaranteeing that once we encrypt a given message (e.g. that calls `impl_authenticated_message_conversion`), we can send that blob.

marsella: Perhaps this could use a line of documentation to explain that context. Alternately / additionally, @picklenerd, do you think it would be better to have a wrapper type for `Message`s encrypted under a `SessionKey`, so it's more visually different from, say, an `Encrypted<StorageKey>` that's encrypted under the master key?

picklenerd: This code allows a developer to take an unencrypted `Message` and turn it into an `Encrypted<Message>` for any `Message` whose JSON representation looks like: 

```json
{
    "ciphertext": [..bytes],
    "associated_data": [..bytes],
    "nonce": [..bytes],
}
```

So not an issue with the current code but definitely not misuse resistant.

I don't think we should use the `impl_message_conversion ` macro for this since it obscures the conversion process.

I don't think this should be a `From` or `TryFrom` either because it's too easy to use without knowing the context for converting. These could just as easily be doing the same JSON conversion and I doubt anyone would check before using it.

I think this should be a pair of methods on the `Encrypted<Message>` type that explicitly state what they do and what guarantees are made about the conversion. If we do the same conversion through JSON, we can at least document the methods and state this.

picklenerd: This is one of the reason why I named the trait methods described in the ticket the way I did. The current macro hides what's happening a bit too much, especially when you can just `try_from()` and `try_into()` the `Message` type.

lock-keeper-client/src/api.rs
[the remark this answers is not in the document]
gijsvl: I think passing it on as an Arc<Mutex<>> as long as possible is the best approach, as soon as we lock it and translate it to an rng, we have to wait till it gets dropped for it to release. I've already encountered a deadlock now because of this, keeping the lock as short as possible is key. Which also means that the function that gets called within this scope ideally is as short as possible.

lock-keeper/src/infrastructure/channel.rs
gijsvl: I wanted to avoid that we accidentally override a `session_key` that is used for a channel and as such transforming a channel into being reused for a different session. (technically, updating from non-authenticated to an authenticated channel is also reusing a channel for a different session, but with a very specific use-case)

lock-keeper-client/src/client.rs
gijsvl: woops, of course.

lock-keeper-client/src/client.rs
gijsvl: Done! Thanks!

lock-keeper/src/crypto/generic.rs
gijsvl: Removed

lock-keeper/src/infrastructure/channel.rs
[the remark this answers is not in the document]
gijsvl: Added!

lock-keeper/src/crypto.rs
gijsvl: I've created custom `try_from_message()` and `try_into_message()` methods on `Encrypted<Message>`

lock-keeper/src/crypto.rs
```diff
@@ -228,6 +249,83 @@ impl Encrypted<StorageKey> {
...
+
+    /// Translates an [`Encrypted<Message>`] to a [`Message`] in order to be
+    /// sent through an authenticated channel.
+    pub(crate) fn try_into_message(self) -> Result<Message, LockKeeperError> {
+        // ciphertext len || ciphertext || associated_data len || associated_data ||
+        // nonce len || nonce
+        let content = iter::once(self.ciphertext.len() as u8)
+            .chain(self.ciphertext.to_owned())
+            .chain(iter::once(
+                Vec::<u8>::from(self.associated_data.clone()).len() as u8,
```
picklenerd, line 269: `ciphertext` and `associated_data` don't have restricted lengths so we can't assume that the length fits in a byte. For example, the `remote_sign_bytes::RequestRemoteSign` message contains an arbitrary vector of bytes so this message can be any length.

picklenerd, line 269: I think we could just add a `len` method to `AssociatedData` instead of cloning here.

gatoWololo:
Jordan has asked me to do a code review to get another set of eyes on this! I will have it in by today.

lock-keeper/src/crypto.rs
gijsvl: I've moved it back to json representation, which is probably the safer bet. Unless you see any potentially better solution?

gijsvl: same comment as above.

picklenerd: I think this is a good solution for now since this is what we use for other encodings. We'll come up with a permanent encoding solution soon.

picklenerd:
I removed #260 from the list of tickets closed by this PR. We're going to leave it open for the derive macro.

lock-keeper-client/src/api/authenticate.rs
```diff
@@ -13,17 +14,20 @@ use lock_keeper::{
...
 };
-use rand::{CryptoRng, RngCore};
+use rand::{rngs::StdRng, CryptoRng, RngCore};
+use tokio::sync::Mutex;
 
 impl LockKeeperClient {
     pub(crate) async fn handle_authentication<T: CryptoRng + RngCore>(
-        channel: &mut ClientChannel,
-        rng: &mut T,
+        channel: &mut ClientChannel<StdRng>,
```
gatoWololo, line 22: The fact our `ClientChannel` and `ServerChannel` have a `StdRng` is an implementation detail, we should not have to think about this when using them.

We should change the `rng` field in `Channel` definition to something like:
```rust
rng: Arc<Mutex<dyn CryptoRng + RngCore>>,
```
(I think the Mutex takes care of the boxing so we don't need a `Box` around the dynamic trait). This way we don't need the type parameter `G` and all our `ClientChannel`s or `ServerChannel`s.

Otherwise we might just be able to change it in the type aliases to be:
```rust
pub type ServerChannel = Channel<Result<Message, Status>, RequestMetadata, StdRng>;
pub type ClientChannel = Channel<Message, ResponseMetadata, StdRng>;
```
This might be cleaner?

gatoWololo (changes requested):
It took a while to go through this hefty PR!

There has been a lot of churn, so it would be good to get this merged even if its now perfect. We think that's the best thing to do. So I have some minor feedback, we will leave #384 open and come back at a later time. I have some thoughts on a design iteration for authenticated channels. I will post these in #384 

I also don't like that our rng is locked behind a mutex shared by all concurrent operations. I know there is concerns about entropy, but I don't think using a Mutex is the right solution for this. @marsella we should have a conversation about this when you get back!

I have some minor feedback for now.

lock-keeper-client/src/api/authenticate.rs
picklenerd: I like the `dyn` trait in the `rng` field if that compiles.

gijsvl: I tried that at first, but it didn't compile that easily. So, first I get this error:
Screenshot 2022-12-05 at 10 58 00
I've tried to use the compiler's suggestion, but it also gets very messy because it needs to match the rng that we pass in. Basically, we would have to change the rng everywhere to this "NewTrait". 

I might be missing something here and I'm open to suggestions...

gatoWololo: Gotcha. I didn't realize Rust didn't allow this! @picklenerd do you have a reason you don't like defining `ClientChannel` as so:
```rust
pub type ClientChannel = Channel<Message, ResponseMetadata, StdRng>;
```

gatoWololo (approved):

marsella (approved):
Sorry, looks like my review is blocking this. LGTM.

PR #374, opened 2022-11-21
Simplified server tasks in Makefile.toml #368

### LINKED ISSUE #368: Clean up server tasks in `Makefile.toml`
We have a large number of `cargo make` tasks to start the server. With the addition of mutual auth, I added two new tasks for each case do the number has grown significantly.

It would be nice to parameterize these. It's possible to forward arguments from cargo make to the commands it runs, but I'm not sure if there's a way to dispatch to other commands based on the the arguments.

If we can't parameterize them this way, it would at least be useful to make the tasks with the shortest names start both servers, since this is what's required to run the full test suite.

picklenerd:
This PR simplifies the `cargo make` server commands.

- Added shorthand alias `cargo make run` to start server in the foreground.
- Renamed attached/detached tasks to be explicit.
- `cargo make start` and `cargo make run` start two servers, one with mutual auth enabled and one with mutual auth disabled.
- Removed task to run locally since it's just a basic `cargo run`.
- Removed `client_auth_enabled` from client `Config` since its only use was invalidated by these changes.
-  Updated README.

Closes #368

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,7 +40,7 @@ jobs:
           args: --debug cargo-make
 
       - name: Start Server
-        run: cargo make start-server-detached-all
+        run: cargo make start
           
       - name: Run all tests
         run: cargo make all-tests
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -3,13 +3,13 @@ default_to_workspace = false
 
 # -- Server Tasks --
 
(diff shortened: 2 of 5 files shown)


gatoWololo (approved):
Looks great! Thanks for doing this :pray:

picklenerd:
I do it for the people

PR #377, opened 2022-11-29
Made a generic secret type

picklenerd:


lock-keeper/src/crypto/generic.rs
```diff
@@ -217,7 +215,13 @@ where
             original_type: PhantomData,
         })
     }
+}
 
+impl<T> Encrypted<T>
+where
+    T: TryFrom<Vec<u8>>,
+    CryptoError: From<<T as TryFrom<Vec<u8>>>::Error>,
```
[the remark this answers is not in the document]
picklenerd: Nope I just wanted to make it a little more idiomatic. I'll switch it back and add a comment about the intent!

databases/lock-keeper-mongodb/src/api/secret.rs
```diff
@@ -19,14 +20,12 @@ impl Database {
     pub(crate) async fn add_user_secret(
```
vidyaap, line 20: I can't actually comment on the right line, but there are a few docs in this file that still reference `StoredEncryptedSecret`. Could you update that to `StoredSecret`?

lock-keeper-client/src/api.rs
```diff
@@ -180,10 +182,10 @@ impl LockKeeperClient {
     pub async fn generate_and_store(
```
vidyaap, line 182: nit: could we update the API function names to say "generate_secret" and "export_secret" too?

lock-keeper-key-server/src/database.rs
```diff
@@ -45,38 +45,33 @@ pub trait DataStore: Send + Sync + 'static {
     ) -> Result<Vec<AuditEvent>, Self::Error>;
 
     // Secret
-    /// Add a [`StoredEncryptedSecret`] to a [`User`]'s list of arbitrary
+    /// Add a [`Encrypted<Secret>`] to a [`User`]'s list of arbitrary
```
vidyaap, line 48: doc: add a `StoredSecret` instead of `Encrypted<Secret>`.

lock-keeper-key-server/src/database.rs
```diff
@@ -45,38 +45,33 @@ pub trait DataStore: Send + Sync + 'static {
...
-        secret: SigningKeyPair,
-        key_id: KeyId,
-    ) -> Result<(), Self::Error>;
-
-    /// Get a [`User`]'s [`StoredSigningKeyPair`] based on its [`KeyId`]
-    async fn get_user_signing_key(
+    /// Get a [`User`]'s [`StoredSecret`] based on its [`KeyId`].
+    /// A [`StoredSecret`] will only be returned if it matches the given
+    /// [`SecretFilter`].
+    async fn get_user_secret_with_filter(
```
vidyaap, line 69: Can we have there be just one `get_user_secret` trait function and move passing `SecretFilter::default()` to whatever is calling it (e.g. in retrieve)?

lock-keeper-key-server/src/database.rs
```diff
@@ -107,3 +102,16 @@ pub trait DataStore: Send + Sync + 'static {
         storage_key: Encrypted<StorageKey>,
     ) -> Result<(), Self::Error>;
 }
+
+#[derive(Clone, Debug, Default)]
+pub struct SecretFilter {
```
vidyaap, line 107: Could you document the intention you described earlier about using this to add more filters eventually?

lock-keeper-key-server/src/operations/remote_sign_bytes.rs
```diff
@@ -22,15 +22,16 @@ impl<DB: DataStore> Operation<DB> for RemoteSignBytes {
...
             .db
-            .get_user_signing_key(&request.user_id, &request.key_id)
+            .get_user_secret(&request.user_id, &request.key_id)
             .await
             .map_err(LockKeeperServerError::database)?
-            .signing_key
-            .to_owned()
             .try_into()?;
 
+        // Pretend to decrypt our placeholder type
```
vidyaap, line 32: 😆

lock-keeper/src/types/database/secrets.rs
```diff
@@ -1,75 +1,150 @@
...
+            retrieved: false,
+        })
+    }
+}
+
+pub mod secret_types {
+    pub const ARBITRARY_SECRET: &str = "arbitrary_secret";
+    pub const SIGNING_KEY_PAIR: &str = "signing_key_pair";
+    pub const REMOTE_SIGNING_KEY: &str = "remote_signing_key";
+}
```
vidyaap, line 77: This is probably just me, but I'm not sure I understand why we need the strings when we also have the enum and are using the enum everywhere besides the StoredSecret itself? I think I remember it being about flexibility to add new secret types - is that still what is it?

vidyaap (commented):
I also wanted to ask if the DB config changes we talked about were for this PR or for the future? I think it would be nice if the URI field didn't say "mongo" for now.

picklenerd:
> I also wanted to ask if the DB config changes we talked about were for this PR or for the future? I think it would be nice if the URI field didn't say "mongo" for now.

I started to do this but decided it should be a separate PR, especially given the discussions we had about where the database configuration should live. I'll make a ticket for that so we don't forget.

lock-keeper/src/types/database/secrets.rs
picklenerd: I think they should be strings in the database so they need some kind of string representation in the code. I don't want to encode the enum into the database type since we might be adding new values to it over time and that will be a breaking change.

The enum is much easier to work with though, which is why I left it in there. I wouldn't be against removing the enum and leaving just the strings.

@gatoWololo Thoughts?

vidyaap: > I think they should be strings in the database so they need some kind of string representation in the code. [...]

Could we use the strum to_string() for this when we initialize StoredSecrets?

gatoWololo: > They could but any time we add secret types to the enum it would be a breaking change to anyone who implemented our DataStore trait.

I'm biased towards enums. I like the `#[non_exaustive]` solution with an enum. But we can keep it as is? If it's annoying later we can switch from strings to enums?

vidyaap (approved):
Thanks for the changes and issue, LGTM!

PR #383, opened 2022-11-30
Use volume mounts for dev/ files

vidyaap:
I mapped the local dev/ folder to a /develop volume on the Docker image and changed some of the config file paths around to match. This also seems to be working with the service provider both with a local build and a github build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,4 +8,6 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry \
     --mount=type=cache,target=/usr/src/lock-keeper-key-server/target \
     cargo install --path ./bin/key-server-cli
 
+VOLUME /app
+
 CMD ["sh", "-c", "key-server-cli ${CONFIG}"]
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -8,4 +8,6 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry \
     --mount=type=cache,target=/usr/src/lock-keeper-key-server/target \
     cargo install --path ./bin/key-server-cli
 
+VOLUME /develop
```
picklenerd, line 11: Can we call this something like `app` instead? We'll be using this outside of development environments too.

lock-keeper-key-server/src/config.rs
```diff
@@ -28,7 +28,7 @@ pub struct Config {
 
 impl Config {
     pub fn from_file(
-        config_path: impl AsRef<Path>,
+        config_path: impl AsRef<Path> + std::fmt::Debug,
```
picklenerd, line 31: I think you can remove this part now!

picklenerd (approved):

PR #384, opened 2022-12-01
Factor out authentication checking into its own function.

gatoWololo:
No longer coupled to session key cache.

Quick change before we start work on the session tickets!

### DIFF
--- a/lock-keeper-key-server/src/server/operation.rs
+++ b/lock-keeper-key-server/src/server/operation.rs
@@ -2,14 +2,25 @@ use async_trait::async_trait;
 use lock_keeper::{
     constants::METADATA,
     infrastructure::channel::ServerChannel,
-    types::{audit_event::EventStatus, operations::ResponseMetadata, Message, MessageStream},
+    types::{
+        audit_event::EventStatus,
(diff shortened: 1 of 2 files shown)


lock-keeper-key-server/src/server/operation.rs
```diff
@@ -31,29 +41,37 @@ pub(crate) trait Operation<DB: DataStore>: Sized + Send + 'static {
         mut context: Context<DB>,
         request: Request<Streaming<Message>>,
     ) -> Result<Response<MessageStream>, Status> {
+        info!("Handling new client request.");
         let (mut channel, rx) = ServerChannel::create(request)?;
 
-        {
-            let mut session_key_cache = context.session_key_cache.lock().await;
-            session_key_cache.check_key::<DB>(channel.metadata())?;
-        }
```
picklenerd, line 39: This scope is here to drop the session key lock as soon as we're done with it. Does the type in the method signature for `check_authentication` accomplish the same thing?

gatoWololo: I have added back the scope to ensure it gets dropped!

PR #387, opened 2022-12-01
Temp fix for error message sleep

picklenerd:
This little chunk of code is already a hack while we figure out a better way to do this, but it shouldn't be sleeping the thread and it shouldn't wait for 100ms. This PR changes it to a `tokio` sleep and reduces the time to 10ms.


gatoWololo (approved):
LGTM

PR #390, opened 2022-12-02
Trait for session key cache WIP #385

### LINKED ISSUE #385: Abstraction layer over the session key cache
(Splitting this work out of #351)

We should use a trait for the session key cache (similar to the one created for the database in #359) to abstract over its operations.  This will allow us to plug in different implementations (e.g. LRU cache over hashmap, or if we wanted to use Redis in the future).  This abstraction would also be used as setup for the remaining work in #351 to improve our session handling.

- [ ] Create a trait for session key cache operations
- [ ] Implement the trait for our current implementation in a separate crate

vidyaap:
I wanted to get some initial questions out of the way before continuing:

- I've just followed what we did for the database for now, but that makes the session key cache an input when starting the server as well - is that weird?
- I would have liked to use a trait function to enforce the expiration time in a constructor, but then I would need to add "Sized" to that function signature - if that is still the better way to do it, I can change it to that.
- Where should the Drop impl go now that the cache is a trait? Or how do we enforce that?
- Should we move some of the logging from the implementation to the server itself based on the responses from the implementation?

Also sorry it looks like a lot of changed files, but it's mostly a rename of a directory!

Closes #385


picklenerd:
> I've just followed what we did for the database for now, but that makes the session key cache an input when starting the server as well - is that weird?

Nope not weird. That's what we'll have to do for all of these pluggable services. This is just dependency injection without the BS!


> I would have liked to use a trait function to enforce the expiration time in a constructor, but then I would need to add "Sized" to that function signature - if that is still the better way to do it, I can change it to that.

Expiration time shouldn't be part of the trait since we want to allow users to configure their cache however they want. Like if they're running `redis` then they probably have their own config files for that which we wouldn't know what to do with. We should defer expiration enforcement to the implementation.


> Where should the Drop impl go now that the cache is a trait? Or how do we enforce that?

That should go in the implementation. For our reference implementation, we can use the same zeroizing code. For external services, we have no control over it so it can't be part of the abstraction.


> Should we move some of the logging from the implementation to the server itself based on the responses from the implementation?

Yeah I think the server should log what it's doing with the cache. The cache should also report what it's doing separately. I think it would be helpful to treat the reference implementation as an external service and pretend it emits its own logs even though they'll also get picked up by `tracing` in reality.

vidyaap:
> Expiration time shouldn't be part of the trait since we want to allow users to configure their cache however they want... 

We are still configuring the expiration time in the server's config though, right? i.e. wouldn't we want to make sure we can tell the cache implementation what the session timeout in the server config is?

picklenerd:
> We are still configuring the expiration time in the server's config though, right? i.e. wouldn't we want to make sure we can tell the cache implementation what the session timeout in the server config is?

Yeah I could go either way on that. I don't think it's possible to enforce it in any way so it would always just be a suggestion for external implementations. I'm not sure if that's valuable, but Jay might have some helpful thoughts on that. We can bring it up in the meeting on Monday.

It makes sense to be in the config if we're using the reference implementation though. We'll need a session cache config section similar to the DB one anyway, so we could put it in there as an optional field that's mandatory for the reference implantation. At least for now.

gatoWololo (approved):
LGTM!

Issue #209, opened 2022-09-07
Reduce cloning of secrets


marsella:
The crypto APIs currently have a lot of parameters that are passed by value. When I first designed them, I was reading straight from the spec and thought that it would be reasonable to consume values when it looked like they would only be used once.

However, this hasn't quite worked out in practice and there are now lots of places in the crypto module where we have to clone secrets and key types in order to get the basic functionality working.

Look through the implementation for clones and try to update the code to remove as many as possible. Some examples:
- The `Encrypted::encrypt` function takes the object by value. Change it to take the object by reference. Update the `Into<Vec<u8>>` bound to work on references (and maybe return a reference or a `Vec<&u8>` as well).

marsella:
Had a conversation with @picklenerd about this. General approach is to have generation functions (or anything that needs to take ownership of a parameter) take it by value. Then the caller can decide whether they're done with that item or if they should clone it.

PR #391, opened 2022-12-05
Encrypt signing keys on the server side #382

### LINKED ISSUE #382: [Placeholder] Implement server-side encrypted storage
[Placeholder Ticket]

gijsvl:
Closes #382


bin/key-server-cli/src/main.rs
```diff
@@ -24,7 +27,19 @@ pub async fn main() {
         .transpose()
         .unwrap();
 
-    let config = Config::from_file(&cli.config, private_key_bytes).unwrap();
+    let server_side_encryption_key_bytes = cli
```
gatoWololo, line 30: We probably don't need "server_side" as part of the name here. Since we are already in the server module. This will cut down on the long names.

bin/key-server-cli/src/main.rs
```diff
@@ -11,6 +11,9 @@ pub struct Cli {
     /// Base64 encoded private key data
     #[clap(long)]
     pub private_key: Option<String>,
+    /// Base64 encoded server side encryption key data
+    #[clap(long)]
+    pub server_side_encryption_key: Option<String>,
```
gatoWololo, line 16: Same as previous comment about naming.

dev/server-side-encryption/generate-test-server-side-encryption-key.sh
```diff
@@ -0,0 +1,9 @@
+SCRIPT_DIR="dev/server-side-encryption"
```
gatoWololo, line 1: I know this script is pretty trivial but could you add a comment sentence on what it does specially the "32" argument to openssl.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -237,6 +205,25 @@ impl Encrypted<SigningKeyPair> {
...
+        let identifying_context = AssociatedData::new().with_bytes(user_id).with_bytes(key_id);
+
+        let context = identifying_context.with_str(ServerSideEncryptionKey::domain_separator());
+
+        if self.associated_data == context {
+            Ok(self.decrypt(&server_side_encryption_key.0)?)
+        } else {
+            Err(CryptoError::DecryptionFailed.into())
+        }
+    }
```
gatoWololo, line 226: Could we get rid of "by_server" in the fn name and "server_side" on the arg. I think the doc string is enough to know this is a server operation.

lock-keeper/src/types/database/secrets.rs
```diff
@@ -1,15 +1,14 @@
...
 use serde::{Deserialize, Serialize};
-use zeroize::ZeroizeOnDrop;
 
 /// Holds user's stored secrets of all types
 #[derive(Debug, Default, Deserialize, Serialize)]
 pub struct StoredSecrets {
     pub arbitrary_secrets: Vec<StoredEncryptedSecret>,
     pub signing_keys: Vec<StoredEncryptedSigningKeyPair>,
-    pub server_created_signing_keys: Vec<StoredSigningKeyPair>,
+    pub server_created_signing_keys: Vec<StoredEncryptedSigningKeyPair>,
```
gatoWololo, line 11: I see this name was already there but how do people feel about just `EncryptedSigningKeyPair`? Trying to keep the names at reasonable length :)

lock-keeper/src/crypto.rs
```diff
@@ -65,6 +64,62 @@ impl OpaqueSessionKey {
...
+
+    /// Returns the server side encryption key found in the file at the given
+    /// path
+    pub fn read_from_file(path: impl AsRef<Path>) -> Result<Self, LockKeeperError> {
+        let bytes = std::fs::read(path)?;
+        Self::read_from_bytes(&bytes)
+    }
+
+    /// Returns the server side encryption key found in the given bytes
+    pub fn read_from_bytes(bytes: &[u8]) -> Result<Self, LockKeeperError> {
```
gatoWololo, line 93: Can we just make this just `from_bytes`

lock-keeper/src/crypto.rs
```diff
@@ -65,6 +64,62 @@ impl OpaqueSessionKey {
...
+        let key = EncryptionKey::from_bytes(
+            bytes
+                .try_into()
+                .map_err(|_| LockKeeperError::InvalidServerSideEncryptionKey)?,
+            AssociatedData::new().with_str("ChaCha20Poly1305 with 96-bit nonce."),
```
gatoWololo, line 98: Could we factor out this string literal into either a constant or a function?

lock-keeper/src/crypto.rs
```diff
@@ -65,6 +64,62 @@ impl OpaqueSessionKey {
...
+        );
+
+        Ok(Self(key))
+    }
+
+    /// Encrypt the given [`SigningKeyPair`] under the
+    /// [`ServerSideEncryptionKey`] using an AEAD scheme.
+    pub fn encrypt_signing_key_pair(
+        &self,
+        rng: &mut (impl CryptoRng + RngCore),
```
gatoWololo, line 108: `generate<Rng: CryptoRng + RngCore>(rng: &mut Rng)` above uses a generic, this function uses an `impl`. We should stick to one convention. I think most of the codebase uses the `impl`. So we probably wanna change `generate`

gatoWololo (changes requested):
Thanks for the PR! Overall it looks great and I only have minor suggestions for names.

lock-keeper/src/crypto/signing_key.rs
gijsvl: Mmm, I'm not sure, there is a `decrypt_signing_key()` method that can be used by the client to decrypt using their storage key. This one uses the server side encryption key, happy to remove the `by_server`, but what do you suggest the name to be to not have a naming collision?

lock-keeper/src/types/database/secrets.rs
gijsvl: To keep things consistent we should also rename `StoredEncryptedSecret` in that case... but I'm not necessarily opposed.

bin/key-server-cli/src/main.rs
gijsvl: Agreed, done.

bin/key-server-cli/src/main.rs
gijsvl: Done

dev/server-side-encryption/generate-test-server-side-encryption-key.sh
gijsvl: Added some extra comments for clarification

lock-keeper/src/crypto.rs
gijsvl: done

lock-keeper/src/crypto.rs
gijsvl: Moved it to a function in `EncryptionKey`

lock-keeper/src/crypto.rs
gijsvl: Made the change, should we also change the `generate` function of StorageKey? I reused that method signature...

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -237,6 +205,25 @@ impl Encrypted<SigningKeyPair> {
...
+    /// subprotocol to retrieve a signing key from the server.
+    pub fn decrypt_signing_key_by_server(
+        self,
+        server_side_encryption_key: &ServerSideEncryptionKey,
+        user_id: UserId,
+        key_id: KeyId,
+    ) -> Result<SigningKeyPair, LockKeeperError> {
+        let identifying_context = AssociatedData::new().with_bytes(user_id).with_bytes(key_id);
+
+        let context = identifying_context.with_str(ServerSideEncryptionKey::domain_separator());
```
marsella, line 219: todo: The associated data for an encrypted signing key pair should describe the key pair, not the encryption key. Here it should be
- user id
- key id
- generation type ("imported" or "server-generated" are both safe)

lock-keeper/src/crypto.rs
```diff
@@ -65,6 +64,62 @@ impl OpaqueSessionKey {
...
+        signing_key_pair: SigningKeyPair,
+        user_id: &UserId,
+        key_id: &KeyId,
+    ) -> Result<Encrypted<SigningKeyPair>, LockKeeperError> {
+        let associated_data = AssociatedData::new()
+            .with_bytes(user_id.clone())
+            .with_bytes(key_id.clone())
+            .with_str(Self::domain_separator());
```
marsella, line 116: todo: The associated data here should describe the signing key pair being encrypted, not the encryption key. So this should vary based on what type of key it is. The existing correlation is the `create_and_encrypt` and `import_and_encrypt` functions on `Secret` (but that doesn't translate directly because in this case, I don't think it necessarily makes sense to bundle generation with encryption).

There are a couple ways you could do this -- make separate functions like on `Secret` or pass an additional enum parameter here called, say, `GenerationType`.

lock-keeper/src/crypto.rs
```diff
@@ -65,6 +64,62 @@ impl OpaqueSessionKey {
     }
 }
 
+/// The server side encryption key is a default-length symmetric encryption key
+/// for an AEAD scheme.
+///
+/// The server side encryption key is used by the key server to securely encrypt
+/// signing keys generated on the server side.
```
marsella, line 71: nit: it's also used to encrypt signing keys _imported_ server-side. Maybe you could generalize to "...signing keys known to the server." (but maybe that is more confusing).

lock-keeper/src/crypto.rs
```diff
@@ -65,6 +64,62 @@ impl OpaqueSessionKey {
...
+        Self::from_bytes(&bytes)
+    }
+
+    /// Returns the server side encryption key found in the given bytes
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self, LockKeeperError> {
+        let key = EncryptionKey::from_bytes(
+            bytes
+                .try_into()
+                .map_err(|_| LockKeeperError::InvalidServerSideEncryptionKey)?,
+            AssociatedData::new().with_str(EncryptionKey::domain_separator()),
```
marsella, line 98: thought: I wonder if we should add associated data here (and in `generate`) to mark this explicitly as a `ServerSideEncryptionKey`. Not sure I built the APIs to make that easy, and I don't think we ever use or look at the AD anyway...

marsella (changes requested):
Thanks for this PR!

Request changes is for this fix: associated data isn't quite being used correctly here. We want the AD for an encrypted key to describe the encrypted thing, not the thing doing the encryption. So, for encrypting a signing key with the server-side encryption key, the AD would be user id, key id, and a description of how the key was generated. (Right now, it's user id, key id, and description of the server-side encryption key). I put notes inline at encryption and decryption about this. I think you modeled this after `StorageKey`, which is both an encryptor and an encryptee :) In this case SigningKeyPair should demo the encryptee behavior.

Other thoughts:
I considered suggesting bundling the encryption and decryption steps into the database insertion and retrieval functions. I'm not sure it gains us much in the current version, but right now the server DB doesn't distinguish between signing key pairs encrypted under the `ServerSideEncryptionKey` vs the client-side `StorageKey`. If we add additional flows later that involve storage, it might not be clear to devs how to generate that `Encrypted<SigningKeyPair>` and they might try to use the wrong key. After consideration, I think it's okay to leave as-is since the documentation on the storage key types is pretty explicit and devs should model new flows after existing server-side flows.

Organizationally, I have let `crypto.rs` get a bit out of hand. Would it make sense to you to have a separate `storage_key` module that holds `ServerSideEncryptionKey` and `StorageKey`, and then to export them from the top-level module?

lock-keeper/src/crypto/signing_key.rs
marsella: Urgh. I was going to suggest at some point that we rename the `decrypt` function in `generic` to `decrypt_inner` and then rename all the special decryptions (`decrypt_secret`, `decrypt_signing_key`) to just plain `decrypt`. But that doesn't help this problem, where we can decrypt `Encrypted<SigningKeyPair>` with two different keys (and they have potentially different associated data so genericizing over them would be a little complicated).

gijsvl:
> Thanks for this PR!
> 
> Request changes is for this fix: associated data isn't quite being used correctly here. We want the AD for an encrypted key to describe the encrypted thing, not the thing doing the encryption. So, for encrypting a signing key with the server-side encryption key, the AD would be user id, key id, and a description of how the key was generated. (Right now, it's user id, key id, and description of the server-side encryption key). I put notes inline at encryption and decryption about this. I think you modeled this after `StorageKey`, which is both an encryptor and an encryptee :) In this case SigningKeyPair should demo the encryptee behavior.

Agreed, my mistake, I'll make that change.

> 
> Other thoughts: I considered suggesting bundling the encryption and decryption steps into the database insertion and retrieval functions. I'm not sure it gains us much in the current version, but right now the server DB doesn't distinguish between signing key pairs encrypted under the `ServerSideEncryptionKey` vs the client-side `StorageKey`. If we add additional flows later that involve storage, it might not be clear to devs how to generate that `Encrypted<SigningKeyPair>` and they might try to use the wrong key. After consideration, I think it's okay to leave as-is since the documentation on the storage key types is pretty explicit and devs should model new flows after existing server-side flows.

I've thought about that, but I think it would be a bad practice to add this to the database code, because I believe the database code should be somewhat independent of any additional functionality, that code should be in charge of storing and retrieving data, while the encryption is somewhat a extra functional choice here. So, yes, I believe leave as-is seems to be the right choice here.

> 
> Organizationally, I have let `crypto.rs` get a bit out of hand. Would it make sense to you to have a separate `storage_key` module that holds `ServerSideEncryptionKey` and `StorageKey`, and then to export them from the top-level module?

Yes, I think that is a good idea, I had considered that as well, will make the change.

lock-keeper/src/crypto.rs
```diff
@@ -65,6 +64,62 @@ impl OpaqueSessionKey {
...
 
+/// The server side encryption key is a default-length symmetric encryption key
+/// for an AEAD scheme.
+///
+/// The server side encryption key is used by the key server to securely encrypt
+/// signing keys generated on the server side.
+#[derive(Debug, Clone, Zeroize, ZeroizeOnDrop)]
+pub struct ServerSideEncryptionKey(EncryptionKey);
+
+impl ServerSideEncryptionKey {
```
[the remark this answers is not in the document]
gijsvl: Probably `RemoteStorageKey` gets closest to be in line with other naming we use.

lock-keeper/src/crypto/signing_key.rs
gijsvl: I've made a change, to take the associated data from the signing key pair, if I understand your comments well that should just be reflected with the encrypted version. The `context()` method on `SigningKeyPair` was only exposed for test, but I've opened it up to be used here, could you make sure this is desirable?

lock-keeper/src/crypto.rs
gijsvl: I've added the fact that it can be imported keys as well

lock-keeper/src/crypto.rs
gijsvl: I'm not sure I understand your comment or what change you want to see.

lock-keeper/src/crypto/signing_key.rs
gijsvl: I've done the rename to `decrypt_inner`, I optimized the code a little bit to remove some duplicate code, but I'm not sure this is remotely what you envisioned...

lock-keeper/src/crypto.rs
```diff
@@ -65,6 +64,62 @@ impl OpaqueSessionKey {
...
+    }
+
+    /// Encrypt the given [`SigningKeyPair`] under the
+    /// [`ServerSideEncryptionKey`] using an AEAD scheme.
+    pub fn encrypt_signing_key_pair(
+        &self,
+        rng: &mut (impl CryptoRng + RngCore),
+        signing_key_pair: SigningKeyPair,
+        user_id: &UserId,
+        key_id: &KeyId,
```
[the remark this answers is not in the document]
gijsvl: I'm not entirely convinced this is going to make things easier at this point, could we maybe just leave this to #209?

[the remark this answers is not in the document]
marsella: Sure

lock-keeper/src/crypto.rs
marsella: Ah, I think I got confused here, sorry. I can't see exactly where this landed in the new diff but I think you're good.

lock-keeper/src/crypto.rs
marsella: A `RemoteStorageKey` wraps an `EncryptionKey` that holds some associated data. Right now, that associated data is just the generic default AD. I was wondering if we should try to make it something specific to `RemoteStorageKey`.

We never access that associated data, though. Unlike other types, it's not incorporated into encryption or serialized at any point. It also matches the `StorageKey`. So I think it's okay to leave as-is.

lock-keeper/src/crypto/signing_key.rs
marsella: It is fine to make the context public!

This looks good to me, thanks for fixing.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -59,6 +67,12 @@ impl Signable for SignableBytes {
     }
 }
 
+pub mod generation_types {
+    pub const SERVER_GENERATED: &str = "server-generated";
+    pub const CLIENT_GENERATED: &str = "client-generated";
+    pub const IMPORTED: &str = "imported key";
+}
```
marsella, line 74: This is nice.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -224,14 +205,45 @@ impl Encrypted<SigningKeyPair> {
...
+    ) -> Result<SigningKeyPair, LockKeeperError> {
+        let identifying_context = AssociatedData::new().with_bytes(user_id).with_bytes(key_id);
 
-        // Keys encrypted under a storage key are either client-generated or imported.
-        if self.associated_data == import_context || self.associated_data == client_context {
-            Ok(self.decrypt(&storage_key.0)?)
+        if possible_context_strings
+            .iter()
+            .map(|context| identifying_context.clone().with_str(context))
+            .any(|x| x == self.associated_data)
```
marsella, line 244: This is really nice!

lock-keeper/src/crypto/storage_key.rs
```diff
@@ -0,0 +1,320 @@
+use super::{
```
marsella, line 1: nit: Please add a line of module-level docs here.

marsella (approved):

picklenerd:
Please squash this into one commit before merging.

lock-keeper/src/crypto/storage_key.rs
gijsvl: Added

lock-keeper/src/crypto/signing_key.rs
gijsvl: Leaving it as is for now? We can discuss a possible rename in a bigger refactor along the lines @marsella suggests, unless you don't agree @gatoWololo ?

gatoWololo: I think it is fine to leave as is!

gatoWololo (approved):
Looks good! Let's get this merged in.

PR #396, opened 2022-12-07
Make TLS optional on the server #348

### LINKED ISSUE #348: Decouple TLS from Key Server application framework
**Background:** The TLS terminating endpoint for the Key Server is currently embedded in the Key Server framework. We anticipate future deployments where the TLS termination may be distributed to other front-end nodes.

**Thoughts on Requirements:**

1. Offer a configuration option that allows the Key Server to be deployed with or without TLS termination. Provide appropriate documentation.
1. Select an open source vendor product that offers TLS termination for development and deployment testing.
1. Document the deployment and configuration of the TLS front-end component for future reference.
1. Hold a peer review/discussion of the product selection, configuration and deployment options.

picklenerd:
This PR makes TLS optional on the server and updates the configs accordingly.

I will make a separate PR to integrate `nginx` into our test stack.

### DIFF
--- a/dev/docker/Server.toml
+++ b/dev/docker/Server.toml
@@ -1,11 +1,13 @@
 address = "0.0.0.0"
 port = 1113
 session_timeout = "60s"
+opaque_path = "/app/opaque"
+opaque_server_key = "/app/opaque/server_setup"
+
+[tls_config]
 private_key = "/app/test-pki/gen/certs/server.key"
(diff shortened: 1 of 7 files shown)


vidyaap (commented):
The changes themselves look good to me, but it looks like this branch has the same mutual auth test error that was pointed out before. I haven't seen it on local runs yet though.

vidyaap (approved):
Thanks for making #402 for the test error! LGTM!

PR #397, opened 2022-12-07
Create separate table for secrets in db

picklenerd:
This PR separates users and secrets into different database tables. This allows us to return a user's info without returning all of their secrets.

Since key ids are 32 bytes, we can't use the same UUID generation method as user ids. Instead, I made a helper type called `HexBytes` that allows us to serialize byte collections as hex strings. This type ensures that the serialized representation of a type is a plain string that will place nice with databases, while the Rust representation is a pure byte array.

@marsella Since this exists now, I changed the `UserId` type back to bytes similar to the `KeyId` type. It's still generated the same way, but we don't use the randomly generated bytes to create a UUID anymore.

### DIFF
--- a/lock-keeper-key-server/src/database.rs
+++ b/lock-keeper-key-server/src/database.rs
@@ -47,11 +47,7 @@ pub trait DataStore: Send + Sync + 'static {
     // Secret
     /// Add a [`StoredSecret`] to a [`User`]'s list of arbitrary
     /// secrets
-    async fn add_user_secret(
-        &self,
-        user_id: &UserId,
-        secret: StoredSecret,
-    ) -> Result<(), Self::Error>;
+    async fn add_user_secret(&self, secret: StoredSecret) -> Result<(), Self::Error>;
 
     /// Get a [`User`]'s [`StoredSecret`] based on its [`KeyId`].
     /// A [`StoredSecret`] will only be returned if it matches the given
--- a/lock-keeper/src/types/database.rs
+++ b/lock-keeper/src/types/database.rs
@@ -2,3 +2,49 @@
 
 pub mod secrets;
 pub mod user;
+
+use std::fmt::Display;
+
+use serde::{Deserialize, Serialize};
+
+use crate::LockKeeperError;
+
+/// Convenience type for serializing byte arrays as hex strings.
+/// Add the `#[serde(try_from = "HexBytes", into = "HexBytes")]` attribute macro
+/// above any type you'd like to serialize this way. This type should only be
+/// used to serialize byte collections. It should not be used directly.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub(crate) struct HexBytes(String);
+
+impl Display for HexBytes {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl<T: AsRef<[u8]>> From<T> for HexBytes {
+    fn from(bytes: T) -> Self {
+        Self(hex::encode(bytes))
+    }
+}
+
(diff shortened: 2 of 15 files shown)


vidyaap:
FYI, I think the duplicate key errors here are because we have a unique index on KeyId which is an array and the mongo "uniqueness" check fails if any one element is repeated between key IDs. I think we had the same issue with UserIds which is why we went from bytes to uuid.

picklenerd:
> FYI, I think the duplicate key errors here are because we have a unique index on KeyId which is an array and the mongo "uniqueness" check fails if any one element is repeated between key IDs. I think we had the same issue with UserIds which is why we went from bytes to uuid.

Thank you! That was a much easier fix knowing where to look!

picklenerd:
I was thinking about this little serialization thing I did for the KeyId and thought it might be better to make a generic type to use for serializing byte arrays. Re-drafting while I update this PR.

lock-keeper/src/types/database.rs
```diff
@@ -2,3 +2,47 @@
...
+use std::fmt::Display;
+
+use serde::{Deserialize, Serialize};
+
+use crate::LockKeeperError;
+
+/// Convenience type for serializing byte arrays as hex strings.
+/// Add the `#[serde(try_from = "HexBytes", into = "HexBytes")]` attribute macro
+/// above any type you'd like to serialize this way. This type should only be
+/// used to serialize byte collections. It should not be used directly.
```
picklenerd, line 15: This last bit may not need to be true. This could be useful in places like the client CLI where we store or display these hex bytes. We can add to this type if we want and make it useful in other situation.

lock-keeper-key-server/src/database.rs
```diff
@@ -47,11 +47,7 @@ pub trait DataStore: Send + Sync + 'static {
     // Secret
     /// Add a [`StoredSecret`] to a [`User`]'s list of arbitrary
```
gatoWololo, line 48: Is this comment still relevant? Since there is no more user's list of arbitrary secrets?

gatoWololo (approved):
Looks good to me! (I reviewed this to see how the new db "schema" looked like.)

lock-keeper-key-server/src/database.rs
picklenerd: Good question. I think the comment implies a relation between the `User` object and the `StoredSecret` object which we would expect to exist in any database implementation. On the other hand, it doesn't reflect our Rust data model anymore.

I could go either way on this.

lock-keeper/src/types/database/user.rs
```diff
@@ -47,19 +47,18 @@ impl User {
 }
 
 /// Unique ID for a user.
-#[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, Eq, Hash)]
-pub struct UserId(String);
+#[derive(Clone, Serialize, Deserialize, Default, PartialEq, Eq, Hash)]
+#[serde(try_from = "HexBytes", into = "HexBytes")]
+pub struct UserId(Box<[u8; 16]>);
```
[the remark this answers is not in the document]
picklenerd: These could take up a lot of memory if we're processing a bunch of requests at once, so this is to guard against stack overflows.

See: https://github.com/boltlabs-inc/key-mgmt/pull/207#discussion_r964086070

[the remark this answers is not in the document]
gatoWololo: Gotcha. Could we add a comment to reflect this on the type? Otherwise this will get lost forever lol

lock-keeper-key-server/src/database.rs
gatoWololo: Yeah, this is a minor thing and the current comment is still good enough.

PR #400, opened 2022-12-09
Report test errors

vidyaap:
Returns an error from `test_suites::run_tests()` if any test fails so that CI will report it.

Also fixes a couple of bugs with logging and the register e2e test. Unsure why the Status -> LKClientError isn't happening there but I just switched to comparing the raw statuses for now.


lock-keeper-tests/src/test_suites.rs
```diff
@@ -25,5 +29,22 @@ pub async fn run_all(config: &Config) -> Result<()> {
...
+}
+
+fn check_errors(results: &[TestResult]) -> Result<()> {
+    use TestResult::*;
+
+    if results.iter().any(|r| *r == Failed) {
+        Err(LockKeeperTestError::TestFailed)
+    } else {
+        Ok(())
+    }
```
picklenerd, line 49: I like this approach but this will only work for `test_suites::run_all`. Could we move this up to `main.rs` and do it for all of the test types?

Alternatively, we could remove the `test_type` option entirely and always run all of the tests. I think the filters accomplish the same goal.

vidyaap: My thought was we only need the program to actually panic in CI which only runs all-tests, but if not I think I prefer moving it to main over removing the test-types.

picklenerd (approved):

PR #401, opened 2022-12-09
Add nginx with TLS #348

### LINKED ISSUE #348: Decouple TLS from Key Server application framework

picklenerd:
This PR adds TLS termination via `nginx` and creates a new environment where this setup can be tested.

In order to make testing easier, I introduced an `Environments` type to `lock-keeper-tests`. This type expands the `client_config` and `mutual_auth_config` fields that were in `lock_keeper_tests::Config` into a mapping from environment name to client config. This allows us to run end-to-end tests across all of our different environments, as well as being able to pull out configs for specific environments for tests that require them.

`client_config` is given the name `standard`. `mutual_auth_config` was given the name `client_auth` (see #410). These two configs are required to be in the environment definition file. The test binary will return an error if one of them is missing.

Other changes:
- root CA cert is now generated using `sha256` instead of `sha1`.
- `key_server_mutual_auth` was renamed `ks_client_auth` in docker compose.
- `http` urls are now allowed in the client. Previously, only `https` urls were allowed.
- `lock-keeper-tests` now prints error messages instead of panicking.
- `lock-keeper-tests::Config` was removed.
- The `run_parallel!` macro now takes a `&TestFilters` instead of a `lock-keeper-tests::Config`.

Closes #348

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,20 @@ services:
     ports:
       - 27017:27017
     stdin_open: true
-    tty: true 
+    tty: true
+
+  nginx:
+    image: nginx:latest
+    container_name: nginx_tls
+    restart: always
+    ports:
+      - 1115:1115
+    depends_on:
+      - "ks_nginx"
+    volumes:
+      - ./dev/config/nginx-tls/nginx.conf:/etc/nginx/nginx.conf
+      - ./dev/test-pki/gen/certs/server.key:/etc/ssl/certs/server.key
+      - ./dev/test-pki/gen/certs/server.chain:/etc/ssl/certs/server.crt
 
   key_server:
     build: .
@@ -20,7 +33,7 @@ services:
     ports:
       - 1113:1113
     healthcheck:
-      test: "chmod +x healthcheck.sh && ./healthcheck.sh http://localhost:1113"
+      test: "sh ./healthcheck.sh http://localhost:1113"
       interval: 10s
       timeout: 5s
     depends_on:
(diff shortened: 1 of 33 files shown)


lock-keeper-tests/src/test_suites/mutual_auth.rs
```diff
@@ -33,24 +30,24 @@ pub async fn run_tests(config: &Config) -> Result<Vec<TestResult>> {
     Ok(results)
 }
 
-async fn mutual_auth_not_required(config: Config) -> Result<()> {
-    let result = LockKeeperClient::health(&config.client_config).await;
+async fn mutual_auth_not_required(environments: Environments) -> Result<()> {
+    let result = LockKeeperClient::health(environments.client_auth_config()?).await;
```
vidyaap, line 34: Should this one be the standard config instead?

picklenerd: Yes!

docker-compose.yml
```diff
@@ -48,3 +58,14 @@ services:
       - "mongodb"
     stdin_open: true
     tty: true
+
+  ks_nginx:
```
vidyaap, line 62: Do we want the health check on this one too?

vidyaap: Never mind, I don't think the same check works for this.

picklenerd: I think I fixed this... we'll see.

vidyaap (approved):
LGTM! Thanks for figuring out the health check!

PR #404, opened 2022-12-13
Remove session timeout default #403

### LINKED ISSUE #403: Remove default for `session_timeout` in server config
The server config has `#[serde(default)]` set for `session_timeout`. This will be 0 if no value is set in the config which makes this configuration unusable.

The local configs should be updated to specify the timeout.

picklenerd:
Removes session timeout default and fixes local configs.

Closes #403


vidyaap (approved):
Thank you!

PR #405, opened 2022-12-14
392 separate configs

### LINKED ISSUE #392: Separate database config from server config
We currently have the database configuration built into the server configuration. This makes sense for the `docker compose` based local workflow, but it doesn't make sense in a normal server environment with an independent database.

The current config looks like this:

```rust
pub struct DatabaseSpec {
    pub mongodb_uri: String,
    pub db_name: String,
}
```

We may need to maintain the URI in the server config, but the `db_name` is an implementation detail of the database. Our `lock-keeper-mongodb` reference implementation should have its own config that contains this field and possibly also the table names. This config should be loaded somehow when we instantiate the database connection.

The server config should be reduced to the minimal amount of configuration necessary to connect to an arbitrary database. This might be difficult to make assumptions about right now, but I feel like a full URI is the safest bet. We should meet up and discuss a general strategy for configuring all of our plugin services before this is implemented.

vidyaap:
Closes #392 

Also I have a separate commit to move the CLI to the bin/ directory - I meant to do that when we moved files for the data store trait.


bin/key-server-cli/src/config.rs
```diff
@@ -0,0 +1,30 @@
...
+    }
+}
+
+impl FromStr for Config {
+    type Err = toml::de::Error;
+
+    fn from_str(config_string: &str) -> Result<Self, Self::Err> {
+        toml::from_str(config_string)
+    }
+}
```
vidyaap, line 30: I am unwrapping here and didn't make an error type for the binary since this should exit if we can't read the config file at all, but would it be better to have an error type anyway?

picklenerd: I think this is fine. I would just change them to `expect` so that you can explain the failure in the panic message.

persistence/lk-session-hashmap/src/api.rs
```diff
@@ -13,7 +14,7 @@ pub struct HashmapKeyCache {
     /// Map from user ids to (timestamp, keys).
     cache: HashMap<UserId, (Instant, OpaqueSessionKey)>,
     /// Time a key is considered valid for.
-    expiration: Duration,
+    session_timeout: Duration,
```
picklenerd, line 17: I think `expiration` is the right word here. `timeout` implies some kind of connection problem.

persistence/lk-session-hashmap/src/config.rs
```diff
@@ -0,0 +1,28 @@
...
+use serde::{Deserialize, Serialize};
+use std::{path::Path, str::FromStr, time::Duration};
+
+use crate::Error;
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub struct Config {
+    #[serde(with = "humantime_serde")]
+    pub session_timeout: Duration,
```
picklenerd, line 12: I think this should be `session_expiration`.

picklenerd (approved):
LGTM!

PR #406, opened 2022-12-17
WIP Postgres Implementation

gatoWololo:
This is ready for review! A few comments:
- You will see some TODOs. These will be addressed in the follow up PR.
- I spoke with Jay and we think bytea is the right type for some of the fields.
- I changed some table fields to use VARCHAR instead of text.


gatoWololo:
@picklenerd @vidyaap @jwantz Feel free to look over at the 1_migration.sql file:
```sql
CREATE TABLE IF NOT EXISTS Users
(
    user_id bytea PRIMARY KEY UNIQUE,
    account_name text NOT NULL UNIQUE,
    storage_key bytea,
    server_registration bytea NOT NULL
);

CREATE INDEX  IF NOT EXISTS idx_user_id ON Users(user_id);

CREATE TABLE IF NOT EXISTS Secrets
(
    key_id bytea UNIQUE NOT NULL,
    user_id bytea NOT NULL,
    secret bytea NOT NULL,
    secret_type TEXT NOT NULL,
    retrieved INT NOT NULL,
    PRIMARY KEY (key_id),
    FOREIGN KEY (user_id) REFERENCES Users(user_id)
);

CREATE TABLE IF NOT EXISTS AuditEvents
(
    account_name TEXT NOT NULL,
    key_id bytea,
    action TEXT NOT NULL,
    event_status TEXT NOT NULL,
    timestamp TIMESTAMPTZ NOT NULL,
    audit_event_id BIGINT GENERATED ALWAYS AS IDENTITY,
    PRIMARY KEY (audit_event_id),
    -- Cannot make account name foreign key as, on a new connection, we try to insert an entry into AuditEvents before
    -- The account_name exists in Users table.
    -- FOREIGN KEY (account_name) REFERENCES Users(account_name),
    FOREIGN KEY (key_id) REFERENCES Secrets(key_id)
);

```

Here are the tables I'm using. You can take a look at `persistence/lock-keeper-sql/src/lib.rs` to see our queries. Let me know if you have any early feedback! We can talk about it during our working session tomorrow. Or whenever people are back from holiday.

persistence/lock-keeper-sql/migrations/1_migration.sql
```diff
@@ -0,0 +1,35 @@
+CREATE TABLE IF NOT EXISTS Users
+(
+    user_id bytea PRIMARY KEY UNIQUE,
```
picklenerd, line 3: @jwantz mentioned that we should have a DB generated field for all of the primary keys. I'm guessing just an auto-incremented integer. We would never actually return it from the API, but it would be a faster index.

Is that right Jay?

jwantz: Yes, that is correct. The DB-generated ID would serve as the primary key for each table and also as the foreign key (with indexing) on associated tables. This raises a new discussion we should have around guidelines for naming our columns.

persistence/lock-keeper-sql/migrations/1_migration.sql
```diff
@@ -0,0 +1,35 @@
...
+    account_name text NOT NULL UNIQUE,
+    storage_key bytea,
+    server_registration bytea NOT NULL
+);
+
+CREATE INDEX  IF NOT EXISTS idx_user_id ON Users(user_id);
+
+CREATE TABLE IF NOT EXISTS Secrets
+(
+    key_id bytea UNIQUE NOT NULL,
```
jwantz, line 13: Very minor, but worthy of discussion... would we like to standardize on some guidelines? For instance, should primary keys follow the same as the table name, such that "Secrets" has a "secret_id"?

jwantz, line 13: I think it would be best to make all primary keys of type "bigint".

jwantz: Actually, I think that should be "bigserial" for primary keys, which is the auto generated type for bigints.

persistence/lock-keeper-sql/migrations/1_migration.sql
```diff
@@ -0,0 +1,35 @@
+CREATE TABLE IF NOT EXISTS Users
+(
+    user_id bytea PRIMARY KEY UNIQUE,
+    account_name text NOT NULL UNIQUE,
+    storage_key bytea,
```
jwantz, line 5: We need to keep in mind that `bytea` is not an ANSI data type and therefore not as portable to other DB implementations. I'm not saying that we shouldn't use it -- I'm just saying we should discuss it in more depth.

persistence/lock-keeper-sql/migrations/1_migration.sql
```diff
@@ -0,0 +1,35 @@
...
+);
+
+CREATE INDEX  IF NOT EXISTS idx_user_id ON Users(user_id);
+
+CREATE TABLE IF NOT EXISTS Secrets
+(
+    key_id bytea UNIQUE NOT NULL,
+    user_id bytea NOT NULL,
+    secret bytea NOT NULL,
+    secret_type TEXT NOT NULL,
```
jwantz, line 16: In an earlier discussion, I think we considered having a `secret_types` table, which would assign an ID to enumerated types of secrets. If we had such a table, then this column would be a foreign key to `secret_type_id`. Is this still a consideration or am I thinking of the wrong thing here?

persistence/lock-keeper-sql/migrations/1_migration.sql
```diff
@@ -0,0 +1,35 @@
...
+    PRIMARY KEY (key_id),
+    FOREIGN KEY (user_id) REFERENCES Users(user_id)
+);
+
+CREATE TABLE IF NOT EXISTS AuditEvents
+(
+    account_name TEXT NOT NULL,
+    key_id bytea,
+    action TEXT NOT NULL,
+    event_status TEXT NOT NULL,
```
jwantz, line 27: I'm not sure that `TEXT` is the correct data type for either `action` or `event_status`. In the future, I think we want to be able to easily search on actions. Should these be integer IDs into a table that holds and enumeration of actions? Maybe we should discuss the intended purpose of these columns in more detail. Thanks.

persistence/lock-keeper-sql/migrations/1_migration.sql
```diff
@@ -0,0 +1,35 @@
...
+    timestamp TIMESTAMPTZ NOT NULL,
```
jwantz, line 28: Note to self: I want to look up and be sure that `TIMESTAMPTZ` is the best approach for future portability.

persistence/lock-keeper-sql/migrations/1_migration.sql
```diff
@@ -0,0 +1,35 @@
...
+    secret bytea NOT NULL,
+    secret_type TEXT NOT NULL,
+    retrieved INT NOT NULL,
+    PRIMARY KEY (key_id),
+    FOREIGN KEY (user_id) REFERENCES Users(user_id)
+);
+
+CREATE TABLE IF NOT EXISTS AuditEvents
+(
+    account_name TEXT NOT NULL,
```
jwantz, line 24: If we add an `account_id` column to the (newly named) `accounts` table, then I think this foreign key should be to the `account_id` column (with a bigint data type).

persistence/lock-keeper-sql/migrations/1_migration.sql
```diff
@@ -0,0 +1,35 @@
...
+    key_id bytea,
+    action TEXT NOT NULL,
+    event_status TEXT NOT NULL,
+    timestamp TIMESTAMPTZ NOT NULL,
+    audit_event_id BIGINT GENERATED ALWAYS AS IDENTITY,
```
jwantz, line 29: This is a very minor comment, but it fits with typical style guidelines for DB developers... I would consider moving all of your primary keys to be the first column in each of the table definitions.

persistence/lock-keeper-sql/migrations/1_migration.sql
picklenerd: We discussed this in our working session and determined that it would actually be useful to return the ID with the model so that we can use it for the rest of the session.

persistence/lock-keeper-sql/migrations/1_migration.sql
picklenerd: I think this is a good point. It would be nice if people could use our migrations directly.

persistence/lock-keeper-sql/migrations/1_migration.sql
picklenerd: Yeah we talked about that. I'm in favor of this so that we can easily keep track of secret types. I think this will be really useful for testing if we end up making allowed secret types configurable.

persistence/lock-keeper-sql/migrations/1_migration.sql
picklenerd: I agree that these should be defined in another table.

persistence/lock-keeper-sql/migrations/1_migration.sql
picklenerd: It looks like `Z` includes a time zone. These should be stored in UTC without the option to specify a time zone.

picklenerd: Reading a little bit more, `TIMEZONETZ` actually converts the timestamp you give it based on the database's configured timezone, and then stores it as UTC. That sounds error prone to me, so I stand behind my previous comment!

persistence/lock-keeper-sql/migrations/1_migration.sql
gatoWololo: > We discussed this in our working session and determined that it would actually be useful to return the ID with the model so that we can use it for the rest of the session.

Sounds good. After talking with Jay, we will include the ID in this PR, but it won't be used in this PR. I will make a follow-up PR which makes this change (since it is non-trivial and the PR is already big enough :sweat_smile: )

persistence/lock-keeper-sql/migrations/1_migration.sql
gatoWololo: @jwantz Sounds good. What would be a good alternative type? I used `bytea` since that's what `sqlx` support converting a Rust `Vec<u8>` or `&[u8]` (vector of bytes or "slice" of bytes, respectively). I want to make sure we are able to convert our data into what sqlx likes.

persistence/lock-keeper-sql/migrations/1_migration.sql
gatoWololo: We will do this in a separate PR!

persistence/lock-keeper-sql/migrations/1_migration.sql
gatoWololo: We will do this in a separate PR!

.github/workflows/rust.yml
```diff
@@ -16,14 +16,14 @@ jobs:
...
-            target
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
+#      - name: Cache
+#        uses: actions/cache@v2
+#        with:
+#          path: |
+#            ~/.cargo/registry
+#            ~/.cargo/git
+#            target
+#          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
```
picklenerd, line 26: Just a reminder to remove these.

.github/workflows/tests.yml
```diff
@@ -24,14 +24,14 @@ jobs:
...
-            target
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
+#      - name: Cache
+#        uses: actions/cache@v2
+#        with:
+#          path: |
+#            ~/.cargo/registry
+#            ~/.cargo/git
+#            target
+#          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
```
picklenerd, line 34: Just a reminder to remove these.

dev/config/PostgresDocker.toml
```diff
@@ -0,0 +1,5 @@
+# This file is meant to be used with our `docker compose` instance of postgres!
+uri = 'postgres://test:test_password@postgres/'
```
picklenerd, line 2: Any config files that have different formats depending on whether or not they're running in Docker should go in the subdirectories below `dev/config`

dev/config/PostgresLocal.toml
```diff
@@ -0,0 +1,6 @@
+# This is for running the Server locally with a dockerized instance of the postgres server.
+# Default name and password as set by the docker-compose.yml postgres service.
+uri = 'postgres://test:test_password@localhost/'
```
picklenerd, line 3: See comment above

dev/config/SessionCacheLocal.toml
```diff
@@ -0,0 +1,4 @@
+# This is for running the Server locally with a dockerized instance of the mongodb server.
+mongodb_uri = 'mongodb://localhost:27017'
```
picklenerd, line 2: See comment above

bin/lock-keeper-client-cli/src/cli_command/authenticate.rs
```diff
@@ -30,7 +30,7 @@ impl CliCommand for Authenticate {
     fn parse_command_args(slice: &[&str]) -> Option<Self> {
         match slice {
             [account, password] => Some(Self {
-                account_name: account.parse().ok()?,
+                account_name: AccountName::from(*account),
```
picklenerd, line 33: This should use `FromStr` and do the annoying error handling to return `None`

bin/lock-keeper-client-cli/src/cli_command/register.rs
```diff
@@ -30,7 +30,7 @@ impl CliCommand for Register {
     fn parse_command_args(slice: &[&str]) -> Option<Self> {
         match slice {
             [account, password] => Some(Register {
-                account_name: account.parse().ok()?,
+                account_name: AccountName::from(*account),
```
picklenerd, line 33: This should use `FromStr` too

docker-compose.yml
```diff
@@ -6,10 +6,28 @@ services:
     container_name: mongodb
     restart: always
     ports:
-      - 27017:27017
+      - "27017:27017"
```
picklenerd, line 9: I think this is the standard MongoDB port. It looks like `5432` is the standard port for Postgres so we should use that.

lock-keeper-tests/src/error.rs
```diff
@@ -31,13 +32,11 @@ pub enum LockKeeperTestError {
...
     #[error("LockKeeperMongoDb error: {0:?}")]
-    LockKeeperMongoDb(#[from] lock_keeper_mongodb::error::Error),
-    #[error("LockKeeperMongoDb error: {0:?}")]
     LkSessionMongodb(#[from] lk_session_mongodb::Error),
-    #[error("MongoDB error: {0:?}")]
-    MongoDb(#[from] mongodb::error::Error),
     #[error("RandError: {0:?}")]
     SessionCache(#[from] lock_keeper_key_server::server::session_cache::SessionCacheError),
+    #[error("PostgressError error: {0:?}")]
+    Sqlx(#[from] PostgresError),
```
picklenerd, line 39: I think this should be named something like `LockKeeperPostgres` so we're not assuming `Sqlx`

lock-keeper/src/crypto.rs
```diff
@@ -259,9 +263,11 @@ impl From<Message> for Vec<u8> {
 
 /// Universally unique identifier for a stored secret or signing key.
 /// Wrapped in a `Box` to avoid stack overflows during heavy traffic.
+/// [KeyId]s are created by implementors of our DataStore trait. So we expose
+/// the internal as pub.
 #[derive(Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 #[serde(try_from = "HexBytes", into = "HexBytes")]
-pub struct KeyId(Box<[u8; 32]>);
+pub struct KeyId(pub Box<[u8; 32]>);
```
picklenerd, line 270: I don't think we want people to be able to create these directly.

cc @marsella

picklenerd: I suppose this is still open to the same deserialization workaround though.

lock-keeper/src/types/operations.rs
```diff
@@ -31,23 +31,28 @@ use super::Message;
...
+    // System only.
     Register,
     RemoteGenerateSigningKey,
     RemoteSignBytes,
     RetrieveSecret,
+    // System only.
     RetrieveAuditEvents,
     RetrieveSigningKey,
+    // System only.
     RetrieveStorageKey,
```
picklenerd, line 56: I have system-onlyless encoded a bit more directly in this PR #431 but I like the idea here. I would just suggest that we group them together under a single label.

sqlx-data.json
```diff
@@ -0,0 +1,237 @@
+{
```
picklenerd, line 1: Is there any way to get this file into the `lock-keeper-sql` crate without making it annoying to run things?

docker-compose.yml
gatoWololo: This _is_ for mongdb. I just added the quotation marks. Since my IDE said it was good practice to do that :sweat_smile:

persistence/lock-keeper-sql/create_tables.sql
```diff
@@ -0,0 +1,41 @@
...
+    server_registration BYTEA NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS Secrets
+(
+    secret_id BIGINT GENERATED ALWAYS AS IDENTITY,
+    key_id BYTEA UNIQUE NOT NULL,
+    user_id BYTEA NOT NULL,
+    -- TODO: Use account_id from Accounts table instead of user_id.
+    -- account_id BIGINT NOT NULL,
```
picklenerd, line 17: Why can't we do this now?

lock-keeper/src/crypto.rs
gatoWololo: We can get rid of this `pub` we are no using. It was mostly for consistency with other types we expect the DB implementation to make. We still have a `impl TryFrom<&[u8]> for KeyId` still though. Which allows someone to create a arbitrary `KeyId` still.

docker-compose.yml
picklenerd: Haha I totally missed that!

picklenerd:
Just a general comment but all of the TODOs should have tickets made and linked in the TODO.

persistence/lock-keeper-sql/create_tables.sql
```diff
@@ -0,0 +1,41 @@
...
+    secret BYTEA NOT NULL,
+    secret_type TEXT NOT NULL,
+    -- TODO: Utilize this field.
```
picklenerd, line 20: This needs to be part of the release or it will be a regression on the spec implementation. We can do it in a separate ticket but I think this is high priority.

persistence/lock-keeper-sql/src/api.rs
```diff
@@ -0,0 +1,456 @@
...
+    connection_pool: PgPool,
+}
+
+impl Debug for PostgresDBInner {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "PostgresDB {{ url: {}}}", self.config.uri())
+    }
+}
+
+impl PostgresDBInner {
```
picklenerd, line 134: Do we need this extra layer? I know this is sort of the pattern we already have but I'm not sure how valuable it is. It looks like the `PgPool` is already behind an `Arc` a few layers into the type so we could just `Arc` the config and have them be fields on the main object.

persistence/lock-keeper-sql/src/api.rs
```diff
@@ -0,0 +1,456 @@
...
+        Ok(())
+    }
+
+    // SELECT key_id, account_name, action, event_status, timestamp
+    // FROM AuditEvents
+    // WHERE
+    // timestamp >= after_date AND
+    // timestamp <= before_date AND
+    // key_id in (key_id1, key_id2, ...) AND
+    // action in (actions1, action2...);
```
picklenerd, line 192: I like this for readability but I'm afraid it will become outdated.

picklenerd: Maybe we could debug log the full query at the end?

persistence/lock-keeper-sql/src/api.rs
```diff
@@ -0,0 +1,456 @@
...
+        &self,
+        user_id: &UserId,
+        key_id: &KeyId,
+        filter: SecretFilter,
+    ) -> Result<StoredSecret, PostgresError> {
+        info!("Fetching user secret.");
+
+        // We use the LIKE operator to support whether filter.secret_type is present or
+        // not. In case it is not, we use a wildcard match for the secret_type
+        // column.
```
picklenerd, line 293: Clever!

persistence/lock-keeper-sql/src/api.rs
```diff
@@ -0,0 +1,456 @@
...
+                    key_id.as_bytes()
+                )
+                .fetch_one(&self.connection_pool)
+                .await?;
+
+                match key_found.count {
+                    // The key doesn't even exist.
+                    0 => Err(PostgresError::NoEntry),
+                    // The key exists but the secret_type or user_id were incorrect.
+                    // TODO: Should we record an audit event here? Or let caller handle the case.
```
picklenerd, line 323: This does feel audit-worthy to me given that it indicates that someone knows the key id for someone else's key.

cc @jwantz @indomitableSwan @marsella

persistence/lock-keeper-sql/src/api.rs
```diff
@@ -0,0 +1,456 @@
...
+
+    #[instrument(skip_all, err(Debug), fields(user_id=?user_id, account_name=?account_name))]
+    pub(crate) async fn create_user(
+        &self,
+        user_id: &UserId,
+        account_name: &AccountName,
+        server_registration: &ServerRegistration<OpaqueCipherSuite>,
+    ) -> Result<User, PostgresError> {
+        info!("Creating new user.");
+        let serialized = rmp_serde::to_vec(server_registration)?;
```
picklenerd, line 340: We're using `bincode` for this already in other places.

persistence/lock-keeper-sql/src/api.rs
```diff
@@ -0,0 +1,456 @@
...
+        Ok(User {
+            user_id: user_id.clone(),
+            account_name: account_name.clone(),
+            storage_key: None,
+            server_registration: server_registration.clone(),
+        })
+    }
+
+    #[instrument(skip_all, err(Debug), fields(account_name=?account_name))]
+    pub(crate) async fn find_user(
```
picklenerd, line 361: I think we should rename this to `find_account` and the `User` type to `Account` just to make it match our database terminology.

persistence/lock-keeper-sql/src/api.rs
```diff
@@ -0,0 +1,456 @@
...
+        Ok(user)
+    }
+
+    #[instrument(skip_all, err(Debug), fields(user_id=?user_id))]
+    pub(crate) async fn delete_user(&self, user_id: &UserId) -> Result<(), PostgresError> {
+        info!("Deleting user.");
+
+        // Delete the entry and
+        let result = sqlx::query!(
+            r#"WITH deleted AS (DELETE FROM Accounts WHERE user_id=$1 RETURNING *) SELECT count(*) AS "count!" FROM deleted"#,
```
picklenerd, line 407: Does this do that cascading delete thing?

persistence/lock-keeper-sql/src/api.rs
```diff
@@ -0,0 +1,456 @@
...
+        )
+        .execute(&self.connection_pool)
+        .await?;
+
+        Ok(())
+    }
+}
+
+/// Create a SQL query list of the form (val1, val2, ...). This function does
+/// not handle empty iterators. The iterator should have at least one element.
```
picklenerd, line 444: I think this should return an error if this is the case.

persistence/lock-keeper-sql/src/config.rs
```diff
@@ -0,0 +1,38 @@
...
+use serde::{Deserialize, Serialize};
+use std::{path::Path, str::FromStr};
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(deny_unknown_fields, rename_all = "snake_case")]
+pub struct Config {
+    /// URI of the database. Should be in the form
+    /// "postgres:username:password@address". It should not include the
+    /// name of the database. This does not include the name of the database.
+    pub(crate) uri: String,
```
picklenerd, line 11: @jwantz is it ok to keep the db credentials in memory?

persistence/lock-keeper-sql/src/types.rs
```diff
@@ -0,0 +1,132 @@
...
+        let secret_id: Option<KeyId> = match event.key_id {
+            None => None,
+            Some(key_id) => {
+                let key_id =
+                    TryFrom::try_from(key_id.as_slice()).map_err(|_| AuditEventConversionError)?;
+                Some(key_id)
+            }
+        };
+        let status =
+            EventStatus::from_str(&event.event_status).map_err(|_| AuditEventConversionError)?;
```
picklenerd, line 118: These errors should probably include details about which field failed to convert.

lock-keeper/src/types/database/user.rs
```diff
@@ -11,22 +11,34 @@ use rand::{CryptoRng, Rng, RngCore};
...
 }
 
+/// Manual implement Debug to avoid printing storage key or server_registration.
+impl Debug for User {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        writeln!(
+            f,
+            "User {{user_id: {:?}, account_name: {:?}}}",
+            self.user_id, self.account_name
+        )?;
```
picklenerd, line 36: We could use `f.debug_struct` here to keep the format nice and standard:

https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct

lock-keeper/src/types/operations.rs
gatoWololo: Neat, I will undo this change then. And we will use the code from your PR.

.github/workflows/rust.yml
gatoWololo: done

.github/workflows/tests.yml
gatoWololo: done

persistence/lock-keeper-sql/create_tables.sql
gatoWololo: Our current DataStore API for creating and retrieving secrets takes a `user_id`. We would need to modify those to accept `account_id` instead. Similarly, our server code and tests would need to be modified to have the `account_id` for a given user. This is a nontrivial change. So I wanted to do it in the follow up PR.

persistence/lock-keeper-sql/src/api.rs
gatoWololo: Done.

persistence/lock-keeper-sql/create_tables.sql
gatoWololo: I implemented the release field. Using the account_id and changing secret_type to be a different type will be done in the follow up PR.

persistence/lock-keeper-sql/src/api.rs
gatoWololo: Extra layer removed!

persistence/lock-keeper-sql/src/api.rs
gatoWololo: Not yet. This will be handled in #435

persistence/lock-keeper-sql/src/api.rs
gatoWololo: Thanks. I couldn't figure out how we were doing it. _changed_

sqlx-data.json
gatoWololo: I don't think so... It looks like it has to exist in the root directory of the workspace.

persistence/lock-keeper-sql/create_tables.sql
picklenerd: Can we still have the field in the table?

lock-keeper/src/crypto.rs
gatoWololo: Got rid of `pub`.

persistence/lock-keeper-sql/create_tables.sql
gatoWololo: We will do this in the follow up PR.

persistence/lock-keeper-sql/src/api.rs
gatoWololo: We will handle this in the follow up PR.

persistence/lock-keeper-sql/src/config.rs
gatoWololo: Added to #435. We will ask Jay during our meeting if its okay.

persistence/lock-keeper-sql/src/types.rs
```diff
@@ -0,0 +1,135 @@
...
+            user_id,
+            secret_type: secret.secret_type,
+            bytes: secret.secret,
+            retrieved: secret.retrieved,
+        })
+    }
+}
+
+impl TryFrom<AccountDB> for Account {
+    type Error = bincode::Error;
```
picklenerd, line 73: I think I would prefer it if these errors were pre-wrapped in the local error type. Then we could change the serialization library without changing the impl definition.

persistence/lock-keeper-sql/src/types.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub(crate) account_name: String,
+    pub(crate) key_id: Option<Vec<u8>>,
+    pub(crate) request_id: Uuid,
+    pub(crate) action: String,
+    pub(crate) event_status: String,
+    pub(crate) timestamp: OffsetDateTime,
+}
+
+impl TryFrom<SecretDB> for StoredSecret {
+    type Error = TryFromSliceError;
```
picklenerd, line 57: See comment about `bincode:Error`

picklenerd (approved):

PR #408, opened 2022-12-19
Replace hashmap session cache with MongoDB #351 #352

### LINKED ISSUE #351: Application Sessions for Key Server
Please refer to #364  for a high level overview of Application Sessions (or LockKeeper Account Sessions). It is acceptable for the proposed terminology (e.g., Application Session) to be refined/reworked during the design and development of Session Management.

Tasks:

1. Document a design for application session management on the Key Server, to include, but not limited to:
- Data elements in the object model (for example: session_id, user_id, creation_timestamp, session_key_id, nonce)
- Session lifetime (and potential state changes)
- Persistence (if any)
- Expiration handling
- Documentation of concerns and future thoughts
2. Peer review of design, including answers to the following questions:
- How far do we take the approach in the current release (e.g., phased approach)?
- Are there any security concerns or constraints to be addressed?
- Do we have concerns around caching and/or availability at this point?
3. Implementation of the proposed design on the Key Server
4. Appropriate unit tests that include expiration and clean up of sessions

### LINKED ISSUE #352: Application Sessions for LK Client
Please refer to #364 for a high level overview of Application Sessions (or LockKeeper Account Sessions). It is acceptable for the proposed terminology (e.g., Application Session) to be refined/reworked during the design and development of Session Management.

Tasks:

1. Document a design for application session management on the LockKeeperClient (this task will be covered in conjunction with the Server design on ticket #351)
2. Peer review of design, including answers to the following questions (again this will be covered in conjunction with the Server design)
3. Implementation of the proposed design on the LockKeeperClient
4. Appropriate end-to-end tests that include expiration and clean up of sessions

vidyaap:
Closes #351 
Closes #352

Other changes:
- A couple of file/struct names to not include "key" in "session key cache" anymore
- Moved session cache unit tests to their own integration test module
- Encryption of session keys

Specific comments:
- Flagging for @marsella:
    - I added encryption of session keys like we talked about - from/to bytes taken from storage keys, encryption using remote storage key in the same way as signing keys.
    - I’m using the same ID generation for session IDs as we have for user IDs so that our ID implementations are consistent and for serialization - are there any issues with this?
- General:
    - I’m using a `delete_many` for the case where we’re overwriting someone’s session because they logged in while they already had a session. Since we don’t have device IDs yet, I can’t use a `delete_one` to delete the one session that is for this user on a specific device, but is there a better way to handle this if we don’t want device IDs yet?
    - What should go in the Zeroize impl now?
    - For deleting a session on logout, I’m assuming that if the user passes our auth checks, then the session ID they passed is legit, and I’m only passing that to delete. Should I still pass the user ID to delete anyway?

*I also still need to add some logging to the session cache - I'm debating the level of detail needed.


lock-keeper-client/src/api/authenticate.rs
```diff
@@ -96,9 +85,12 @@ async fn authenticate_finish(
 
     let server_finish: server::AuthenticateFinish = channel.receive().await?;
 
-    if server_finish.success {
-        Ok(client_login_finish_result)
-    } else {
-        Err(LockKeeperClientError::ServerReturnedFailure)
-    }
+    let session_key: OpaqueSessionKey = client_login_finish_result.session_key.try_into()?;
+    let master_key = MasterKey::derive_master_key(client_login_finish_result.export_key)?;
```
marsella, line 89: Good containment.

lock-keeper/src/crypto.rs
```diff
@@ -66,6 +66,57 @@ impl OpaqueSessionKey {
...
+            .ok_or(CryptoError::ConversionError)?
+            .to_vec();
+
+        Ok(Self(key_bytes.try_into()?))
+    }
+}
+
+impl Encrypted<OpaqueSessionKey> {
+    /// Decrypt a session key server-side.
+    pub fn decrypt_session_key_by_server(
```
marsella, line 113: suggestion: You can call this just `decrypt_session_key`. The `_by_server` postfix on the signing key is there because there are multiple types of key that can decrypt that type. In this case, only the server has to handle it, so we don't necessarily have to specify which decryption function is which.

lock-keeper/src/crypto.rs
```diff
@@ -66,6 +66,57 @@ impl OpaqueSessionKey {
     ) -> Result<Encrypted<Message>, CryptoError> {
         Encrypted::encrypt(rng, &self.0, message, &AssociatedData::new())
     }
+
+    fn context(&self) -> &AssociatedData {
+        &self.0.context
+    }
+
+    pub(crate) fn domain_separator() -> &'static str {
+        "opaque session key"
```
marsella, line 75: nit: Change this to `OPAQUE-derived Lock Keeper session key`, and then up in L49, replace the hard-coded string with a call to this function.

lock-keeper/src/crypto.rs
```diff
@@ -66,6 +66,57 @@ impl OpaqueSessionKey {
...
+        OpaqueSessionKey::domain_separator()
+            .as_bytes()
+            .iter()
+            .copied()
+            .chain::<Vec<u8>>(key.0.to_owned().into())
+            .collect()
+    }
+}
+
+impl TryFrom<Vec<u8>> for OpaqueSessionKey {
```
marsella, line 90: todo: Add a unit test that these methods are complements. See any of the `X_to_vec_u8_conversion_works()` tests for an example.

lock-keeper/src/crypto/storage_key.rs
```diff
@@ -59,6 +59,17 @@ impl RemoteStorageKey {
         )
         .map_err(LockKeeperError::Crypto)
     }
+
+    /// Encrypt the given [`OpaqueSessionKey`] under the
+    /// [`RemoteStorageKey`] using an AEAD scheme.
+    pub fn encrypt_session_key(
```
marsella, line 65: todo: add unit test that `decrypt(encrypt(session_key))` works. See `remote_storage_key_works` test as an example. 

It might be easier to add it in the `crypto` module -- whichever you prefer is fine.

lock-keeper/src/types/operations.rs
```diff
@@ -62,19 +67,55 @@ pub trait ConvertMessage: Sized + for<'a> Deserialize<'a> + Serialize {
...
+    }
+}
+
+impl TryFrom<HexBytes> for SessionId {
+    type Error = LockKeeperError;
+
+    fn try_from(bytes: HexBytes) -> Result<Self, Self::Error> {
+        Ok(SessionId(Box::new(bytes.try_into()?)))
+    }
+}
```
marsella, line 97: suggestion: add a test that these work together.

marsella (dismissed):
This looks good to me, but the crypto module needs some tests for confidence!

lock-keeper/src/crypto/storage_key.rs
vidyaap: Kept it in storage_key!

lock-keeper/src/types/operations.rs
vidyaap: Added a similar test to the "to_vec_u8_works" tests in types/operations.

lock-keeper-key-server/src/server.rs
```diff
@@ -165,6 +165,11 @@ impl<DB: DataStore> LockKeeperRpc for LockKeeperKeyServer<DB> {
             .ok_or(lock_keeper::LockKeeperError::MetadataNotFound)?
             .try_into()?;
 
+        let session_id = metadata
+            .session_id()
+            .ok_or(LockKeeperServerError::InvalidAccount)?
```
picklenerd, line 170: Should this be a session key error?

lock-keeper-key-server/src/server.rs
```diff
@@ -303,11 +310,17 @@ impl<DB: DataStore> LockKeeperKeyServer<DB> {
             .user_id()
             .ok_or(LockKeeperServerError::InvalidAccount)?
             .clone();
+        let session_id = channel
+            .metadata()
+            .session_id()
+            .ok_or(LockKeeperServerError::InvalidAccount)?
```
picklenerd, line 316: Should this be a session key error?

lock-keeper-key-server/src/server/session_cache.rs
```diff
@@ -0,0 +1,97 @@
...
+
+/// Cache holding our sessions, per user, after authentication. Maps
+/// [`UserId`]s to [`Session`]s. Sessions should be tagged with a
+/// timestamp. A session is considered invalid after the expiration time has
+/// elapsed.
+#[async_trait]
+pub trait SessionCache: Send + Sync {
+    /// Store a newly created session for the specified user. The previous
+    /// session for that user should be overwritten.
+    async fn store_session(&self, session: Session) -> Result<(), SessionCacheError>;
```
picklenerd, line 65: Is there a plan for this method? It looks like it's only used by `create_session` right now.

picklenerd:
> I’m using a delete_many for the case where we’re overwriting someone’s session because they logged in while they already had a session. Since we don’t have device IDs yet, I can’t use a delete_one to delete the one session that is for this user on a specific device, but is there a better way to handle this if we don’t want device IDs yet?

I wouldn't even think about devices yet. That's going to be a pretty significant change.

I like delete_many for this regardless. I could see some sequence of events happening where there are two sessions stored for a single user_id, so this would still delete all of the sessions for that user. I don't think it's possible with the current code but someone could write that bug later.

picklenerd:
> For deleting a session on logout, I’m assuming that if the user passes our auth checks, then the session ID they passed is legit, and I’m only passing that to delete. Should I still pass the user ID to delete anyway?

I think this is good. If someone has my session ID but not my user ID, I probably want my session deleted anyway. Not like they could log back in.

picklenerd:
> What should go in the Zeroize impl now?

Which one?

lock-keeper-key-server/src/server.rs
vidyaap: As in a new LKServerError specifically for missing session IDs? I think I wasn't sure whether I should specify since missing user IDs also say invalid account?

lock-keeper-key-server/src/server/session_cache.rs
vidyaap: I didn't want to have to create a session outside of the session code to pass into `create_session` so I have create_session to create and just leave storing to the implementor.  But I think we also only create a session in one place so if we'd rather have one function for storing a session, I can just make and pass in the session after all.

vidyaap:
> > What should go in the Zeroize impl now?
> 
> Which one?

There was a Drop impl on the whole hashmap cache before that seemed to zeroize every value in the cache. I wasn't sure if we still wanted to do that for the DB table now?

picklenerd:
> There was a Drop impl on the whole hashmap cache before that seemed to zeroize every value in the cache. I wasn't sure if we still wanted to do that for the DB table now?

If mongo has an easy way to do that, I guess you could. I don't think it matters. The responsibility to zeroize the database would be on the DB administrator for an external database. We really shouldn't be messing with the database at all from the Rust code.

lock-keeper-key-server/src/server.rs
picklenerd: I think it's ok in this case since session IDs aren't a fundamental part of the account. You can use `invalid user id` to brute force user ids, but it wouldn't accomplish much if you did the same thing for a session id since it'll just expire anyway and you still need the user id.

lock-keeper-key-server/src/server/session_cache.rs
picklenerd: Gotcha. I like this approach.

picklenerd (approved):
LGTM!

@marsella I checked the updates based on your comments and resolved them since you're already out for the holidays.

lock-keeper/src/crypto.rs
```diff
@@ -436,4 +487,17 @@ mod test {
 
         Ok(())
     }
+
+    #[test]
+    fn session_key_to_vec_u8_conversion_works() -> Result<(), LockKeeperError> {
+        for _ in 0..1000 {
+            let session_key = OpaqueSessionKey::try_from(GenericArray::from([42; 64]))?;
```
marsella, line 494: nit: This should be a randomly-generated array. As-is, we're testing the same static array 1000 times, which doesn't give us any benefit over testing it once. Can you access the `create_test_session_key` method here?

lock-keeper/src/crypto/storage_key.rs
```diff
@@ -321,4 +333,16 @@ pub(super) mod test {
         assert_eq!(result, signing_key);
         Ok(())
     }
+
+    #[test]
+    fn session_key_encryption_works() -> Result<(), LockKeeperError> {
+        let mut rng = rand::thread_rng();
+        let encryption_key = RemoteStorageKey::generate(&mut rng);
+        let session_key = OpaqueSessionKey::try_from(GenericArray::from([42; 64]))?;
```
marsella, line 341: nit: Would prefer a random key here, as with `create_test_session_key`.

marsella (commented):
Thanks for following up (and for reviewing the changes, @picklenerd). I do have two small comments about how opaque session keys are generated in tests. I made #415 so when folks are back online, we can address them.

PR #410, opened 2022-12-20
Rename "mutual auth" to "client auth"

picklenerd:
TLS has a feature called "client authentication" that allows the server to verify the certificates of connecting clients. We've been calling this feature "mutual authentication" in code since we authenticate both ways, but that's more of a description of the full TLS setup. The actual code implementation that currently uses the name "mutual auth" is really just doing "client auth".

This PR renames everything referencing "mutual auth" to reference "client auth" instead. This change brings our codebase's terminology in line with the terminology used in `rustls` and should make it easier for newer developers to troubleshoot TLS problems.


gatoWololo (approved):
Looks good! Couldn't find any instances of "mutual" on the repo :mag:

picklenerd:
I'm going to leave this open for a bit in case anyone disagrees with this change.

marsella:
I didn't review but I'm not morally opposed to the name change.

PR #412, opened 2022-12-21
334 add cli scripting

### LINKED ISSUE #334: Add scripting support to client CLI
The client CLI should have basic scripting support that would allow multiple commands to be run in sequence. The syntax should be the same as typing commands manually, with a single command on each line.

- [x] Add support for parsing a script into a sequence of commands and executing those commands.
- [x] Add a script validator that can be run against scripts without running the script itself.
- [x] Add command line options to accept scripts as strings or file names.
- [x] Add command line option to repeat scripts a number of times or indefinitely with a custom delay between repeats.
- [ ] ~Commands that support optional key names should require a name in scripts.~ Removing in favor of the ability to repeat commands
- [ ] ~(optional) Add support for a "script repository" that contains script names mapped to script contents on the local file system. Allow these scripts to be called by name from the CLI.~

picklenerd:
Closes #334


bin/lock-keeper-client-cli/src/scripting.rs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        Ok(Self(commands))
+    }
+}
+
+fn parse_repeats<'a>(
+    num_repeats: usize,
+    input: &mut impl Iterator<Item = &'a str>,
+) -> anyhow::Result<Vec<Box<dyn CliCommand>>> {
```
gatoWololo, line 65: We have a `DynCommand` alias that I have been using to avoid writing out the full boxed typed.

gatoWololo (approved):
Looks good! I just have one smol comment.

PR #414, opened 2022-12-27
Add request ID to audit event and client response #411

### LINKED ISSUE #411: Add request ID to audit event and its filters
Now that each request has a request ID that is returned to the client, we can also attach it to audit events.  This would then allow us to filter audit events by request.

vidyaap:
Closes #411 

Other changes:
- Change LockKeeperResponse to return the entire result so we can return it in the error case as well
- Generate request IDs client-side instead of server-side
- A request ID spans an entire client operation, not each individual request to the server that the operation might take

This ended up touching more files than expected, so I may try to split out some of the refactoring into another PR if people don't review before I'm back!

### DIFF
--- a/lock-keeper-client/src/api.rs
+++ b/lock-keeper-client/src/api.rs
@@ -17,7 +17,8 @@ mod retrieve;
 mod retrieve_audit_events;
 
 use crate::{
-    client::Password, config::Config, LockKeeperClient, LockKeeperClientError, LockKeeperResponse,
+    client::Password, config::Config, response::Metadata, LockKeeperClient, LockKeeperClientError,
+    LockKeeperResponse,
 };
 use lock_keeper::{
     crypto::{Export, Import, KeyId, Secret, Signable, Signature},
@@ -31,7 +32,7 @@ use rand::{rngs::StdRng, SeedableRng};
 use serde::{Deserialize, Serialize};
 use std::sync::Arc;
 use tokio::sync::Mutex;
-use tracing::error;
+use uuid::Uuid;
 
 pub use self::{
     generate_secret::GenerateResult, remote_generate_signing_key::RemoteGenerateResult,
@@ -64,18 +65,13 @@ impl LockKeeperClient {
     }
 
     /// Expire the current session and session key for this user.
-    pub async fn logout(&self) -> Result<LockKeeperResponse<()>, LockKeeperClientError> {
+    pub async fn logout(&self) -> LockKeeperResponse<()> {
         // Create channel to send messages to server
-        let metadata = self.create_metadata(ClientAction::Logout);
-        let client_channel = Self::create_authenticated_channel(
-            &mut self.tonic_client(),
-            &metadata,
-            self.session_key().clone(),
-            self.rng.clone(),
-        )
-        .await?;
-
-        self.handle_logout(client_channel).await
+        let request_id = Uuid::new_v4();
+        LockKeeperResponse {
+            result: self.handle_logout(request_id).await,
+            metadata: Some(Metadata { request_id }),
+        }
     }
 
     /// Authenticate to the Lock Keeper key server as a previously registered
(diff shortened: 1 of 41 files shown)


lock-keeper-client/src/api.rs
```diff
@@ -102,65 +101,69 @@ impl LockKeeperClient {
...
+        Ok(())
     }
 
     /// Export an arbitrary key from the key servers.
     ///
     /// Calling this function on a signing key will generate an error.
     /// Output: If successful, returns the requested key material in byte form.
-    pub async fn export_secret(
+    pub async fn export_secret(&self, key_id: &KeyId) -> LockKeeperResponse<Export> {
+        // Create channel: this will internally be a `retrieve` channel
```
gatoWololo, line 153: Old, left over comment.

lock-keeper-client/src/api.rs
```diff
@@ -169,29 +172,38 @@ impl LockKeeperClient {
...
+        Ok(Export::from(local_storage.material))
     }
 
     /// Export signing key pair material from the key servers.
     ///
     /// Calling this function on an arbitrary key will generated an error.
     /// Output: If successful, returns the requested key material in byte form.
-    pub async fn export_signing_key(
+    pub async fn export_signing_key(&self, key_id: &KeyId) -> LockKeeperResponse<Export> {
+        // Create channel: this will internally be a `retrieve` channel
```
gatoWololo, line 193: Same here.

lock-keeper-client/src/api.rs
```diff
@@ -200,25 +212,29 @@ impl LockKeeperClient {
...
             .await?;
 
-        let local_storage = data.ok_or(LockKeeperClientError::ExportFailed)?;
-        let exported_signing_key = Export::from(local_storage.material);
-
-        Ok(LockKeeperResponse {
-            data: exported_signing_key,
-            metadata,
-        })
+        let local_storage = option_local_storage.ok_or(LockKeeperClientError::ExportFailed)?;
```
gatoWololo, line 219: We probably want to inline this optional_* variable and just call ok_or as part of our call chain. (Same for the other case above)

lock-keeper-client/src/api.rs
```diff
@@ -169,29 +172,38 @@ impl LockKeeperClient {
...
             .await?;
 
-        let local_storage = data.ok_or(LockKeeperClientError::ExportFailed)?;
-        let exported_secret = Export::from(local_storage.material);
-
-        Ok(LockKeeperResponse {
-            data: exported_secret,
-            metadata,
-        })
+        let local_storage = option_local_storage.ok_or(LockKeeperClientError::ExportFailed)?;
```
gatoWololo, line 184: inline var

lock-keeper-tests/src/test_suites/end_to_end/operations.rs
```diff
@@ -54,62 +55,53 @@ pub(crate) async fn check_audit_events(
...
 
-    // Get the fourth last event, the last 3 are for retrieving audit logs and
-    // authenticating
-    let fourth_last = audit_event_log
-        .len()
-        .checked_sub(6)
-        .map(|i| &audit_event_log[i]);
-    assert!(fourth_last.is_some());
+    // Get the last event in the list of events
+    let last_event = audit_event_log.last().unwrap();
```
gatoWololo, line 77: So much nicer!!!

gatoWololo (approved):
Looks good! I only have very minor comments. I might fix these tomorrow and merge it (it would be nice to have these events merged so I can rebase on them). I'm currently running into issues in the integration tests due to audit event logs. Since the ordering isn't what the tests expect.

gatoWololo:
I will merge this as I really want to rebase my SQL things onto `develop`, including this change! I made some minor changes based on my review. @vidyaap I hope that okay :sweat_smile:

PR #421, opened 2023-01-03
Updated spans to manually skip fields #416

### LINKED ISSUE #416: Logging not capturing span variables.
I realize now I was wrong about how the `instrument` macro works. We are not recording variables as part of the `span` as previously thought. For example:
```rust
#[instrument(skip_all, err(Debug), fields(account_name))]
async fn operation(
        self,
        channel: &mut ServerChannel<Unauthenticated>,
        context: &mut Context<DB>,
    ) -> Result<(), LockKeeperServerError>
```
By default, `instrument` tries to record all arguments to this function. So we use the `skip_all` option to avoid this. Then we manually opt in to recording the `account_name` by specifying `fields(account_name)`.

This does not work as expected. The `fields(account_name)` is incomplete and will not record the value of `account_name`; this merely declares the field for our logger.

We either need to call our `record_field` function at the very beginning, e.g:
```
record_field("account_name", account-name);
```
Or write the instrument macro as:
```rust
#[instrument(skip_all, err(Debug), fields(account_name=?account_name))]
```
The `?` tells `tracing` to use the `Debug` impl of this type. This can get a little hairy:
```rust
#[instrument(skip_all, err(Debug), fields(account_name=?account_name, event_type=?event_type, options=?options))]
```

The other options is to stop using the `skip_all` macro manually list the arguments we want to skip. We just have to be careful to skip every field that should not be included.

To close this ticket:
- [x] Ensure all instances of `instrument` macro are correctly written.
- [x] Update README about logging to reflect the correct way to do it.

picklenerd:
This PR updates `#[instrument]` macros for `tracing` to manually skip function arguments in cases where a subset of arguments should be included in the span.

In cases where all arguments are skipped, we continue to use `skip_all`.

Closes #416

### DIFF
--- a/bin/key-server-cli/src/main.rs
+++ b/bin/key-server-cli/src/main.rs
@@ -46,15 +46,14 @@
 //! Please follow the following template for instrumenting your function with
 //! spans:
 //! ```text
-//! #[instrument(skip_all, err(Debug), fields(arg1, arg2))]
-//! fn foo(arg1: _, arg2: _) -> _ {}
+//! #[instrument(skip(sensitive_arg), err(Debug))]
+//! fn foo(arg1: _, arg2: _, sensitive_arg SuperSecret) -> _ {}
 //! ```
 //! By default, `instrument` will create a new span where every argument to a
 //! function is a span field. *This is not desirable for our key server as we
-//! want to avoid logging sensitive data!* So *we always use the `skip_all`
-//! option* to avoid automatically logging all arguments. Instead you can use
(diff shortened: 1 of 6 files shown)


gatoWololo:
The changes so far look good. I did find a few other instances:
image

Any reasons these weren't also updated?

picklenerd:
> Any reasons these weren't also updated?

This is the case I talked about in the PR description where we're skipping all of the arguments so I left the concise definition in place.

gatoWololo (approved):
Sorry forgot to approve last Friday!

PR #422, opened 2023-01-03
Update protobuf action and add token #420

### LINKED ISSUE #420: CI rate limit errors
CI has occasionally been giving rate limit errors lately. We should make sure we're authenticating and deal with any billing issues that come along with that.

```
Install Protoc
Run arduino/setup-protoc@v1
  with:
    version: 3.x
    include-pre-releases: false
Error: API rate limit exceeded for 104.45.204.224. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```

picklenerd:
This PR updates the GitHub action for installing protobuf to the latest version and adds `repo-token` to the action which should solve the issue.

See: https://github.com/actions/runner-images/issues/602

Closes #420

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -26,9 +26,10 @@ jobs:
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
 
       - name: Install Protoc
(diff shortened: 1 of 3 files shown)


gatoWololo (approved):
Looks good!

PR #428, opened 2023-01-04
Added some CLI scripts for quick testing

picklenerd:
This PR adds a few CLI scripts to the `dev` directory that can be used for quick testing. I also added some `cargo make` tests for them so they can be run without a giant command.


vidyaap (approved):
LGTM!

PR #429, opened 2023-01-05
Manually close channel at the end of handle operation #386

### LINKED ISSUE #386: Hyper Error: Broken Pipe
We sometimes see an error from hyper when sending the final response from the server back to the client. This error is easy for me to reproduce locally:
1) Using the client CLI do a `authentication` request.
2) At the end of the server operation the logs will show:
```
  2022-12-01T18:58:16.237349Z ERROR lock_keeper_key_server::server::service: Hyper(hyper::Error(Io, Kind(BrokenPipe)))
    at lock-keeper-key-server/src/server/service.rs:87
```

This is a sorta known issue. The server does a:
```
// Give the client a moment to receive the error before dropping the channel
thread::sleep(Duration::from_millis(100));
```
At the end of a protocol which returned an error.

Further investigation is needed to see why this happens and get rid of the ugly sleep.

vidyaap:
Closes #386 

It seems like the broken pipe error was happening when the client drops its receiver at the end of a protocol, so I added a manual call to `closed` to the server's sender which tells it that the receivers have dropped.
I'd only gotten the error when running the server locally - I don't get it anymore, but I'd like if folks tested this out wherever they were seeing the error before too.

Also includes a quick doc string for RequestMetadata.

### DIFF
--- a/lock-keeper-key-server/src/server.rs
+++ b/lock-keeper-key-server/src/server.rs
@@ -144,7 +144,7 @@ impl<DB: DataStore> LockKeeperRpc for LockKeeperKeyServer<DB> {
     async fn get_user_id(
         &self,
         request: Request<tonic::Streaming<Message>>,
-    ) -> Result<Response<Self::GenerateSecretStream>, Status> {
+    ) -> Result<Response<Self::GetUserIdStream>, Status> {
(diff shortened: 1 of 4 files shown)


gatoWololo (approved):
Approving. I will make a separate ticket for the issue still lingering.

PR #431, opened 2023-01-05
Add audit event support to client CLI #272

### LINKED ISSUE #272: Add audit events command to interactive client
The interactive client doesn't currently support retrieving audit events even though the client does. This command should be added.

- [x] `audit` should retrieve all audit events with no filters
- [x] `audit type:["all" | "system-only" | "key-only"]` should specify the given event type
- [x] ~`audit before:[yyyy-mm-dd]` should use a before-date filter with midnight as the time~
- [x] ~`audit after:[yyyy-mm-dd]` should use an after-date filter with midnight as the time~
- [x] `audit keys:[key_name_1, key_name_2, ...]` should use a key_id filter. Keys should be specified by name and key_ids looked up by `lkic`
- [x] `audit request-id:[uuid]`
- [x] All of the above options should be combinable
- [x] Replace existing `println!("Not implemented")` block with a proper command
- [x] Add documentation to the `HELP_TEXT` constant and README
- [x] Add example to README

picklenerd:
Client CLI can now retrieve audit event logs. I implemented all of the filters except for the date filters since they depend on `mongodb` types which will be replaced soon.

Other changes:
- Removed help text from README. This will end up being outdated so I don't think it's worth trying to maintain.
- Remote signing now accepts all remaining text after the key name as the data to be signed. This lets you use spaces in your data string.

I added a `merge` method to `AuditEventOptions` which didn't end up being used due to the `CliComand` trait definition for `parse_command_args`. I still plan to use this method after more refactoring so I'd like to leave this in.

Closes #272

### DIFF
--- a/bin/lock-keeper-client-cli/src/cli_command/get_audit_events.rs
+++ b/bin/lock-keeper-client-cli/src/cli_command/get_audit_events.rs
@@ -1,29 +1,111 @@
-use std::time::Duration;
+use std::{
+    str::FromStr,
+    time::{Duration, SystemTime},
+};
 
 use crate::{cli_command::CliCommand, state::State};
-use anyhow::Error;
+use anyhow::{anyhow, Error};
 use async_trait::async_trait;
+use lock_keeper::types::audit_event::{AuditEventOptions, EventType};
+use lock_keeper_client::LockKeeperClient;
+use uuid::Uuid;
 
 #[derive(Debug)]
-pub struct GetAuditEvents {}
+pub struct GetAuditEvents {
+    event_type: EventType,
+    key_names: Option<Vec<String>>,
+    request_id: Option<Uuid>,
+}
 
 #[async_trait]
 impl CliCommand for GetAuditEvents {
-    async fn execute(self: Box<Self>, _state: &mut State) -> Result<Duration, Error> {
-        println!("{:?}: Not implemented", GetAuditEvents {});
-        // Return zero duration since this command is not implemented
-        Ok(Duration::ZERO)
+    async fn execute(self: Box<Self>, state: &mut State) -> Result<Duration, Error> {
+        let key_ids = match self.key_names {
+            Some(key_names) => {
+                let mut key_ids = Vec::new();
+                for key_name in key_names {
+                    key_ids.push(state.get_key_id(&key_name)?.key_id.clone());
+                }
+
+                Some(key_ids)
+            }
+            None => None,
+        };
+        let options = AuditEventOptions {
+            key_ids,
+            request_id: self.request_id,
+            ..Default::default()
+        };
+
+        let credentials = state.get_credentials()?;
+
+        // Authenticate user to the key server
+        let lock_keeper_client = LockKeeperClient::authenticated_client(
+            &credentials.account_name,
+            &credentials.password,
+            &state.config,
+        )
+        .await
+        .result?;
+
+        let now = SystemTime::now();
+        // If successful, proceed to generate a secret with the established session
+        let audit_event_log = lock_keeper_client
+            .retrieve_audit_event_log(self.event_type, options)
+            .await
+            .result?;
+        let elapsed = now.elapsed()?;
+
+        println!("Audit Events:");
+        for event in audit_event_log {
+            println!("----------------------------------");
+            println!("{event}");
+        }
+        println!("----------------------------------");
+
+        Ok(elapsed)
     }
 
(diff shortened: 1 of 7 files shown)


bin/lock-keeper-client-cli/src/cli_command/get_audit_events.rs
```diff
@@ -1,29 +1,117 @@
...
+        for event in audit_event_log {
+            println!("----------------------------------");
+            println!("Request ID: {}", event.request_id());
+            if let Some(key_id) = event.key_id() {
+                println!("{key_id:?}");
+            }
+            println!("{}", event.date());
+            println!("{}", event.action());
+            println!("{}", event.status());
+        }
```
vidyaap, line 69: How do you feel about using this format for the Display implementation of AuditEvent, and then printing the event here directly? Right now, its Display impl is a full sentence, but I think this is easier to read/parse.

bin/lock-keeper-client-cli/src/cli_command/get_audit_events.rs
```diff
@@ -37,6 +125,45 @@ impl CliCommand for GetAuditEvents {
...
+            let key_names: Vec<String> = option_value.split(',').map(ToString::to_string).collect();
+            Ok(ParsedAuditEventOption::KeyNames(key_names))
+        }
+        "request-id" => Ok(ParsedAuditEventOption::RequestId(option_value.parse()?)),
+        "before" => {
+            anyhow::bail!("`before` option is not implemented");
+        }
+        "after" => {
+            anyhow::bail!("`after` option is not implemented");
+        }
```
vidyaap, line 160: I saw the date filters weren't checked off on the issue either - is the plan to do those in another PR or to not include them in the CLI?

lock-keeper/src/types/audit_event.rs
```diff
@@ -110,10 +111,164 @@ impl EventType {
...
+            match &mut self.key_ids {
+                Some(key_ids) => {
+                    key_ids.extend(incoming_key_ids);
+                }
+                None => self.key_ids = Some(incoming_key_ids),
+            }
+        }
+
+        self.after_date = other.after_date.or(self.after_date);
+        self.before_date = other.before_date.or(self.before_date);
```
vidyaap, line 138: I'm not sure I understand what's happening with the dates when you merge. The tests make it seem like the original dates get overwritten with the merged dates, but I'm not sure what that means semantically when you merge options? I think I would have expected to pick the "earlier after" and the "later before" if that makes sense.

lock-keeper/src/types/audit_event.rs
```diff
@@ -83,7 +83,8 @@ impl Display for AuditEvent {
 }
 
 /// Options for which types of events to retrieve from the key server
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, EnumString, Display)]
+#[strum(serialize_all = "kebab-case")]
 pub enum EventType {
```
vidyaap, line 88: Could you add GetUserId to system-only events? I'm not sure it should be a key event.

lock-keeper/src/types/audit_event.rs
picklenerd: Yeah it currently overwrites but I like your idea better!

picklenerd: I don't want to write a bunch of date stuff since the types are about to change so I think I'm just going to throw this merge function in a branch and remove it from this PR.

bin/lock-keeper-client-cli/src/cli_command/get_audit_events.rs
picklenerd: I'll make a new ticket for those.

bin/lock-keeper-client-cli/src/cli_command/get_audit_events.rs
picklenerd: I can get behind that.

lock-keeper/src/types/audit_event.rs
picklenerd: Added. I changed the way the filters work a bit so that they use a constant instead of allocating new vectors.

bin/lock-keeper-client-cli/src/cli_command/get_audit_events.rs
picklenerd: #434

vidyaap (approved):
Thanks, looks good to me!

PR #433, opened 2023-01-06
Use randomly generated session keys in tests #415

### LINKED ISSUE #415: Fix testing nits around opaque session key generation
#408 adds some tests of new properties of the opaque session key. These tests use a hardcoded session key, but our convention in cryptographic tests is to use randomized key material.

The benefit of randomized key material is to (slowly) get increased test coverage over the domain of possible keys. The downside of this approach is that if a randomized test fails, it's not easy to determine what the exact test case was. We are testing out an approach in tss-ecdsa to log RNG seeds in case of a failure, so tests can be both randomized and reproducible. If that solution is satisfactory, we'll probably port it over here. But for now, it seems better to match our randomization conventions.

There are [two unresolved comments](https://github.com/boltlabs-inc/key-mgmt/pull/408#pullrequestreview-1231664581) in #408 pointing out specific locations that can be switched to our existing convention.

vidyaap:
Closes #415 

Apologies, I think I missed the `create_test_session_key` function the first time!

### DIFF
--- a/lock-keeper/src/crypto.rs
+++ b/lock-keeper/src/crypto.rs
@@ -358,7 +358,9 @@ mod test {
 
     // In practice, a session key will be a pseudorandom output from OPAQUE.
     // We'll use random bytes for the test key.
-    fn create_test_session_key(rng: &mut (impl CryptoRng + RngCore)) -> OpaqueSessionKey {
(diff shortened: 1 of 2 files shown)


marsella (approved):
Looks perfect, thank you for following up!

PR #436, opened 2023-01-09
Remove user IDs from stream messages #289

### LINKED ISSUE #289: Remove `user_id` from requests
Once we have basic sessions working, we'll have the user ID for the authenticated user before we run handler code on the server.

We have a few operations that send the user_id as part of their request. The handlers for those requests are supposed to compare the passed user_id to the authenticated user's user_id. This step will be redundant once we have authenticated sessions since we already know the user is who they say they are if we can decrypt their messages with the session key. We can then use the user_id associated with the session in those handlers.

Once we're using the stored user_id, we can remove that parameter from the requests that currently have it. We will likely need some kind of dummy data to put in those requests so that we have something to encrypt with the session key. A simple string with an identifier for the message type would probably suffice. This might be worth adding to all of the message types.

- [ ] Add the `user_id` to the server's `Context` type if it's not already there.
- [ ] Remove `user_id` from `client::Generate` and `client::Store` in the `generate` operation.
- [ ] Remove `user_id` from `client::Request` in the `import_signing_key` operation.
- [ ] Remove `user_id` from `client::RequestRemoteGenerate` in the `remote_generate` operation.
- [ ] Remove `user_id` from `client::Request` in the `retrieve_storage_key` operation.
- [ ] Remove `user_id` from `client::Request` in the `retrieve` operation.
- [ ] Remove `user_id` from `client::RequestSigningKey` in the `retrieve` operation.
- [ ] For each of these operations, update the handler code to use the `user_id` stored in the `Context`.
- [ ] For each of these operations, if the resulting message has no fields, add some kind of dummy data.

vidyaap:
Closes #289

### DIFF
--- a/lock-keeper-client/src/api/generate_secret.rs
+++ b/lock-keeper-client/src/api/generate_secret.rs
@@ -28,7 +28,7 @@ impl LockKeeperClient {
         let storage_key = self.retrieve_storage_key(request_id).await?;
 
         // Generate step: get new KeyId from server
-        let key_id = get_key_id(&mut channel, self.user_id()).await?;
+        let key_id = channel.receive::<server::Generate>().await?.key_id;
         // Store step: encrypt secret and send to server to store
         let wrapped_secret = generate_and_store(
             &mut channel,
@@ -46,22 +46,6 @@ impl LockKeeperClient {
     }
 }
 
-async fn get_key_id(
-    channel: &mut ClientChannel<Authenticated<StdRng>>,
-    user_id: &UserId,
-) -> Result<KeyId, LockKeeperClientError> {
-    // Send UserId to server
-    let generate_message = client::Generate {
-        user_id: user_id.clone(),
-    };
-    channel.send(generate_message).await?;
-
-    // Get KeyId from server
-    let generate_result: server::Generate = channel.receive().await?;
-
-    Ok(generate_result.key_id)
-}
-
 async fn generate_and_store(
     channel: &mut ClientChannel<Authenticated<StdRng>>,
     user_id: &UserId,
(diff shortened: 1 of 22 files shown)


lock-keeper-key-server/src/operations/create_storage_key.rs
```diff
@@ -40,57 +43,26 @@ impl<DB: DataStore> Operation<Authenticated<StdRng>, DB> for CreateStorageKey {
...
-        request.account_name
-    );
+) -> Result<(), LockKeeperServerError> {
+    info!("Storing storage key.");
+    let client_message: client::SendStorageKey = channel.receive().await?;
 
     let user = context
         .db
-        .find_user(&request.account_name)
+        .find_user_by_id(&user_id)
```
vidyaap, line 57: I wasn't sure about making two DB calls in here to make this check, but I also didn't want to wait until the database function itself to check. Which would be better here?

lock-keeper-client/src/api/generate_secret.rs
```diff
@@ -28,40 +28,24 @@ impl LockKeeperClient {
         let storage_key = self.retrieve_storage_key(request_id).await?;
 
         // Generate step: get new KeyId from server
-        let key_id = get_key_id(&mut channel, self.user_id()).await?;
+        let key_id_result: server::Generate = channel.receive().await?;
```
picklenerd, line 31:
```suggestion
        let key_id = channel.receive::<server::Generate>().await?.key_id;
```

If you want to store the key_id directly here, you could do it like this.

picklenerd (approved):
LGTM. I'm not a huge fan of the extra error case where the metadata is missing the user_id but there's no way to avoid that right now. I'm working on addressing this in another branch.

PR #438, opened 2023-01-09
AccountId refactoring #413

### LINKED ISSUE #413: Ensure that all audit events are tied to an account
We create audit events for registration even if registration fails and an account is not created. This means that audit events may or may not be associated with an actual account and complicates the database mapping between audit events and accounts.

In the software developer working session, we discussed the merits of having audit events for failed registration. This seems like something that should be logged in the server for security purposes, but not recorded as an audit event. It seems unlikely that anyone would perform audits on non-existent accounts.

We are proposing to remove registration audit events that are created before the account is created, and replacing the `account_name` field in the database with `account_id` which maps directly to the (soon to be created) `accounts` table.

We would like to discuss this potential change with the rest of the team and @indomitableSwan in particular.

picklenerd:
WIP

This PR does some refactoring based on our new database structure. It replaces every usage of `AccountName` or `UserId` with `AccountId` which is the actual database ID for the user. Sessions, secrets, and audit events have all been updated in the database to reflect this.

When we fetch a session from the cache, we check the `Accounts` table to make sure that the `account_id` is mapped to a real user. This is because we don't assume that the session cache and main db are stored in the same database, so we can't make the `account_id` a foreign key. Since we already have the account information, the whole struct is now stored in the `Authenticated` payload in the channel. This lets us use the `account_id` for any database operations and the `user_id` for any crypto operations without having the query the database again.

I also moved the two channel types into their respective crates and gave them their own `AUTH` types. The server's version of `Authenticated` is different than the client's.

Finally, I added a `--stardard-only` or `-s` flag to run e2e tests only in the standard environment.


lock-keeper-key-server/src/operations/authenticate.rs
```diff
@@ -49,32 +73,45 @@ impl<DB: DataStore> Operation<Unauthenticated, DB> for Authenticate {
...
+        request_id,
+        ClientAction::Authenticate,
+        EventStatus::Started,
+    )
+    .await?;
 
     let server_login_start_result = {
         let mut local_rng = context.rng.lock().await;
 
         ServerLogin::start(
```
vidyaap, line 111: I think `ServerLogin::start` is the only call left that could fail but not get audited right now - is it worth making another manual audit here if it does, or are we relying on the logs in this case?

lock-keeper-key-server/src/operations/create_storage_key.rs
```diff
@@ -43,34 +40,30 @@ impl<DB: DataStore> Operation<Authenticated<StdRng>, DB> for CreateStorageKey {
...
-        .find_account_by_id(&user_id)
-        .await?
-        .ok_or(LockKeeperServerError::InvalidAccount)?;
+    let account_id = channel.account_id();
+    logging::record_field("account_id", &account_id);
 
-    if user.storage_key.is_some() {
+    if channel.account().storage_key.is_some() {
         return Err(LockKeeperServerError::StorageKeyAlreadySet);
     }
```
vidyaap, line 56: Does the storage key need to be set in this object too, as part of setting it in the DB?

lock-keeper-key-server/src/operations/authenticate.rs
picklenerd: Good question. Outside of server errors, the only way I think this can fail is if the account name doesn't match the opaque setup. That definitely seems audit-worthy to me. I'll add this.

lock-keeper-tests/src/config.rs
```diff
@@ -30,11 +30,23 @@ impl TryFrom<Cli> for Environments {
 
         let mut configs = HashMap::new();
         for (name, path) in environment_paths {
+            // The `standard_only` flag lets you run e2e tests locally against a single
```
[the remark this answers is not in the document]
gatoWololo: :pray:

lock-keeper-key-server/src/server/audit_events.rs
```diff
@@ -0,0 +1,29 @@
...
+};
+use tracing::instrument;
+use uuid::Uuid;
+
+use crate::LockKeeperServerError;
+
+use super::{database::DataStore, Context};
+
+#[instrument(skip(context))]
+pub(crate) async fn create_audit_event<DB: DataStore>(
```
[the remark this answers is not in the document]
picklenerd: I couldn't find a good place to put it. `server.rs` is already really busy and nowhere else seemed appropriate.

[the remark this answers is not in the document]
picklenerd: I guess this could just go in the context.

lock-keeper-key-server/src/operations/authenticate.rs
picklenerd: Ok this is actually a pain based on how to opaque type seems to operate. Also, now that I think about it the account_name not matching the opaque setup sounds like more of a server error than an audit event.

lock-keeper-key-server/src/server/audit_events.rs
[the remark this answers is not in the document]
picklenerd: Ok I made a `context` module instead and made this a method on the context. I've been wanting to do that anyway!

lock-keeper-key-server/src/operations/create_storage_key.rs
picklenerd: Yes probably! Added a method to the channel and updated this function.

vidyaap: Just realizing that technically we would always call authenticate after this which would pull the Account from the DB with the storage key, but this feels safer anyway 😆

lock-keeper-key-server/src/operations/authenticate.rs
vidyaap: I was thinking about what it means if someone puts in the wrong account name, and that being why they don't match. But I also did just try it out, and ended up having to check is_err on the `start` result, so I'll leave it up to you.

picklenerd: What do you mean by putting in the wrong account name? From what I understand, this is going to fetch the account based on whatever account name they put in, so the only way I can see this returning an error is if the opaque registration is corrupted in some way.

vidyaap: Oh never mind, I processed flow that wrong - this makes sense to me as a server error then!

vidyaap (approved):
Looks good!

PR #447, opened 2023-01-12
Update version to 0.3

picklenerd:
Plus:

- Removed the version numbers from our local dependencies since they don't do anything.
- Added a `cargo make` task to run all of the CLI scripts for easier testing.
- Removed unused client feature flag


vidyaap (approved):
LGTM!

PR #453, opened 2023-01-17
Make a separate database error type within the server #366

### LINKED ISSUE #366: Re-visit database errors
Database errors ended up being kind of tricky during the `DataStore` trait ticket #338. One of the downsides of the current approach is that some useful database errors get swallowed up into a generic "Internal Server Error". The. `InvalidAccount` error is one example that should be forwarded to the client.

We should revisit database errors, decide how we want them to work and how much granularity they should have, and implement the changes we come up with.

vidyaap:
Closes #366 

This is mostly just directly passing the existing error messages from the postgres crate to the server, but I had some questions:
- Do the "no entry" and "invalid count" errors need to be split into which table they came from (e.g. "no such key" vs. "no such user") or can we assume based on where the error originates?
- @gatoWololo I mapped the "EmptyIterator" error to "InvalidAuditEventOptions" since that's the only place I saw it used - do you anticipate it being used elsewhere? Or should the one in the postgres crate also say something about the audit event options?

I also changed the "associated key data" error to say key "metadata" instead since associated data is a thing on its own, but I'm not really sure about using the word "metadata" either.

### DIFF
--- a/lock-keeper-key-server/src/server.rs
+++ b/lock-keeper-key-server/src/server.rs
@@ -168,8 +168,7 @@ impl<DB: DataStore> LockKeeperRpc for LockKeeperKeyServer<DB> {
         let user_id = context
             .db
             .find_account(metadata.account_name())
-            .await
-            .map_err(LockKeeperServerError::database)?
+            .await?
             .ok_or(LockKeeperServerError::InvalidAccount)?
             .user_id;
 
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,3 @@
 target/
 .idea/
+**/logs/
--- a/lock-keeper-key-server/src/database.rs
+++ b/lock-keeper-key-server/src/database.rs
@@ -17,7 +17,29 @@ use lock_keeper::{
     },
 };
 use opaque_ke::ServerRegistration;
+use thiserror::Error;
+use tonic::Status;
 use uuid::Uuid;
+
+#[derive(Debug, Error)]
+pub enum DatabaseError {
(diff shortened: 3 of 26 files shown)


lock-keeper-key-server/src/server.rs
```diff
@@ -169,7 +169,7 @@ impl<DB: DataStore> LockKeeperRpc for LockKeeperKeyServer<DB> {
             .db
             .find_account(metadata.account_name())
             .await
-            .map_err(LockKeeperServerError::database)?
+            .map_err(LockKeeperServerError::Database)?
```
picklenerd, line 172: We could add this to the error definition:

```
impl From<DatabaseError> for Status {
    fn from(err: DatabaseError) -> Self {
        Status::internal(err.to_string())
    }
}
```

This would let us use `?` for all of these. 

Then in `impl From<LockKeeperServerError> for Status` we'd just do this instead of manually creating the status.

```
LockKeeperServerError::Database(err) => err.into(),
```

picklenerd:
> Do the "no entry" and "invalid count" errors need to be split into which table they came from (e.g. "no such key" vs. "no such user") or can we assume based on where the error originates?

I don't think so. These will get logged but the server probably doesn't need to care.

picklenerd:
> I also changed the "associated key data" error to say key "metadata" instead since associated data is a thing on its own, but I'm not really sure about using the word "metadata" either.

I think that's fine. That is what it is!

picklenerd (approved):

PR #454, opened 2023-01-18
Added docker image for fully set up database

picklenerd:
This PR creates a separate Docker image for a database that includes our migrations and pushes this image to the container registry.

This will allow us to use the key server and database in other projects without having to run migrations on the DB.


gatoWololo (approved):
Looks good.

PR #455, opened 2023-01-18
Add reconnection attempts to db

picklenerd:
This PR allows the database to try multiple times to connect before declaring failure. The number of retries and time between retries is configurable.


gatoWololo (approved):
Looks good.

PR #458, opened 2023-01-20
Add cli scripts to CI #448

### LINKED ISSUE #448: Add CLI scripts to CI
Now that we have a few CLI scripts, we can run them in CI as both a way to test the CLI and an extra check that our systems work.

The code that runs scripts currently may or may not exit with an error code if one of the operations in the script fails. This should be an option for the script runner. It could be useful to script a failed operation and a follow up operation that still needs to run. On the other hand, CI needs the CLI to exit with an error code if something fails.

We can add a CLI argument that enables or disables exiting with an error code on failure.

- [ ] Add `all-scripts` cargo make task to the dependency list in the `ci` task.
- [ ] Add CLI argument to exit with an error code if an operation fails within a script.

picklenerd:
Closes #448

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -76,7 +76,7 @@ dependencies = ["ci-test"]
 # -- CI Tasks --
 
 [tasks.ci]
-dependencies = ["ci-format", "ci-clippy", "ci-build", "ci-test", "ci-doc-test", "ci-docs"]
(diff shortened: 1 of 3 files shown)


picklenerd:
Closing for now because this doesn't work in CI and I don't want to spend time on it

PR #460, opened 2023-01-20
Clean up error handling in server CLI main.

### LINKED ISSUE #459: Unable to tell which start-up file is missing.
When starting Lock-Keeper, it is not possible to tell which config file was missing:
```bash
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })', bin/key-server-cli/src/main.rs:232:14
```
We should either tweak our error to return the offending file or add logging to we can tell which file was missing.

gatoWololo:
- Closes #459.
- Adds extra error variant for `LockKeeperServerError` for reporting file IO failures.
- Uses display for errors propegated up to main.
- Adds more explicit messages for remaining unwrap failures in main.rs.

### DIFF
--- a/lock-keeper-key-server/src/error.rs
+++ b/lock-keeper-key-server/src/error.rs
@@ -44,8 +44,14 @@ pub enum LockKeeperServerError {
     EnvVar(#[from] std::env::VarError),
     #[error(transparent)]
     Hyper(#[from] hyper::Error),
+    /// Generic kitchen sink IO error. Use [LockKeeperServerError::FileIo] if
+    /// the IO error is specifically related to working with files.
     #[error(transparent)]
     Io(#[from] std::io::Error),
+    /// IO error specific to file IO failing. Allows us to include the file that
+    /// failed as part of the error.
+    #[error("File IO error. Cause: {0}. On file: {1}")]
+    FileIo(std::io::Error, PathBuf),
     #[error(transparent)]
     LockKeeper(#[from] lock_keeper::LockKeeperError),
     #[error("OPAQUE protocol error: {}", .0)]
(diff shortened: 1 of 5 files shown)


lock-keeper-key-server/src/error.rs
```diff
@@ -44,8 +44,17 @@ pub enum LockKeeperServerError {
...
     #[error(transparent)]
     Hyper(#[from] hyper::Error),
+    /// Generic kitchen sink IO error. Use [LockKeeperServerError::FileIo] if
+    /// the IO error is specifically related to working with files.
+    // We would like to treat specific IO errors differently, e.g. failing to read a file should
+    // return the offending file as part of the error. So we avoid implementing a From instance on
+    // on this variant.
     #[error(transparent)]
-    Io(#[from] std::io::Error),
+    Io(std::io::Error),
```
picklenerd, line 53: Without the `#[from]` I don't think this variant is useful for anything since you can't `?` it.

picklenerd: Unless I'm missing something, I think we can leave `#[from]` and continue handling special cases manually like with `FileIo`.

picklenerd (approved):
One comment but looks great otherwise. I'm looking forward to having useful file errors!

lock-keeper-key-server/src/error.rs
gatoWololo: There is a few places where I manually use `.map_error(LockKeeperServerError::IO)?` but you do have to explicitly use it like this:
```rust
            file.write_all(bincode::serialize(&server_setup)?.as_slice())
                .map_err(LockKeeperServerError::Io)?;
```
The idea was that the developer has to manually choose which variant, so they don't accidentally `?` a file error without the file. I could see us keeping the `#[from]` instance so there is a default conversion for `?`.

picklenerd: I think that would be the most usable. Otherwise you'd have to choose to not do anything special with the error any time you use it which seems cumbersome.

gatoWololo: Makes sense. Fixed!

PR #461, opened 2023-01-20
Implement AsRef<[u8]> for KeyId

picklenerd:
It's time for this madness to stop. We shouldn't have to allocate a vector and turn this into an iterator to get its bytes.

current:
```
    let key_id: HexBytes = maestro
        .client
        .remote_generate()
        .await
        .result?
        .key_id
        .into_iter()
        .collect::<Vec<u8>>()
        .into();
```

desired:
```
    let key_id = maestro
        .client
        .remote_generate()
        .await
        .result?
        .key_id
        .as_ref()
        .into();
```

There's no misuse resistance in the current implementation where `into_iter` is the only option.

### DIFF
--- a/lock-keeper/src/crypto.rs
+++ b/lock-keeper/src/crypto.rs
@@ -286,6 +286,12 @@ impl TryFrom<&[u8]> for KeyId {
     }
 }
 
+impl AsRef<[u8]> for KeyId {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_ref()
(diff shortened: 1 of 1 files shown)


gatoWololo (approved):
> It's time for this madness to stop. We shouldn't have to allocate a vector and turn this into an iterator to get its bytes.

I agree :joy:

marsella (approved):
Yes, this looks reasonable!

PR #464, opened 2023-01-25
Make TLS optional in the client

picklenerd:
Jay mentioned that disabling TLS could be useful for troubleshooting in the future. Our current client code doesn't seem to work for `http` and it's not as simple as it sounds to make it work.


picklenerd:
Closing as not useful

PR #469, opened 2023-01-30
Moved docker logs to logs dir

picklenerd:
This PR moves docker logs into the `/dev/logs` directory instead of having them live side-by-side.

It also implements the clippy fixes for 1.67.


gatoWololo:
I think we should pin the Rust version as part of this PR:
```diff
      - uses: actions-rs/toolchain@v1
        with:
-          toolchain: stable
+          toolchain: 1.67
          components: clippy
          default: true
```
We can manually update it as needed with clippy fixes to avoid random CI breakage.

picklenerd:
> I think we should pin the Rust version as part of this PR:

Done!

gatoWololo (approved):

Issue #292, opened 2022-10-17
Improve code quality for cryptography -> bytes conversions


marsella:
There are a bunch of methods in the `crypto` module of `lock-keeper` that involve converting between some type and bytes (`Vec<u8>`). These exist to provide consistent formatting for types when they are encrypted or signed. 

Typically, these conversions involve taking each field in a struct, writing its length in bytes (1 byte) and then serializing the field itself. Sometimes the length is fixed or known in advance, but often it isn't. Serialization varies by field -- sometimes there's a `to_bytes` method, but it's often inconsistent what ti takes and what error type it returns.

While working on #235, we identified two concrete ways to improve the code quality on these conversions.
1. Write a little helper function to read and write the length in bytes + content for various types. Given that the serialization method varies so much, we might have to play around with what parameter this takes or maybe write a macro (??). This should be instantiated for any type that is a field in a cryptography type with a `TryFrom<Vec<u8>>` implementation.

2. Try using lambda functions to simplify error handling. The types being serialized return a wide variety of errors, but we always want the conversion function to return a `CryptoError::ConversionError`. In the current implementation, every line that throws an error has to map it to a `ConversionError`, which is incredibly repetitious. One suggestion made was to put a little anonymous function inside each `try_from` method that will return a generic error type, so we only have to do one conversion to `ConversionError`.

At time of writing, these changes could be applied to
- `StorageKey`
- `generic::Secret`
- `AssociatedData` (this currently doesn't prepend the length in its implementation, but it should)
- `EncryptionKey`
- `Encrypted<T>`
- `arbitrary_secret::Secret`
- `SigningKeyPair`

--- 
Original notes from #262:
> The interface for converting to / from Vec<u8> for SigningKeyPairs (as well as other types not addressed in this PR) is quite ugly. Per discussion with @picklenerd, @gatoWololo, and @vidyaap, we should try to write a generic little parser to read a length, then read that many bytes and return (len, bytes, remaining_bytes), and apply this to SigningKeyPair, StorageKey, AssociatedData, EncryptionKey, and Secret. On second thought, maybe this should be its own issue.
>
> That same interface has a ton of error handling that all ends up returning the same error. Try wrapping this in a lambda, throwing up the errors with ?, then converting to CryptoError::ConversionError exactly once.

Issue #237, opened 2022-09-14
Make associated data more parseable


marsella:
Right now, the associated data used in our AEAD schemes and as context for secrets and encryption keys is just a string. We can rebuild the exact string if we want to check that AD is correctly formed, but it's pretty fragile--building it with the right arguments in the wrong order will fail, even when they are semantically the same.

We should replace this with something more easily parseable, like a key-value store. It should be
- any associated data is deterministically (?) serializable
- the set of keys is known at compile-time
- any unknown keys will cause the AD to fail to parse
- each key appears either 0 or 1 time in a given AD
- some keys might have a fixed set of possible values, others will have unbounded values

We can also add a method to check whether two ADs match on a given key and overall.

gatoWololo:
Would a serializable struct be enough to give us what we want? For example:
```rust
struct AssociatedData {
   key1: Option<Type1>,
   key2: Option<Type2>,
   ...
}
```

> any associated data is deterministically (?) serializable

We can serialize this struct into json, RON, or some other format for storing.

> the set of keys is known at compile-time

The keys are just the struct fields which are known at compile-time. Adding additional fields/keys is not an issue if wrap them in an `Option` or set a default value. The deserializer can look at the bytes and simply return `None` if a field is missing.

> any unknown keys will cause the AD to fail to parse

Already true if we deserialize into our struct.

> each key appears either 0 or 1 time in a given AD

This will also be enforced by our serialization/deserialization.


> some keys might have a fixed set of possible values, others will have unbounded values

The types of our fields can be strings, but also integers, enums, etc. Supporting this requirement while also giving us type safety.

> We can also add a method to check whether two ADs match on a given key and overall.

I envision us serializing for storage, but otherwise we would work with the Rust types. So we can compare ADs via a `Eq` implementation or by comparing field values manually.

marsella:
Oh yeah, this sounds perfect! I had been imagining a hash map, but this is much better suited to the requirements.

PR #470, opened 2023-01-31
Storage For Arbitrary Secrets #432

### LINKED ISSUE #432: Key Server: Support Storage of Arbitrary Client Data
**Background:** For other related projects, we may use the Key Server to store encrypted shards or to store arbitrary backup data (blobs). The blobs may already be encrypted by the client calling application prior to being sent to the Key Server. While we could utilize the arbitrary secrets functionality, it is not a perfect fit for the application. We might also be able to utilize the import functionality, but again, it is not a perfect fit. 

**Developer Note:** The developer assigned to this ticket should work closely with the Bolt cryptographers to select an appropriate course of action -- i.e., reuse an existing feature of the Key Server or implement a new feature. The requirements in this ticket are meant as a starting point, but should not impede the developer and cryptographers from selecting a better path forward. Please capture your notes after those discussions.

**Requirements:**

1. Provide a Key Server action that allows the client to store an arbitrary blob of data -- the arbitrary blob of data may have already been encrypted by the client (or may not)
1. Provide a Key Server action to retrieve an arbitrary blob of data
1. Write a log statement when either action (store or retrieve) is invoked that includes the account_id and blob_id (the blob_id might be the key_id -- TBD by the designer)
1. Capture audit events when a blob is stored or retrieved
1. The Key Server should encrypt the data blob before placing into the persistent storage of the Key Server
1. The Key Server should have a configurable maximum size of the blob to accept -- that is, if a blob exceeds the pre-configured maximum size, then an appropriate error should be returned and the request should be rejected

**Tasks:**

1. Work in conjunction with the Bolt cryptographers to determine whether we will use pre-existing functionality of the Key Server to satisfy these requirements or if we will add a new feature to the Key Server
1. Document your decisions and reasoning
1. If changes are required of the Key Server and/or LK Client, then document the enhancements/design
1. if changes are required, then this ticket should also track the implementation

gatoWololo:
I will make a write up and comment on some things tomorrow! But it is ready for review.

Closes #432

### DIFF
--- a/lock-keeper-client/src/client.rs
+++ b/lock-keeper-client/src/client.rs
@@ -241,8 +241,10 @@ impl LockKeeperClient {
             | ClientAction::RemoteSignBytes
             | ClientAction::RetrieveSecret
             | ClientAction::RetrieveAuditEvents
+            | ClientAction::RetrieveServerEncryptedBlob
             | ClientAction::RetrieveSigningKey
-            | ClientAction::RetrieveStorageKey => {
+            | ClientAction::RetrieveStorageKey
+            | ClientAction::StoreServerEncryptedBlob => {
                 return Err(LockKeeperClientError::AuthenticatedChannelNeeded)
             }
         }?;
@@ -288,10 +290,16 @@ impl LockKeeperClient {
             ClientAction::Register => client.register(stream).await,
             ClientAction::RemoteGenerateSigningKey => client.remote_generate(stream).await,
             ClientAction::RemoteSignBytes => client.remote_sign_bytes(stream).await,
+            ClientAction::RetrieveServerEncryptedBlob => {
+                client.retrieve_server_encrypted_blob(stream).await
+            }
             ClientAction::RetrieveSecret => client.retrieve_secret(stream).await,
             ClientAction::RetrieveAuditEvents => client.retrieve_audit_events(stream).await,
             ClientAction::RetrieveSigningKey => client.retrieve_secret(stream).await,
             ClientAction::RetrieveStorageKey => client.retrieve_storage_key(stream).await,
+            ClientAction::StoreServerEncryptedBlob => {
+                client.store_server_encrypted_blob(stream).await
+            }
 
             // These actions generate an error because they should be on an unauthenticated channel
             ClientAction::Authenticate | ClientAction::Register => {
--- a/lock-keeper-key-server/src/server.rs
+++ b/lock-keeper-key-server/src/server.rs
@@ -72,9 +72,11 @@ impl<DB: DataStore> LockKeeperRpc for LockKeeperKeyServer<DB> {
     type GetUserIdStream = MessageStream;
     type ImportSigningKeyStream = MessageStream;
     type LogoutStream = MessageStream;
+    type StoreServerEncryptedBlobStream = MessageStream;
     type RegisterStream = MessageStream;
     type RemoteGenerateStream = MessageStream;
     type RemoteSignBytesStream = MessageStream;
+    type RetrieveServerEncryptedBlobStream = MessageStream;
     type RetrieveSecretStream = MessageStream;
     type RetrieveAuditEventsStream = MessageStream;
     type RetrieveStorageKeyStream = MessageStream;
@@ -146,6 +148,20 @@ impl<DB: DataStore> LockKeeperRpc for LockKeeperKeyServer<DB> {
         Ok(response)
     }
 
+    async fn store_server_encrypted_blob(
+        &self,
+        request: Request<Streaming<Message>>,
+    ) -> Result<Response<Self::StoreServerEncryptedBlobStream>, Status> {
+        let (channel, response) = self.create_authenticated_channel(request).await?;
+        handle_authenticated_request(
+            operations::StoreServerEncryptedBlob,
+            self.context(),
+            channel,
+        )
+        .await?;
+        Ok(response)
+    }
+
     async fn remote_generate(
         &self,
         request: Request<tonic::Streaming<Message>>,
@@ -166,6 +182,20 @@ impl<DB: DataStore> LockKeeperRpc for LockKeeperKeyServer<DB> {
     ) -> Result<Response<Self::RemoteSignBytesStream>, Status> {
         let (channel, response) = self.create_authenticated_channel(request).await?;
         handle_authenticated_request(operations::RemoteSignBytes, self.context(), channel).await?;
+        Ok(response)
+    }
+
+    async fn retrieve_server_encrypted_blob(
+        &self,
+        request: Request<Streaming<Message>>,
+    ) -> Result<Response<Self::RetrieveServerEncryptedBlobStream>, Status> {
+        let (channel, response) = self.create_authenticated_channel(request).await?;
+        handle_authenticated_request(
+            operations::RetrieveServerEncryptedBlob,
+            self.context(),
+            channel,
+        )
+        .await?;
         Ok(response)
     }
 
--- a/lock-keeper-tests/src/test_suites/end_to_end/operations.rs
+++ b/lock-keeper-tests/src/test_suites/end_to_end/operations.rs
@@ -83,7 +83,7 @@ pub(crate) async fn check_audit_events(
         .result?;
 
     // Get all events that match given expected values.
-    let matching_events = audit_event_log.into_iter().filter(|event| {
+    let matching_events = audit_event_log.iter().filter(|event| {
         event.key_id == key_id
             && event.status == expected_status
             && event.client_action == expected_action
@@ -92,7 +92,7 @@ pub(crate) async fn check_audit_events(
     assert_eq!(
         matching_events.count(),
         1,
-        "Exactly one audit event should have matched."
+        "Exactly one audit event should have matched. All events: {audit_event_log:?}",
     );
 
     Ok(())
--- a/lock-keeper-tests/src/test_suites/end_to_end/test_cases/retrieve.rs
+++ b/lock-keeper-tests/src/test_suites/end_to_end/test_cases/retrieve.rs
@@ -4,6 +4,7 @@ use lock_keeper::types::{
     operations::{retrieve_secret::RetrieveContext, ClientAction},
 };
 use lock_keeper_client::{api::GenerateResult, Config};
+use rand::Rng;
 use tonic::Status;
 
 use crate::{
@@ -29,11 +30,16 @@ pub async fn run_tests(config: &Config, filters: &TestFilters) -> Result<Vec<Tes
         retrieve_null_works(config.clone()),
         cannot_retrieve_fake_key(config.clone()),
         cannot_retrieve_after_logout(config.clone()),
+        store_retrieve_server_encrypted_blob(config.clone()),
+        blob_size_too_large(config.clone()),
     )?;
 
     Ok(result)
 }
 
+/// Create a client-generated secret via `generate_secret`.
+/// Fetch the secret back out and ensure the material matches what we stored.
+/// Check the audit events for a successful retrieval.
 async fn retrieve_local_only_works(config: Config) -> Result<()> {
     let state = init_test_state(&config).await?;
     let client = authenticate(&state).await.result?;
@@ -123,6 +129,74 @@ async fn cannot_retrieve_after_logout(config: Config) -> Result<()> {
 
     let res = client.retrieve_secret(&key_id, RetrieveContext::Null).await;
     compare_status_errors(res, Status::unauthenticated(NO_SESSION))?;
+
+    Ok(())
+}
+
+/// Tests our server encrypted blob functionality by storing and retrieving a
+/// data blob.
+async fn store_retrieve_server_encrypted_blob(config: Config) -> Result<()> {
+    let state = init_test_state(&config).await?;
+    let client = authenticate(&state).await.result?;
+    let data: Vec<u8> = std::iter::repeat_with(|| rand::thread_rng().gen())
+        .take(512)
+        .collect();
+
+    // Store blob and check audit events.
+    let server_response = client.store_server_encrypted_blob(data.clone()).await;
+    let key_id = server_response.result?;
+    let request_id = server_response.metadata.clone().unwrap().request_id;
+
+    check_audit_events(
+        &state,
+        EventStatus::Successful,
+        ClientAction::StoreServerEncryptedBlob,
+        request_id,
+        Some(key_id.clone()),
+    )
+    .await?;
+
+    // Retrieved store blob, check data matches what we stored and audit events.
+    let lk_response = client.retrieve_server_encrypted_blob(&key_id).await;
+    let retrieved_blob = lk_response.result?;
+    let request_id = lk_response.metadata.clone().unwrap().request_id;
+
+    assert_eq!(data, retrieved_blob);
+    check_audit_events(
+        &state,
+        EventStatus::Successful,
+        ClientAction::RetrieveServerEncryptedBlob,
+        request_id,
+        Some(key_id),
+    )
+    .await?;
+
+    Ok(())
+}
+
+/// Trying to store a too-large data blob will result in server error.
+async fn blob_size_too_large(config: Config) -> Result<()> {
+    let state = init_test_state(&config).await?;
+    let client = authenticate(&state).await.result?;
+    // Our testing server is configured to accept a max blob size of 1024.
+    let data_blob = vec![24; 1025];
+
+    // Store blob and check audit events.
+    let server_response = client.store_server_encrypted_blob(data_blob.clone()).await;
+    assert!(
+        server_response.result.is_err(),
+        "Server should return error due to too large blob size "
+    );
+
+    let request_id = server_response.metadata.clone().unwrap().request_id;
+    check_audit_events(
+        &state,
+        EventStatus::Failed,
+        ClientAction::StoreServerEncryptedBlob,
+        request_id,
+        None,
+    )
+    .await?;
 
     Ok(())
 }
--- a/lock-keeper/src/crypto.rs
+++ b/lock-keeper/src/crypto.rs
@@ -22,12 +22,14 @@ use zeroize::{Zeroize, ZeroizeOnDrop};
 use crate::types::database::account::UserId;
 
 mod arbitrary_secret;
+mod data_blob;
 mod generic;
 mod signing_key;
 mod storage_key;
 
 use crate::rpc::Message;
 pub use arbitrary_secret::Secret;
+pub use data_blob::DataBlob;
 use generic::{AssociatedData, EncryptionKey};
 pub use generic::{CryptoError, Encrypted};
 pub use signing_key::{
@@ -310,8 +312,10 @@ impl KeyId {
         let domain_separator = b"Lock Keeper key ID";
 
         let hasher = Sha3_256::new();
-        let user_id_length = u8::try_from(user_id.len()).map_err(CryptoError::TryFromIntError)?;
-        let random_len = u8::try_from(RANDOM_LEN).map_err(CryptoError::TryFromIntError)?;
+        let user_id_length =
+            u8::try_from(user_id.len()).map_err(|_| CryptoError::CannotEncodeDataLength)?;
+        let random_len =
+            u8::try_from(RANDOM_LEN).map_err(|_| CryptoError::CannotEncodeDataLength)?;
 
         let bytes = hasher
             .chain_update(domain_separator.len().to_be_bytes())
--- a/lock-keeper/src/crypto/data_blob.rs
+++ b/lock-keeper/src/crypto/data_blob.rs
@@ -1,0 +1,152 @@
+//! This module defines `[DataBlob]`: A blob of bytes for storing in the server.
+//! As well as accompanying functionality for server-side encrypting and
+//! decrypting `[DataBlob]s`.
+
+use crate::{
+    crypto::{generic, generic::AssociatedData, CryptoError, Encrypted, KeyId, RemoteStorageKey},
+    types::database::account::UserId,
+    LockKeeperError,
+};
+use serde::{Deserialize, Serialize};
+use tracing::{error, instrument};
+use zeroize::ZeroizeOnDrop;
+
+/// A blob of bytes for storing in the server.
+///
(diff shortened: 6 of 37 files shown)


lock-keeper-key-server/src/operations/retrieve_blob.rs
```diff
@@ -0,0 +1,64 @@
...
+            .db
+            .get_secret(
+                account_id,
+                &request.key_id,
+                SecretFilter::secret_type(DATA_BLOB),
+            )
+            .await?;
+
+        let blob: Encrypted<DataBlob> =
+            serde_json::from_slice(&stored_secret.bytes).map_err(LockKeeperError::SerdeJson)?;
```
gatoWololo, line 51: @picklenerd Is it fine to call `serde_json` directly like this? I haven't seen this done directly like this in any other `Operation`

lock-keeper/src/crypto/data_blob.rs
```diff
@@ -0,0 +1,73 @@
+//! TODO Document
```
gatoWololo, line 1: @marsella I added my own module to the crypto code. `DataBlob` is implemented in terms of a `generic::Secret`. I based this code off the `arbitrary_secret` module.

(DataBlob has to live in the crypto module to have access to `AssociatedData` and other encryption/decryption methods). Originally I tried making `DataBlob` not based on `generic::Secret` but then I would have to do something similar to `impl TryFrom<Vec<u8>> for SigningKeyPair` where the bytes are hand-parsed. Which seemed like the wrong approach.

gatoWololo: I wanted to make sure this was the right approach? I will ping you for an actual crypto review once the PR comes out of draft.

lock-keeper/src/crypto/data_blob.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    // TODO: Document
+    pub fn create(data: Vec<u8>, user_id: &UserId, key_id: &KeyId) -> Result<Self, CryptoError> {
+        let context = AssociatedData::new()
+            .with_bytes(user_id.clone())
+            .with_bytes(key_id.clone())
+            .with_str(Self::GENERATION_TYPE);
+
+        // TODO: Are we okay with this redundant copy-ing.
+        Ok(DataBlob(Secret::from_parts(&data, &context)))
```
gatoWololo, line 31: @picklenerd @marsella To create the underlying secret, I call `Secret::from_parts` which takes references, so the data blob is copied unnecessarily. Is this fine? Or I can add an analogous method which takes the data by value.

lock-keeper/src/crypto/data_blob.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /// TODO
+    pub fn blob_data(self) -> Vec<u8> {
+        self.0.get_material().to_vec()
+    }
```
gatoWololo, line 37: @picklenerd @marsella This is another place where the underlying `generic::Secret` API doesn't quite fit our needs.
When calling retrieve blob:
1) The server fetches the data blob which is stored as a `StoredSecret` from the DB.
2) I deserialize the data bytes into an `Encrypted<DataBlob>` and decrypt that.
3) We now have a `DataBlob` and I want to return the underlying bytes to the client.
The only way to get the material is via:
```rust
pub(super) fn get_material(&self) -> &[u8] {
        &self.material
    }
```
So I have to call `to_vec()` when I just want to move out the bytes. Rust doesn't let me move out the bytes since `generic::Secret` implements `Drop`.

lock-keeper-key-server/src/operations/retrieve_blob.rs
picklenerd: I don't think this needs a server implementation at all. The client can call `RetrieveSecret` and then deal with deserialization itself.

picklenerd: Yeah this is fine. The server doesn't normally care about the secret type so it never deserializes anything.

lock-keeper-client/src/client.rs
```diff
@@ -241,8 +241,10 @@ impl LockKeeperClient {
             | ClientAction::RemoteSignBytes
             | ClientAction::RetrieveSecret
             | ClientAction::RetrieveAuditEvents
+            | ClientAction::RetrieveDataBlob
             | ClientAction::RetrieveSigningKey
-            | ClientAction::RetrieveStorageKey => {
+            | ClientAction::RetrieveStorageKey
+            | ClientAction::StoreDataBlob => {
```
picklenerd, line 247: After looking at the code, I think these should be `RetrieveEncryptedDataBlob` and `StoreEncryptedDataBlob`. The thing that makes these different than normal user secrets is that they're encrypted by the server, so I think we should highlight this with the names.

If you agree, I think everything in this PR should be renamed similarly.

lock-keeper/src/crypto/data_blob.rs
picklenerd: I'd prefer making `from_parts` take the data by value. There's rarely a good reason to take parameters by reference and then immediately clone them IMO.

Tried this and it was easy to propagate:

```
    pub(super) fn from_parts(secret_material: impl Into<Vec<u8>>, context: AssociatedData) -> Self
```

lock-keeper/src/crypto/data_blob.rs
picklenerd: I'd say add an `into_parts` method mirroring `from_parts` to get the bytes out.

lock-keeper-client/src/client.rs
gatoWololo: Sounds good!

lock-keeper/src/crypto/data_blob.rs
marsella: Yeah, I'm fine if you change that. I had a floating code quality issue I never handled to go fix all the things that take references but should take values in the crypto module.

lock-keeper/src/crypto/data_blob.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/// This implementation is required to use the `[Encrypted::encrypt]` function.
+impl TryFrom<Vec<u8>> for DataBlob {
+    type Error = CryptoError;
+    fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
+        Ok(DataBlob(value.try_into()?))
+    }
+}
```
marsella, line 57: suggestion: It looks like you mirrored the implementation from `arbitrary_secret`, but I think it would be good to add a domain separator here like in the key pairs (see e.g. [signing key pair](https://github.com/boltlabs-inc/key-mgmt/blob/develop/lock-keeper/src/crypto/signing_key.rs#L334-L360)).

Otherwise arbitrary secret vecs could decrypt into data blobs and vice versa, which opens the door to problems.

lock-keeper/src/crypto/data_blob.rs
marsella: Yep, this is a great approach. One module per type of data stored at the server.

The hand-parsing `TryFrom`s are awful, I agree, but I sadly left a comment below that we may want to add an extra field to the bytes conversion for encryption to make them less ambiguous.

lock-keeper/src/crypto/data_blob.rs
picklenerd: You can easily call `get_material` to get a reference to the bytes and then convert them into a `Vec` right now.

So in the current setup, there are two copies of the secret and only one of them gets cleaned up. In the proposed setup, there is one copy of the secret and it doesn't get cleaned up.

gatoWololo: > But maybe I'm misunderstanding. Can we already arbitrarily pull the bytes out of a Secret without dropping them?

Currently (before this PR). You can already borrow a reference to the `material` and call `clone/into_vec` on that copy. The original bytes will eventually be dropped and zeroized, but you can do whatever you want with your copy. However, you can only do this from within the `Crypto` module due to the `pub(super)` field. 

> Does it just add the additional cost of allocating a vec?

Yeah, I'm just trying to avoid the cost of allocation and copying.

> I was going to say, if DataBlob is a less-secret thing that gets returned to the client, and types based on Secret are more-secret things that always live on the server, then maybe DataBlob should use a different underlying type.

A good middle ground that would allow us to reuse the existing `Secret` code without exposing methods we don't want: Rename the current `Secret` implementation to something like `SecretMaterial` which exposes more methods. Then `Secret` would just be a wrapper around `SecretMaterial` which restricts the API. Then `DataBlob` could also just be a wrapper around `SecretMaterial`.

Anyways. Thanks for the info. I can do something sensible from here and you can review it when the PR is ready!

gatoWololo: > I'd say add an into_parts method mirroring from_parts to get the bytes out.

@picklenerd If it changes your opinion at all. We can't simply move out the bytes due to the type having a destructor:
```
error[E0509]: cannot move out of type `generic::Secret`, which implements the `Drop` trait
   --> lock-keeper/src/crypto/generic.rs:297:9
    |
297 |         self.material
    |         ^^^^^^^^^^^^^
    |         |
    |         cannot move out of here
    |         move occurs because `self.material` has type `Vec<u8>`, which does not implement the `Copy` trait
```
The only way I know to do this is slightly hacky :mask: :
```rust
pub(super) fn move_material(mut self) -> Vec<u8> {
        let mut material = Vec::new();
        std::mem::swap(&mut self.material, &mut material);
        material
    }
```

lock-keeper-client/src/client.rs
indomitableSwan: > After looking at the code, I think these should be `RetrieveEncryptedDataBlob` and `StoreEncryptedDataBlob`. The thing that makes these different than normal user secrets is that they're encrypted by the server, so I think we should highlight this with the names.
> 

@picklenerd Do you mean "encrypted by the client" in the above? 

> If you agree, I think everything in this PR should be renamed similarly.

lock-keeper/src/crypto/data_blob.rs
marsella: Ah, okay. `get_material` is not well documented with its intended use right now. If you look at the other place it's used, it transforms a `Secret` into an `Export`. The way I'd like to see it used in general is to transform from one `ZeroizeOnDrop` type to another. In the blob retrieval method, I'd rather see it return an equivalent type to `Export` (`ExportedBlob`? or as Omar suggested, `BlobData`?) that has explicit guidance for data handling and implements `ZeroizeOnDrop` (or is explicit about not being secret data), instead of returning a plain `Vec<u8>` or `&[u8]` that could be used for anything.

My recommended approach:
- Replace `get_material` with `into_parts`, with explicit documentation that the raw output of `into_parts` should not be e.g. returned directly to a user
- Refactor my terrible memory handling in `From<Secret> for Export`
- Create a new `BlobData` type with documentation about the secrecy level of the data it holds
- Update this function to return a `BlobData`

Does that sound reasonable?

> In both cases, the secret being consumed doesn't get cleaned up.

By this, do you mean the `Vec` that comes out of `get_material`/`into_parts`? Should we manually zeroize it or something?

marsella: Oh I didn't see Omar's last two comments when I wrote the above, but I am open to the approach you described.

lock-keeper-client/src/client.rs
picklenerd: The blobs are also encrypted by the server.

gatoWololo: I think Swan means that the name is slightly ambiguous on who is doing the encryption/decryption. For example, a client function named `store_encrypted_blob` makes it unclear whether the function expects the blob to be encrypted already.

gatoWololo: Jordan suggested the name `ServerEncryptedBlob` which I like. So we will use something along those lines.

indomitableSwan: I'm also unclear how useful this naming is, when the server should _always_ be encrypting data that needs to be stored, no matter what type of data that is...

lock-keeper/src/crypto/data_blob.rs
gatoWololo: This has been moved to a separate issue #476. I now realize there a few other issues related to crypto floating around.

lock-keeper-key-server/src/server.rs
```diff
@@ -169,6 +180,15 @@ impl<DB: DataStore> LockKeeperRpc for LockKeeperKeyServer<DB> {
...
 
+    async fn retrieve_blob(
+        &self,
+        request: Request<Streaming<Message>>,
+    ) -> Result<Response<Self::RetrieveBlobStream>, Status> {
+        let (channel, response) = self.create_authenticated_channel(request).await?;
+        handle_authenticated_request(operations::RetrieveBlob, self.context(), channel).await?;
+        Ok(response)
+    }
+
```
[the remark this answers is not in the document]
gatoWololo: Good question! In general we try to avoid macros unless super necessary. There is already some autogeneration magic going on with this trait as it is created from our RPC file. While it is fairly boilerplate-y it is also straightforward to understand.

I know it took me a while to get my head wrapped around how this trait worked. So wrapping it in a macro would reduce the overall code (by hiding the different function implementation behind the macro) but it would also make it harder to understand what is happening.

The copy and paste code here is fairly innocuous, and we haven`t had to change it much. So we just leave it.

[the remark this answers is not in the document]
picklenerd: Agreed. If it were a big block of code or we had to modify it a lot, I think a macro would be ok. Feels like overkill for this.

lock-keeper-tests/src/test_suites/end_to_end/operations.rs
```diff
@@ -83,7 +83,7 @@ pub(crate) async fn check_audit_events(
         .result?;
 
     // Get all events that match given expected values.
-    let matching_events = audit_event_log.into_iter().filter(|event| {
+    let matching_events = audit_event_log.iter().filter(|event| {
```
gatoWololo, line 86: When this check failed, it was annoying to debug as no additional information was provided. Now we print out the unexpected result for easier debugging.

lock-keeper/src/crypto.rs
```diff
@@ -310,8 +312,9 @@ impl KeyId {
         let domain_separator = b"Lock Keeper key ID";
 
         let hasher = Sha3_256::new();
-        let user_id_length = u8::try_from(user_id.len()).map_err(CryptoError::TryFromIntError)?;
-        let random_len = u8::try_from(RANDOM_LEN).map_err(CryptoError::TryFromIntError)?;
+        let user_id_length =
+            u8::try_from(user_id.len()).map_err(CryptoError::CannotEncodeDataLength)?;
+        let random_len = u8::try_from(RANDOM_LEN).map_err(CryptoError::CannotEncodeDataLength)?;
```
gatoWololo, line 317: Updated error message to report it was related to length!

lock-keeper/src/crypto/data_blob.rs
```diff
@@ -0,0 +1,154 @@
...
+        Ok(bytes)
+    }
+}
+
+/// This implementation is required to use the `[Encrypted::encrypt]` function.
+impl TryFrom<Vec<u8>> for DataBlob {
+    type Error = CryptoError;
+    fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
+        // Check domain separator and strip it.
+        let ds: Vec<u8> = DataBlob::domain_separator().into();
```
gatoWololo, line 74: @marsella Per your request, I added a domain separator to the bytes. Lmk if this was the correct way to do it.

lock-keeper/src/crypto/generic.rs
```diff
@@ -364,6 +364,43 @@ impl TryFrom<Vec<u8>> for Secret {
     }
 }
 
+/// Helper type for parsing byte array into integers and slices.
+struct ParseBytes {
+    bytes: Vec<u8>,
+    offset: usize,
+}
```
gatoWololo, line 371: I made this small abstraction to help with parsing bytes out. You can see its usage above. If we like it we can change the other crypto types to also use it. (I didn't want to make this PR any bigger and wanted feedback on it first).

persistence/lock-keeper-sql/src/api.rs
```diff
@@ -249,24 +249,30 @@ impl PostgresDB {
...
         }
 
-        // Add filtering based on actions.
-        let _ = query.push("client_action_id IN ");
-        // Turn the actions into their integer value for faster searching.
-        let actions = event_type.client_actions();
-        let actions = actions.iter().map(|a| *a as i64);
-
-        append_value_list(&mut query, actions)?;
+        // Add filtering based on actions if the user specific event types.
```
gatoWololo, line 252: We only try filtering on client action if the client action is something other than "All".

lock-keeper-tests/src/test_suites/end_to_end/test_cases/retrieve.rs
```diff
@@ -126,3 +132,71 @@ async fn cannot_retrieve_after_logout(config: Config) -> Result<()> {
...
+
+    Ok(())
+}
+
+/// Trying to store a too-large data blob will result in server error.
+async fn blob_size_too_large(config: Config) -> Result<()> {
+    let state = init_test_state(&config).await?;
+    let client = authenticate(&state).await.result?;
+    // This blob is too big for server.
+    let data_blob = vec![24; 2048];
```
picklenerd, line 182: It might be useful to add a note here that the expected max blob size is 1024 and to check the config value if this test fails.

lock-keeper/src/crypto/generic.rs
```diff
@@ -25,7 +26,7 @@ pub enum CryptoError {
     RandomNumberGeneratorFailed,
     /// Length of data too large for our integer data type.
     #[error("Encryption/Decryption failed due to data length.")]
-    TryFromIntError(#[from] TryFromIntError),
+    CannotEncodeDataLength(#[from] TryFromIntError),
```
picklenerd, line 29: I'm a little worried that this might swallow up all `TryFromIntError` in the future where they might not actually be data length errors.

lock-keeper/src/crypto/generic.rs
picklenerd: My understanding is that the format for this is generally `[length_1, data_with_length_1, length_2, data_with_length_2]`.

I was imagining a more specific implementation for this where you could just ask for the next chunk of data and the function would first read the length and then read that much data, returning only the data.

persistence/lock-keeper-sql/src/api.rs
```diff
@@ -249,24 +249,30 @@ impl PostgresDB {
...
+            let actions = actions.iter().map(|a| *a as i64);
+
+            append_value_list(&mut query, actions)?;
+            let _ = query.push(" AND ");
+        }
 
         // Ensure account name matches, otherwise a client could fetch anyone's audit
         // events if they guess the request_id.
-        let _ = query.push("AND account_id=").push_bind(account_id.0);
+        let _ = query.push(" account_id=").push_bind(account_id.0);
```
picklenerd, line 267: I feel like these `AND` should be before the thing that's being ANDED. Is there a downside to removing the `AND` from line 249, keeping it here, and moving it to the front in the match block above?

lock-keeper/src/crypto/generic.rs
gatoWololo: Right, that would make sense! The current serialization format isn't great #237 and we might change it if we have extra cycles soon. I wasn't sure how much time to put into this, e.g. I could also change the other parsing code in the crypto module to use this abstraction. Is this worth iterating on now? I'm good either way.

lock-keeper-tests/src/test_suites/end_to_end/test_cases/retrieve.rs
gatoWololo: I like this idea. I have added the comment. This is an end to-end-test though, I don't think I have a way to check the server config from this code to check the configured max size?

lock-keeper/src/crypto/generic.rs
gatoWololo: Good point. Fixed.

persistence/lock-keeper-sql/src/api.rs
gatoWololo: This is way better :sweat_smile: done!

lock-keeper-tests/src/test_suites/end_to_end/test_cases/retrieve.rs
picklenerd: Yeah I don't think there's a way to do that. We could try to read the file but that's fragile.

lock-keeper/src/crypto/generic.rs
picklenerd: Nah we can do that separately!

picklenerd (approved):

lock-keeper-tests/src/test_suites/end_to_end/test_cases/retrieve.rs
gatoWololo: Right, that's why I didn't add a client-config max_size, since there is no way to ensure the client and server are synchronized on the max size.

lock-keeper/src/crypto/generic.rs
marsella: I have an ancient issue about doing something like this #292!

lock-keeper/src/crypto/data_blob.rs
```diff
@@ -0,0 +1,152 @@
...
+
+        // Check length before splitting.
+        if ds.len() > value.len() {
+            return Err(CryptoError::ConversionError);
+        }
+        // It is now safe to call split_at.
+        let (expected_ds, secret) = value.split_at(ds.len());
+        if ds != expected_ds {
```
marsella, line 82: nit: I would call the one from line 74 the `expected_ds` and this one the actual ds.

lock-keeper/src/crypto/data_blob.rs
marsella: Looks perfect!

marsella (approved):
This looks good to me, thanks for taking the time to think through this abstraction carefully!

PR #472, opened 2023-02-01
Minor Clean Ups

gatoWololo:
- Check returned rows on inserts. If wrong number found, this represents a bug in our implementation.
- Correct Documentation.
- Adds env logging support for running tests. Makes it easier to debug test runs.


picklenerd (approved):
LGTM after you fix CI

PR #474, opened 2023-02-03
Add key IDs to key operation contexts #471

### LINKED ISSUE #471: KeyID filter doesn't work when retrieving audit events.
To reproduce:

1. Generate a new key.
2. Use this key to sign some data.
3. Fetch the audit logs for this key.

If you look through all of the audit logs, you can see that the key_id field is null for all events.

Vidya said that we just need to add the key_id to the context so that it can be logged.

vidyaap:
Closes #471 

Also changed the sqlx-prepare task name, added session cache tests to the integration test return value, and added a key ID check to e2e audit event tests. I thought about making an integration test for this (instead of e2e), but we don't have the LockKeeperClient there to take key actions with the key before checking the audit for it.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -107,9 +107,10 @@ args = ["doc", "--all-features", "--no-deps", "--document-private-items", "--wor
 # -- Database Tasks --
 
 # Update the `sqlx-data.json` file
-[tasks.refresh-db-cache]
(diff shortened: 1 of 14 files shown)


gatoWololo (approved):
LGTM!

picklenerd (approved):

Issue #473, opened 2023-02-03
Encryption/Decryption Fails on Larger Types Due to Silent Conversion


gatoWololo:
My current work on #432 is blocked by a bug on our `Crypto` module. Basically, `DataBlob` fails to decrypt when the underlying data's length does not fit into a byte. This is due to a silent truncation by the `as` (type cast expression operator):
```rust
impl From<Secret> for Vec<u8> {
    fn from(secret: Secret) -> Self {
        // key len || key material || context len || context
        let ad: Vec<u8> = secret.context.to_owned().into();
        iter::once(secret.material.len() as u8)
            .chain(secret.material.to_owned())
            .chain(iter::once(ad.len() as u8))
            .chain(ad)
            .collect()
    }
}
```
Where the expression `as u8` in this function silently truncate larger lengths than what fits in a byte.

Solution:
- [x] Change the implementation to use `TryFrom` calls instead of `as` and propagate errors. This way, in the future, this will be an error instead of a silent failure.

gatoWololo:
This was closed by #475

PR #475, opened 2023-02-03
Explicit integer conversions for encrypting data.

gatoWololo:
Changes crypto module to use TryFrom calls instead of as and propagate errors. This way, in the future, this will be an error instead of a silent failure.

 Makes progress towards #473.

### DIFF
--- a/lock-keeper/src/crypto/generic.rs
+++ b/lock-keeper/src/crypto/generic.rs
@@ -7,12 +7,14 @@ use serde::{Deserialize, Serialize};
 use std::{iter, marker::PhantomData};
 use thiserror::Error;
 
-use std::convert::Infallible;
+use std::{convert::Infallible, num::TryFromIntError};
 use zeroize::{Zeroize, ZeroizeOnDrop};
 
 /// Errors that arise in the cryptography module.
 #[derive(Debug, Clone, Copy, Error)]
 pub enum CryptoError {
+    #[error("Conversion error")]
+    ConversionError,
     #[error("Encryption failed")]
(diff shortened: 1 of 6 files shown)


lock-keeper/src/crypto/generic.rs
```diff
@@ -180,9 +194,20 @@ impl<T> Encrypted<T>
...
+    // From"
     CryptoError: From<<T as TryFrom<Vec<u8>>>::Error>,
-    Vec<u8>: From<T>,
+
+    // These two bounds state:
+    // "Our T can be converted into a Vec<u8> via TryFrom...
+    Vec<u8>: TryFrom<T>,
+    // ...where the associated Error type of the TryFrom can be converted into a CryptoError via
+    // From"
+    CryptoError: From<<Vec<u8> as TryFrom<T>>::Error>,
```
gatoWololo, line 210: These bounds are pretty clever. So I added some documentation on them!

marsella (approved):
This looks good to me.

One thought: I think I didn't do serialization of associated data consistently, so in some places you'll see manual appending of the length of the associated data, and in other places we just serialize it directly without prepending the length. It would be better to implement the `Vec<u8>` transformations on `AssociatedData` directly so we don't leave the option open to forget to prepend the length. I think that would be a separate ticket+PR, though.

gatoWololo:
Sounds good! If you write that ticket we can make sure that gets done at some point :robot:

marsella:
> Sounds good! If you write that ticket we can make sure that gets done at some point robot

I made #486 to address this!

PR #481, opened 2023-02-07
Specify format of datetime when deserializing

vidyaap:
To address errors with date filters.

I'm not sure about requiring ISO format directly in lock keeper like this, but the issue with the date filters was that the ones being passed from another server were being deserialized as just strings instead of OffsetDatetime. That went away when I added the serde tag, but I'm not sure it really addresses the problem.


picklenerd:
I think this solves the maestro problem since maestro and lock keeper now agree on the format.

picklenerd (approved):

PR #485, opened 2023-02-10
Add date filters to CLI audit event function #434

### LINKED ISSUE #434: Add date range support to audit even command in client CLI
I skipped implementing date ranges in #272 since they depend on bson types which we won't be using after MongoDB is removed. It would be useful to include these in the CLI for testing purposes.

Implement the following commands:
- [ ] `audit before:[yyyy-mm-dd]` should use a before-date filter with midnight as the time
- [ ] `audit before:[ISO 8601 datetime]` should use a before-date filter with the date and time given in ISO 8601 format
- [ ] `audit after:[yyyy-mm-dd]` should use an after-date filter with midnight as the time
- [ ] `audit after:[ISO 8601 datetime]` should use an after-date filter with the date and time given in ISO 8601 format

vidyaap:
Closes #434 

Not sure if there is a cleaner way to do the error handling in the parse_date function?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ serde_json = "1.0.85"
 serde_with = "1"
 strum = { version = "0.24.1", features = ["derive"] }
 thiserror = "1"
-time = { version = "0.3.17", features = ["serde"]}
+time = { version = "0.3.17", features = ["parsing", "serde"]}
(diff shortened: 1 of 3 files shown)


picklenerd (approved):
LGTM! I like how it allows for timeless queries!

PR #489, opened 2023-02-15
Add no-config option to client cli

picklenerd:
This PR removes the default config path from the client CLI and adds the option to provide a URI and signing CA chain directly from the command line. This will make the CLI much more usable outside of lock keeper.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -144,27 +144,27 @@ alias = "ci-test"
 # Run the client CLI in interactive mode
 [tasks.cli]
 command = "cargo"
-args = ["run", "--bin", "lock-keeper-client-cli"]
+args = ["run", "--bin", "lock-keeper-client-cli", "--", "--config", "dev/config/local/Client.toml"]
 
 # Runs a CLI script to generate a retrieve a few keys
 [tasks.script-generate]
 command = "cargo"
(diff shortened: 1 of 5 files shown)


gatoWololo (approved):
Looks good! I found the `client_config` method a bit hard to parse. I tried refactoring it:
```rust
impl Cli {
    pub fn client_config(&self) -> anyhow::Result<Config> {
        if let Some(config) = &self.config {
            return Ok(Config::from_file(config, None)?);
        }

        let server_uri = self.server_uri.as_ref().ok_or(anyhow!(
            "If `config` argument is not provided, `uri` must be provided"
        ))?;
        let ca_chain = self.ca_chain.as_ref().ok_or(anyhow!(
            "If `config` argument is not provided, `ca_chain` must be provided"
        ))?;

        let client_auth = if let (Some(certificate_chain), Some(private_key)) =
            (&self.client_chain, &self.private_key)
        {
            Some(ClientAuth {
                certificate_chain: certificate_chain.clone(),
                private_key: Some(private_key.clone()),
            })
        } else {
            None
        };

        let config_file = ConfigFile {
            server_uri: server_uri.clone(),
            ca_chain: ca_chain.clone(),
            client_auth,
        };
        Ok(Config::from_config_file(config_file, None)?)
    }
}
```
Feel free to take any part of this if you like that better. Or just keep yours.

PR #493, opened 2023-03-07
Add check_session operation

picklenerd:


lock-keeper/src/types/audit_event.rs
```diff
@@ -99,6 +99,7 @@ const ALL_ACTIONS: &[ClientAction] = &[
     ClientAction::RetrieveSigningKey,
     ClientAction::RetrieveStorageKey,
     ClientAction::StoreServerEncryptedBlob,
+    ClientAction::CheckSession,
```
vidyaap, line 102: Minor, but do we need to add this if we don't make audit events for this action?

lock-keeper-tests/src/test_suites/end_to_end/test_cases/check_session.rs
```diff
@@ -0,0 +1,55 @@
...
+use crate::{
+    config::TestFilters,
+    error::Result,
+    run_parallel,
+    test_suites::end_to_end::{operations::authenticate, test_cases::init_test_state},
+    utils::TestResult,
+};
+
+pub async fn run_tests(config: &Config, filters: &TestFilters) -> Result<Vec<TestResult>> {
+    println!("{}", "Running generate tests".cyan());
```
vidyaap, line 14: Update string

vidyaap (commented):
LGTM! Just left a couple of small things

lock-keeper/src/types/audit_event.rs
vidyaap: My bad, I meant do we need to include it in "all actions" but actually this is probably better for consistency anyway.

lock-keeper-tests/src/test_suites/end_to_end/test_cases/check_session.rs
```diff
@@ -0,0 +1,55 @@
...
+    let client = authenticate(&state).await.result?;
+
+    client.logout().await.result?;
+
+    let res = client.check_session().await;
+    assert!(matches!(
+        res,
+        Ok(SessionStatus {
+            is_session_valid: false
+        })
```
vidyaap, line 51: Looking at this test, I'm wondering if check_status should return an error for a bad session rather than a false within the successful response? e.g. for status codes, or if we wanted to match on the whole result somewhere?

vidyaap: Never mind, based on what we talked about, the true/false works better!

vidyaap (approved):

PR #500, opened 2023-03-23
Implement delete key #490

### LINKED ISSUE #490: Add Lock-Keeper Endpoint for Key Deletion
Currently there is no way for a client to request a stored key to be deleted. This would be useful for other projects that rely on Lock-Keeper.

Tasks:
- [ ] Implement a new server endpoint for deleting a key specified by a key ID.
- [ ] Implement database method for deleting a key based on a key ID. The account ID should match the current user.
- [ ] Implement client code for calling this endpoint.
- [ ] Add integration and end-to-end test for this new endpoint.

vidyaap:
Closes #490


persistence/lock-keeper-sql/src/api.rs
```diff
@@ -359,6 +367,32 @@ impl PostgresDB {
         }
     }
 
+    /// This function verifies the user_id and key type matches. Otherwise will
+    /// return a IncorrectKeyMetadata error.
+    #[instrument(skip_all, err(Debug), fields(account_id=?account_id, key_id=?key_id))]
+    pub(crate) async fn delete_secret_impl(
```
vidyaap, line 373: Do we want to use the secret type filter here, and also do we want the more comprehensive key count check like we have at the end of get_secret?

picklenerd: IMO the secret type isn't important. I think the count check is fine too. Key ID is unique so this is either going to affect 0 or 1 rows.

picklenerd (approved):
LGTM!

picklenerd (approved):

PR #520, opened 2023-06-01
Add optimized query for server encrypted blobs

picklenerd:
This PR adds a separate query for server-encrypted blobs with a much faster database query.
It also adds blob support to the client CLI.

This new query makes a few changes:
1. Removes the `UPDATE` from blob queries which makes this query a simple read that can happen in parallel.
2. `INNER` join instead of `LEFT` join which filters results before doing the join.
3. Adds an index on (key_id, account_id) which should speed up the WHERE clause.
4. Replaces `LIKE` with `=` for the `secret_type` which could speed up comparison in some cases.

The old query takes around 600ms on average. The new query takes around 300µs on average.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -142,7 +142,7 @@ alias = "ci-test"
 # Run the client CLI in interactive mode
 [tasks.cli]
 command = "cargo"
-args = ["run", "--bin", "lock-keeper-client-cli", "--", "--config", "dev/config/local/Client.toml"]
+args = ["run", "--bin", "lock-keeper-client-cli", "--", "--config", "dev/config/local/Client.toml", "${@}"]
 
 # Runs a CLI script to generate a retrieve a few keys
 [tasks.script-generate]
--- a/bin/lock-keeper-client-cli/src/cli_command/retrieve_blob.rs
+++ b/bin/lock-keeper-client-cli/src/cli_command/retrieve_blob.rs
@@ -1,0 +1,66 @@
+use std::time::{Duration, SystemTime};
+
+use crate::{cli_command::CliCommand, state::State};
+use anyhow::Error;
+use async_trait::async_trait;
+use lock_keeper_client::LockKeeperClient;
+
+#[derive(Debug)]
+pub struct RetrieveBlob {
+    name: String,
+}
+
+#[async_trait]
+impl CliCommand for RetrieveBlob {
+    async fn execute(self: Box<Self>, state: &mut State) -> Result<Duration, Error> {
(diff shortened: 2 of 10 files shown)


Makefile.toml
```diff
@@ -142,7 +142,7 @@ alias = "ci-test"
 # Run the client CLI in interactive mode
 [tasks.cli]
 command = "cargo"
-args = ["run", "--bin", "lock-keeper-client-cli", "--", "--config", "dev/config/local/Client.toml"]
+args = ["run", "--bin", "lock-keeper-client-cli", "--", "--config", "dev/config/local/Client.toml", "${@}"]
```
gatoWololo, line 145: What does this do? :open_mouth:

bin/lock-keeper-client-cli/src/cli_command/retrieve_blob.rs
```diff
@@ -0,0 +1,66 @@
...
+            &credentials.account_name,
+            &credentials.password,
+            &state.config,
+        )
+        .await
+        .result?;
+
+        let entry = state.get_key_id(&self.name)?;
+
+        let now = SystemTime::now();
```
gatoWololo, line 29: Any reason you use `SystemTime` instead of `std::time::Instant`?

gatoWololo (approved):
Looks goods! Please fix CI issues.

bin/lock-keeper-client-cli/src/cli_command/retrieve_blob.rs
picklenerd: No reason

Makefile.toml
picklenerd: > Looks goods! Please fix CI issues.

Sorry this was supposed to be a draft 🤦

picklenerd: Lets you pass command-line arguments through cargo make

picklenerd:
> Looks goods! Please fix CI issues.

Sorry this was supposed to be a draft 🤦

bin/lock-keeper-client-cli/src/cli_command/retrieve_blob.rs
picklenerd: This is all copy/pasted for the most part.

gatoWololo:
Looks good, and you don't really need my approval again.

Could you add a few words to either a comment on this PR or to the code on why the new queries are faster and how much faster they are? This is both for documentation and for my own knowledge :)

Issue #496, opened 2023-03-17
Allow user to create a `SigningPublicKey` from bytes or pem file.


gatoWololo:
We are planning to use Lock-Keeper's crypto module as a building block for other projects. Currently it is not possible to build a standalone public key (`SigningPublicKey`). We would like to allow the user to build a `SigningPublicKey` from a `.pem` file.

- [ ] Allow constructing a `SigningPublicKey` from a PEM-encoded string.

gatoWololo:
Closed by #525

PR #525, opened 2023-06-09
Add additional methods/types for general use of crypto module.

### LINKED ISSUE #524: Crypto Module: Add Necessary Types/Methods for General Use
- Add `Signature` as a possible error variant in `CryptoError`.
- Add methods to `SigningPublicKey` so it be used standalone to verify signatures.
- Add constructor to `SigningPublicKey` to create it from a `pem` file.
- Allow serializing `Signature` to der format.
- Add `RecoverableSignature` type.
- Add stand-alone `SigningPrivateKey` type.

gatoWololo:
Closes #524 and #496

See those issues for the list of changes this adds.

@marsella I had a few questions about these changes: I tried to leave the existing crypto code intact, and instead add new types and methods. I'm unsure if this is the right approach and instead I could merge some types together more. I will mark these as comments on the PR.

I'm happy to change the code as needed to meet your cryptography standards :bow:

### DIFF
--- a/lock-keeper/src/crypto/signing_key.rs
+++ b/lock-keeper/src/crypto/signing_key.rs
@@ -2,18 +2,20 @@ use crate::{
     crypto::{
         generic::EncryptionKey,
         signing_key::generation_types::{CLIENT_GENERATED, IMPORTED, SERVER_GENERATED},
-        RemoteStorageKey,
+        RemoteStorageKey, SigningPrivateKey,
     },
     types::database::account::UserId,
     LockKeeperError,
 };
-use k256::ecdsa::{
-    self,
-    signature::{Signer, Verifier},
-    SigningKey, VerifyingKey,
+use k256::{
+    ecdsa::{self, signature::DigestVerifier, VerifyingKey},
+    elliptic_curve::sec1::ToEncodedPoint,
 };
 use rand::{CryptoRng, RngCore};
 use serde::{Deserialize, Serialize};
+use sha3::Digest;
+use std::str::FromStr;
+use tracing::error;
 use zeroize::ZeroizeOnDrop;
 
 use super::{generic::AssociatedData, CryptoError, Encrypted, Export, KeyId, StorageKey};
@@ -31,7 +33,7 @@ pub trait Signable: AsRef<[u8]> {
         &self,
         public_key: &SigningPublicKey,
         signature: &Signature,
-    ) -> Result<(), LockKeeperError>;
+    ) -> Result<(), CryptoError>;
 }
 
 /// Wrapper used to declare arbitrary bytes as [`Signable`].
@@ -52,18 +54,15 @@ impl AsRef<[u8]> for SignableBytes {
 
 impl Signable for SignableBytes {
     fn sign(&self, signing_key_pair: &SigningKeyPair) -> Signature {
-        Signature(signing_key_pair.signing_key.sign(&self.0))
+        signing_key_pair.signing_key.sign(&self.0)
     }
 
     fn verify(
         &self,
         public_key: &SigningPublicKey,
         signature: &Signature,
-    ) -> Result<(), LockKeeperError> {
-        Ok(public_key
-            .0
-            .verify(&self.0, &signature.0)
-            .map_err(|_| CryptoError::VerificationFailed)?)
+    ) -> Result<(), CryptoError> {
+        public_key.verify(&self.0, signature)
     }
 }
 
@@ -78,23 +77,17 @@ pub mod generation_types {
 /// pair.
 ///
 /// This can be generated locally by the client or remotely by the server.
-#[derive(Debug, Clone, PartialEq, Eq, ZeroizeOnDrop)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct SigningKeyPair {
-    signing_key: SigningKey,
-    #[zeroize(skip)]
+    signing_key: SigningPrivateKey,
     context: AssociatedData,
 }
-
-/// The public component of an ECDSA signing key, and context about the key.
-#[allow(unused)]
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-pub struct SigningPublicKey(VerifyingKey);
 
 impl SigningKeyPair {
     /// Create a new `SigningKeyPair` with the given associated data.
     fn generate(rng: &mut (impl CryptoRng + RngCore), context: &AssociatedData) -> Self {
         Self {
-            signing_key: SigningKey::random(rng),
+            signing_key: SigningPrivateKey::generate(rng),
             context: context.clone(),
         }
     }
@@ -106,7 +99,7 @@ impl SigningKeyPair {
 
     /// Retrieve the public portion of the key.
     pub fn public_key(&self) -> SigningPublicKey {
-        SigningPublicKey(self.signing_key.verifying_key())
+        self.signing_key.public_key()
     }
 
     /// Retrieve the context associated with the signing key.
@@ -154,8 +147,7 @@ impl SigningKeyPair {
             .with_str(IMPORTED);
 
         let signing_key = Self {
-            signing_key: SigningKey::from_bytes(key_material)
-                .map_err(|_| CryptoError::ConversionError)?,
+            signing_key: SigningPrivateKey::from_bytes(key_material)?,
             context: context.clone(),
         };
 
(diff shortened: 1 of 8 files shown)


lock-keeper/src/crypto/signing_key.rs
```diff
@@ -194,6 +188,54 @@ impl SigningKeyPair {
...
 }
 
+/// The public component of an ECDSA signing key.
+#[allow(unused)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct SigningPublicKey(pub VerifyingKey);
+
+impl SigningPublicKey {
+    /// Verify a `Signature` using this public key. This function will hash the
+    /// message using ??? TODO.
```
gatoWololo, line 198: I could use some help: What hash is used by the `verify` method here?

lock-keeper/src/crypto/signing_private_key.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /// Get `v` (recovery ID) component of signature.
+    pub fn v(&self) -> u8 {
+        self.recovery_id
+    }
+}
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct SigningPrivateKey(ecdsa::SigningKey);
```
gatoWololo, line 64: I created a separate `SigningPrivateKey` type here, totally separate from the existing `SigningKeyPair`. We can change our `SigningKeyPair` to use this new type. E.g:
```diff
pub struct SigningKeyPair {
-    signing_key: ecdsa::SigningKey,
+    signing_key: SigningPrivateKey
}
```
This would allow us to merge some of the functions while still allowing both `SigningPrivateKey` and `SigningKeyPair` to be used independently.

lock-keeper/src/crypto/signing_private_key.rs
```diff
@@ -0,0 +1,187 @@
...
+impl SigningPrivateKey {
+    // TODO: What format are these bytes expected in?
+    pub fn from_bytes(bytes: &[u8]) -> Self {
+        Self(ecdsa::SigningKey::from_bytes(bytes).unwrap())
+    }
+
+    // TODO: What format are these bytes in?
+    pub fn as_bytes(&self) -> Vec<u8> {
+        self.0.to_bytes().as_slice().to_vec()
+    }
```
gatoWololo, line 75: It is unclear to me from the docs what format, if any, these function use for the serialized representation.

lock-keeper/src/crypto/signing_private_key.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub fn sign_prehash(&self, hashed: impl AsRef<[u8]>) -> Result<Signature, CryptoError> {
+        let signature = self
+            .0
+            .sign_prehash(hashed.as_ref())
+            .map_err(CryptoError::Signature)?;
+        Ok(Signature(signature))
+    }
+
+    /// Sign a pre-hashed message returning a `[RecoverableSignature]`
+    pub fn sign_prehash_recoverable(
```
gatoWololo, line 97: We could fold these methods into our `Signable` trait if we wanted.  There is a bit of duplication since we have multiple variants of this function: sign, sign_prehash, sign_recoverable, sign_prehash. Not sure if there is a way to minimize this.

lock-keeper/src/crypto/signing_private_key.rs
gatoWololo, line 64: Our other products have code written which expect a stand-alone signing private key, that's why I created this separate type instead of using our existing `SigningKeyPair`. Also `SigningKeyPair` expects `AssociatedContext` which I wasn't sure what to give it.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -194,6 +188,54 @@ impl SigningKeyPair {
     }
 }
 
+/// The public component of an ECDSA signing key.
+#[allow(unused)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct SigningPublicKey(pub VerifyingKey);
```
gatoWololo, line 194: This was just moved down from higher in the file.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -9,11 +9,13 @@ use crate::{
 };
```
gatoWololo, line 9: This file has gotten really big. I avoided moving types around to make the PR easier to understand. But we may wanna break the file down.

lock-keeper/src/crypto/signing_key.rs
marsella: Yikes, this is really hard to track down. The verify method is parameterized by the type of the signature; the signature should have an associated type called `Digest` that'll implement a hash function. All the verify methods are generic over the hash. 
The signature type we use is `ecdsa::Signature`, which is a type alias for `ecdsa_core::Signature<Secp256k1>`. I poked around a bit and I think that it uses `sha2-256` for the hash function -- [see this instantiation](https://docs.rs/k256/0.11.0/src/k256/ecdsa.rs.html#92-96) of `DigestPrimitive` for the `Secp256k1` curve.

lock-keeper/src/crypto/signing_private_key.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    /// Get `r` component of signature.
+    pub fn r(&self) -> [u8; 32] {
+        self.signature.0.split_bytes().0.into()
+    }
+
+    /// Get `s` component of signature.
+    pub fn s(&self) -> [u8; 32] {
+        self.signature.0.split_bytes().1.into()
+    }
```
marsella, line 55: question: Are these methods used somewhere? I don't love giving direct access to these; it's not really a security issue but, like, the whole purpose of a signature is to verify it, and since we provide functions to do that, it would be ideal to restrict access to these internal types.

lock-keeper/src/crypto/signing_private_key.rs
marsella: Implementing `SigningKeyPair` in terms of your `SigningPrivateKey` seems fine to me. The associated context for the key pair was (in the old protocol) intended to be all the versioning and context info that we would want to validate when we pulled + decrypted one of our keys out of storage. I don't know the maestro protocol well enough to understand if we're maintaining state like that, but it's fine to use this separate type for the Maestro use case.

marsella (changes requested):
Thanks Omar, this looks pretty good. The mechanics of this look pretty good, I have a few important documentation-related comments that I think we need for the contexts in which this'll be used. It would also be helpful to chat about the structure of this for a few minutes.

marsella:
Also, general request for you: could you please wait to squash your commits until we are done with the review cycle? I know it makes rebasing more annoying, but it's going to be much easier for me to re-review if I can see incremental changes via the commit history.

lock-keeper/src/crypto/signing_private_key.rs
gatoWololo: Cool. Makes sense.

lock-keeper/src/crypto/signing_key.rs
gatoWololo: Thanks! This was really helpful.

lock-keeper/src/crypto/signing_private_key.rs
gatoWololo: For the sake of simplicity, I will not do this.

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -194,6 +188,54 @@ impl SigningKeyPair {
...
+        signature: &Signature,
+    ) -> Result<(), CryptoError> {
+        self.0.verify(message.as_ref(), &signature.0).map_err(|e| {
+            error!("{e}");
+            CryptoError::VerificationFailed
+        })
+    }
+
+    /// Verify a `Signature` using this public key. This function expects `hash`
+    /// to be an already hashed message.
```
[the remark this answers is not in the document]
gatoWololo: I ended up not doing this since I got rid of all the `.*pre_hash.*` methods. Let me know if this works :)

lock-keeper/src/crypto/signing_key.rs
```diff
@@ -77,18 +77,17 @@ pub mod generation_types {
 /// pair.
 ///
 /// This can be generated locally by the client or remotely by the server.
-#[derive(Debug, Clone, PartialEq, Eq, ZeroizeOnDrop)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
gatoWololo, line 80: Like we have talked about in the zeroize tickets, only the underlying type `SigningPrivateKey` implements `ZeroizeOnDrop` so this type no longer needs it.

lock-keeper/src/crypto/signing_private_key.rs
```diff
@@ -0,0 +1,195 @@
...
+        Self(ecdsa::SigningKey::random(rng))
+    }
+
+    /// Retrieve the public portion of the key.
+    pub fn public_key(&self) -> SigningPublicKey {
+        SigningPublicKey::from(self.0.verifying_key())
+    }
+
+    /// Sign a message returning a `[Signature]`. This function will hash the
+    /// message using SHA3 keccak256.
```
marsella, line 110: nit: the standard is SHA3-256; the algorithm selected by the standard is the Keccak family of hash functions. It might be slightly more standard to write this (here and below) like "SHA3-256 (Keccak)"

lock-keeper/src/crypto/signing_key.rs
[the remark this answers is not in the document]
marsella: I think that's okay for now. What's the revealed module structure of this library, though? It would be nice to add some high-level documentation _somewhere_ to be explicit that this is a low-level cryptographic library and it does not do any validation or verification of messages; that must be handled by the calling application.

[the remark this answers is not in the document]
marsella: This is kind of awkward because this module is intended to be resuable, but the `lock_keeper::crypto` module also includes a bunch of application-specific stuff for the old protocol.

It would be nice if we could bundle the bits we use externally in a separate module or something so it was clear which parts are "use this to consolidate our cryptography" and which are "outdated crypto for a protocol you're not using".

lock-keeper/src/crypto/signing_private_key.rs
```diff
@@ -0,0 +1,195 @@
...
+    pub fn s(&self) -> [u8; 32] {
+        self.signature.0.split_bytes().1.into()
+    }
+
+    /// Get `v` (recovery ID) component of signature.
+    ///
+    /// Useful for interoperability with other libraries.
+    pub fn v(&self) -> u8 {
+        self.recovery_id
+    }
```
marsella, line 75: suggestion: we had talked about consolidating these into a single `into_parts` method.

lock-keeper/src/crypto/signing_key.rs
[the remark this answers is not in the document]
gatoWololo: I agree, I tried really hard not to move things around for the sake of not bloating this PR. I made a separate issue to address this. See #526 and lmk if I should add anything else to that new issue.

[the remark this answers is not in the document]
marsella: Yep, that was a good call. Thanks for writing up the issue.

marsella (approved):
Looks good, thanks!

PR #531, opened 2023-07-07
530 opaque password encryption

### LINKED ISSUE #530: OPAQUE password encryption for key servers

tposavec:
Closes #530 

The branch for this PR should have been named 530-opaque-password-encryption


lock-keeper/src/crypto/generic.rs
```diff
@@ -92,14 +92,40 @@ impl AssociatedData {
...
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct Encrypted<T> {
     pub(super) ciphertext: Vec<u8>,
     pub(super) associated_data: AssociatedData,
     pub(super) nonce: chacha20poly1305::Nonce,
     pub(super) original_type: PhantomData<T>,
 }
 
+//////
```
gatoWololo, line 103: Could you get rid of these extra comments lines. The first line of the doc string should start with `///` followed by the one sentence description.

lock-keeper/src/crypto/generic.rs
```diff
@@ -92,14 +92,40 @@ impl AssociatedData {
...
+///
+/// An implementation of the `std::fmt::Debug` trait for Password.
+///
+/// The Debug trait controls how a value is formatted when using the `println!`
+/// or `format!` macros with the {:?} format specifier.
+impl<T> std::fmt::Debug for Encrypted<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Encrypted")
+            .field(
+                "\nCiphertext",
```
gatoWololo, line 113: I'm not sure I like the newlines and tab characters in this Debug impl. Debug usually doesn't have any formatting. (`Display` is usually used for prettier printting of types.). I think its fine if it is all in a single line but I don't feel too strongly about it.

lock-keeper/src/crypto/generic.rs
```diff
@@ -369,6 +395,140 @@ impl TryFrom<Vec<u8>> for Secret {
...
+        Self {
+            material: iter::repeat_with(|| rng.gen()).take(len).collect(),
+            context,
+        }
+    }
+
+    //////
+    ///
+    /// Create a new password from its constituent parts.
+    pub(super) fn from_parts(password_material: Vec<u8>, context: AssociatedData) -> Option<Self> {
```
gatoWololo, line 431: Even for simple error cases like this, you should use a `-> Result<Self, CryptoError>`. You can make a new `EmptyPassword` variant on `CryptoError`

lock-keeper/src/crypto/generic.rs
```diff
@@ -369,6 +395,140 @@ impl TryFrom<Vec<u8>> for Secret {
...
+//////
+///
+/// An implementation of the `std::fmt::Debug` trait for Password.
+///
+/// The Debug trait controls how a value is formatted when using the `println!`
+/// or `format!` macros with the {:?} format specifier.
+impl std::fmt::Debug for Password {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Password")
+            .field("material", &format_args!("{:02x?\n}", &self.material))
```
gatoWololo, line 470: Should this actually have a Debug impl that prints the material? Or should it be "REDACTED"?

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,262 @@
...
+    generic::{self, AssociatedData, CryptoError},
+    Encrypted, PasswordKey,
+};
+
+use super::Export;
+
+//////
+///
+/// The [`Password`] type.
+
```
gatoWololo, line 21: Extra blank space

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,262 @@
...
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, ZeroizeOnDrop)]
+pub struct Password(pub(super) generic::Password);
+
+impl std::fmt::Debug for Password {
```
gatoWololo, line 25: Similar question about printing "REDACTED" to avoid password accidentally printing to logs?

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,262 @@
...
+    ///
+    /// # Errors
+    ///
+    /// * Returns a `LockKeeperError` if there is an error during encryption,
+    ///   decryption, or verification.
+    #[test]
+    fn password_encrypt_decrypt() -> Result<(), LockKeeperError> {
+        let mut rng = rand::thread_rng();
+
+        const KEY_SERVER_NAME: &str = "test_key_server_1";
```
gatoWololo, line 223: This could be a constant defined at the `mod test` level and both unit tests use it.

lock-keeper/src/crypto/password_key.rs
```diff
@@ -0,0 +1,163 @@
...
+        let password_encryption_key_from_file_bytes =
+            password_encryption_key_from_file.0.clone().into_bytes();
+
+        println!(
+            "\nPassword encryption key from a file is...\n{:02x?}\n\n",
+            password_encryption_key_from_file_bytes
+        );
+
+        // clean up by deleting the temp password key file
+        fs::remove_file(PASSWORD_KEY_FILENAME).expect("Failed to remove a temp password key file.");
```
gatoWololo, line 152: If this test fails for some reason, this line will never execute and the file will remain. The best way to handle this is to make a second function which executes your code and uses `?` to return errors, then the caller unit-test can always clean up the file and then match on the result of the first function call. Lmk if that makes sense.

gatoWololo (changes requested):
Overall the Rust code looks good! I have small comments in a few things.

lock-keeper/src/crypto/password_key.rs
```diff
@@ -0,0 +1,163 @@
...
+use rand::{CryptoRng, RngCore};
+use std::path::Path;
+use zeroize::{Zeroize, ZeroizeOnDrop};
+
+//////
+///
+/// The [`PasswordKey`] is a default-length symmetric encryption key
+/// for an AEAD scheme. It can be used to securely encrypt passwords.
+#[derive(Debug, Clone, Zeroize, ZeroizeOnDrop)]
+pub struct PasswordKey(pub(super) EncryptionKey);
```
marsella, line 16: todo: We can discuss but I don't think the existing `EncryptionKey` type is a better choice here than using the `chacha20poly1305` key type directly. That type was designed for the old protocol and we don't need the additional associated data it contains.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,262 @@
...
+    /// * A tuple containing the newly created `Self` instance and the encrypted
+    ///   representation.
+    /// * Returns an error of type `LockKeeperError` if the encryption process
+    ///   fails.
+    pub fn encrypt(
+        rng: &mut (impl CryptoRng + RngCore),
+        password_plaintext: &[u8],
+        password_encryption_key: &PasswordKey,
+        key_server_name: Vec<u8>,
+    ) -> Result<(Self, Encrypted<Self>), LockKeeperError> {
```
marsella, line 103: thought: I'm not sure what your use case is but it's not clear to me that you actually need to return the `Password` type here if you're just encrypting bytes. I would expect that either the calling application is dealing with `Password`s from end to end (like, they use the `Password::new` method to create one and pass a `Password` to this method) or they would just use bytes and never have to know about the `Password` type at all. The former is probably more misuse resistant but will also be more work to make sure that type provides all the functionality you need from it.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,262 @@
...
+        let context = AssociatedData::new().with_bytes(key_server_name);
```
marsella, line 104: todo: Let's discuss associated data.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,262 @@
...
+}
+
+impl TryFrom<Vec<u8>> for Password {
+    type Error = CryptoError;
+    fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
+        Ok(Password(value.try_into()?))
+    }
+}
+
+impl Encrypted<Password> {
```
marsella, line 47: todo: I think we might want to make a new record type for your specific use case that doesn't use the `PhantomData` type because it doesn't provide the guarantees I actually wanted. It'll be similar to the existing `Encrypted<T>` but without the `original_type` field.

lock-keeper/src/crypto/generic.rs
```diff
@@ -369,6 +395,140 @@ impl TryFrom<Vec<u8>> for Secret {
...
+/// `Vec<u8>`
+impl TryFrom<Password> for Vec<u8> {
+    type Error = CryptoError;
+
+    fn try_from(password: Password) -> Result<Self, Self::Error> {
+        // password len || password material || context len || context
+        let ad: Vec<u8> = password.context.to_owned().into();
+
+        let password_length = u16::try_from(password.material.len())
+            .map_err(|_| CryptoError::CannotEncodeDataLength)?;
```
marsella, line 488: question: I'm guessing this is reasonable but can you confirm that in the application where we'll use this, we don't allow passwords of more than 65k bits?

lock-keeper/src/crypto/generic.rs
```diff
@@ -369,6 +395,140 @@ impl TryFrom<Vec<u8>> for Secret {
...
+    }
+
+    //////
+    ///
+    /// Retrieve the password material.
+    ///
+    /// Return a reference to the underlying bytes of the password.
+    /// This should only be used to print the password.
+    pub(super) fn borrow_material(&self) -> &[u8] {
+        &self.material
```
marsella, line 457: issue: This is usually called [`as_ref`](https://doc.rust-lang.org/std/convert/trait.AsRef.html) but also we should not be printing passwords in the logs. Reasonable to use for debugging as you write this PR but it needs to be removed before merging.

marsella (changes requested):
First of all, I'd like to confirm I have the context correct. I'm not very familiar with the Maestro design, so feel free to point me toward existing documentation; I think it uses a OPAQUE in a non-standard way.
Typically, OPAQUE is used in a scenario where you have a human user who can remember a password. The benefit of the protocol is twofold:
1. it can produce a suitably random shared cryptographic key even when the user input has insufficient entropy (like most passwords)
2. It authenticates a user without having to send their password (or some thinly-veiled version of it) to the server.
I believe Swan told me we're using OPAQUE in the institutional setting, where the passwords are chosen by a machine (not a human) and stored in computer memory (not a brain). Is that correct? That is, is this password database stored on the "client" (e.g. institutional service provider) side? I ask because one primary benefit of OPAQUE is that you do not have to store passwords in a database.

To start off, I'd like to share a bit about the structure and history of this module. Much of the code in this key-mgmt repo was originally written for a different protocol from the one we're now using. We build half of it then abandoned it for other projects. I don't think the structure of this crypto module is particularly well suited to the current protocol that we're working on (or indeed, to the previous one, but again we stopped building that product so I did not go back and refactor it to be better). 

In particular, when I write cryptographic libraries, I try to aim them toward either the high-expertise context (a more generic crypto library that you could mess up if you use it wrong) or the low-expertise context (an application-specific library that is carefully tailored to the application). This module does a bit of both, with the `generic` module defining a wrapper around an encryption library and the other modules using that in application-specific ways, but still allowing the caller to make decisions about associated data (easy to mess up). 

Part of the reason for having a generic + an application specific section here was because we had a lot of types that were fundamentally cryptographic keys, but were used for different things. In that context, it was useful to put shared functionality into the `generic.rs` module and reuse it across the other modules. Moving forward, I don't think there's much value to repeating that structure with `Password`s. We also defined some wrappers around different key types to add things like associated data that were used for various purposes in the old protocol; these wrappers are not inherently useful outside of that context.

There are also some inefficiencies with the `Encrypted` abstraction itself. In particular, the `PhantomData` is not useful because it doesn't get serialized, so if you stick this encrypted blob in a database, you can try to decrypt it into any type. The point of that `PhantomData` was to bind encrypted blobs to specific types but that doesn't work here. So, for example, when you extract an `Encrypted` out of the database, you have to decide what its `T` is; the type does not automatically inform you of it.

The takeaway of all this is that building directly off the existing work is not ideal. I would like to touch base with you and sketch out a self-contained passwords module that uses the same encryption scheme as the existing work but doesn't bind itself to the extant types and architecture. I'm guessing that involve moving all the password-related stuff from `generic.rs` to be inline in the existing `password.rs` module, but let's discuss before you make any significant changes.

I made a few comments inline; would recommend looking through those but again, let's talk before you put it a bunch of effort to change stuff.

lock-keeper/src/crypto/generic.rs
marsella: I would default to redacting.

lock-keeper/src/crypto/password_key.rs
```diff
@@ -0,0 +1,163 @@
...
+    ///
+    /// Generate a new 32-byte [`PasswordKey`].
+    pub fn generate(rng: &mut (impl CryptoRng + RngCore)) -> Self {
+        Self(EncryptionKey::new(rng))
+    }
+
+    //////
+    ///
+    /// Returns the [`PasswordKey`] stored in a file.
+    pub fn read_from_file(path: impl AsRef<Path>) -> Result<Self, LockKeeperError> {
```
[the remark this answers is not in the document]
marsella: On second look, there might not be a standardized format for chacha keys.

tposavec:
> First of all, I'd like to confirm I have the context correct. I'm not very familiar with the Maestro design, so feel free to point me toward existing documentation; I think it uses a OPAQUE in a non-standard way. Typically, OPAQUE is used in a scenario where you have a human user who can remember a password. The benefit of the protocol is twofold:
> 
>     1. it can produce a suitably random shared cryptographic key even when the user input has insufficient entropy (like most passwords)
> 
>     2. It authenticates a user without having to send their password (or some thinly-veiled version of it) to the server.
>        I believe Swan told me we're using OPAQUE in the institutional setting, where the passwords are chosen by a machine (not a human) and stored in computer memory (not a brain). Is that correct? That is, is this password database stored on the "client" (e.g. institutional service provider) side? I ask because one primary benefit of OPAQUE is that you do not have to store passwords in a database.
> 
> 
> To start off, I'd like to share a bit about the structure and history of this module. Much of the code in this key-mgmt repo was originally written for a different protocol from the one we're now using. We build half of it then abandoned it for other projects. I don't think the structure of this crypto module is particularly well suited to the current protocol that we're working on (or indeed, to the previous one, but again we stopped building that product so I did not go back and refactor it to be better).
> 
> In particular, when I write cryptographic libraries, I try to aim them toward either the high-expertise context (a more generic crypto library that you could mess up if you use it wrong) or the low-expertise context (an application-specific library that is carefully tailored to the application). This module does a bit of both, with the `generic` module defining a wrapper around an encryption library and the other modules using that in application-specific ways, but still allowing the caller to make decisions about associated data (easy to mess up).
> 
> Part of the reason for having a generic + an application specific section here was because we had a lot of types that were fundamentally cryptographic keys, but were used for different things. In that context, it was useful to put shared functionality into the `generic.rs` module and reuse it across the other modules. Moving forward, I don't think there's much value to repeating that structure with `Password`s. We also defined some wrappers around different key types to add things like associated data that were used for various purposes in the old protocol; these wrappers are not inherently useful outside of that context.
> 
> There are also some inefficiencies with the `Encrypted` abstraction itself. In particular, the `PhantomData` is not useful because it doesn't get serialized, so if you stick this encrypted blob in a database, you can try to decrypt it into any type. The point of that `PhantomData` was to bind encrypted blobs to specific types but that doesn't work here. So, for example, when you extract an `Encrypted` out of the database, you have to decide what its `T` is; the type does not automatically inform you of it.
> 
> The takeaway of all this is that building directly off the existing work is not ideal. I would like to touch base with you and sketch out a self-contained passwords module that uses the same encryption scheme as the existing work but doesn't bind itself to the extant types and architecture. I'm guessing that involve moving all the password-related stuff from `generic.rs` to be inline in the existing `password.rs` module, but let's discuss before you make any significant changes.
> 
> I made a few comments inline; would recommend looking through those but again, let's talk before you put it a bunch of effort to change stuff.

Thanks, @marsella!  This is good context info to have & understand.   I made some incorrect assumptions/decisions based on the current LK crypto lib design just by looking at the codebase.  Should have talked to you sooner to get the full context, including the historical background & future plans for the LK crypto lib.  Anyway, as we discussed on the call this morning, I will refactor my code based on your comments.

PR #536, opened 2023-07-20
Move Sharding to Lock-Keeper.

### LINKED ISSUE #535: Add Sharding Code to Lock-Keeper
In order to consolidate our Crypto code in one place. We will be moving the current sharding and encryption code from a different repository into our crypto module here.

- [ ] Move Sharding code into Lock-Keeper.

gatoWololo:
Closes #535

### DIFF
--- a/lock-keeper/Cargo.toml
+++ b/lock-keeper/Cargo.toml
@@ -34,12 +34,15 @@ zeroize.workspace = true
 
 
 # Other dependencies
+aes-gcm = "0.10.2" # Used for encryption/decryption of sharded keys.
 anyhow = "1.0"
 chacha20poly1305 = "0.10"
 hkdf = "0.12"
 k256 = { version = "0.13.1", features = ["ecdsa", "pem", "serde"] }
 sha3 = "0.10"
 hex = "0.4"
+# vsss-rs used for shamir f.
+vsss-rs = "3.2"
 
 [build-dependencies]
 tonic-build = "0.8.0"
--- a/lock-keeper/src/crypto/generic.rs
+++ b/lock-keeper/src/crypto/generic.rs
@@ -15,23 +15,17 @@ use zeroize::{Zeroize, ZeroizeOnDrop};
 /// Errors that arise in the cryptography module.
 #[derive(Debug, Error)]
 pub enum CryptoError {
-    #[error("Conversion error")]
-    ConversionError,
-    #[error("Encryption failed")]
-    EncryptionFailed,
-    #[error("Decryption failed")]
-    DecryptionFailed,
-    #[error("Key derivation failed: {0}")]
-    KeyDerivationFailed(hkdf::InvalidLength),
-    #[error("Signature generation failed: {0}")]
-    Signature(#[from] ecdsa::signature::Error),
-    #[error("RNG failed")]
-    RandomNumberGeneratorFailed,
     /// Length of data too large for our integer data type.
     #[error("Encryption/Decryption failed due to data length.")]
     CannotEncodeDataLength,
-    #[error("Signature did not verify")]
-    VerificationFailed,
+    #[error("Failed to combine shards into key: {0}")]
+    CombineShardsFailed(String),
+    #[error("Conversion error")]
+    ConversionError,
+    #[error("Decryption failed")]
+    DecryptionFailed,
+    #[error("Encryption failed")]
+    EncryptionFailed,
     /// The `impl<T> Encrypted<T>` has some trait bounds for converting a
     /// `TryFrom::Error` associated type into a CryptoError.
     /// Rust automatically implements `TryFrom<T, Error=Infallible>` when
@@ -39,6 +33,24 @@ pub enum CryptoError {
     /// From<CryptoError> for Infallible` holds true in this case.
     #[error(transparent)]
     Infallible(#[from] Infallible),
+    #[error("Key derivation failed: {0}")]
+    KeyDerivationFailed(hkdf::InvalidLength),
+    #[error("Failed to convert scalar to non-zero scalar.")]
+    NonZeroScalarConversion,
+    #[error("RNG failed")]
+    RandomNumberGeneratorFailed,
+    #[error("Incorrect size for seal key: {0}.")]
+    IncorrectSealKeySize(usize),
+    #[error("Signature generation failed: {0}")]
+    Signature(#[from] ecdsa::signature::Error),
+    #[error("Failed to decrypt shard: {0}")]
+    ShardDecryptionFailed(String),
+    #[error("Failed to encrypt shard: {0}")]
+    ShardEncryptionFailed(String),
+    #[error("Failed split key into shards: {0}")]
+    ShardingFailed(String),
+    #[error("Signature did not verify")]
+    VerificationFailed,
 }
 
 /// The associated data used in [`Encrypted`] AEAD ciphertexts and
--- a/lock-keeper/src/crypto/sharding.rs
+++ b/lock-keeper/src/crypto/sharding.rs
@@ -1,0 +1,297 @@
+use crate::crypto::{CryptoError, SigningPrivateKey, SigningPublicKey};
+use aes_gcm::{aead::Aead, AeadCore, Aes256Gcm, Key, KeyInit, Nonce};
+use k256::{elliptic_curve::PrimeField, NonZeroScalar, Scalar};
+use rand::rngs::OsRng;
+use serde::{Deserialize, Serialize};
+use std::ops::Deref;
+use vsss_rs::{combine_shares, shamir};
+use zeroize::Zeroize;
+
+// TODO: Is this comment true? I don't see why NUM_SHARDS must be a constant.
+// The number of shards we want to use must be a constant
+pub const NUM_SHARDS: usize = 3;
+pub const SHARD_THRESHOLD: usize = 3;
+/// Seal key is 32 bytes.
+pub const SEAL_KEY_LENGTH: usize = 32;
+
+/// TODO What exactly is a seal key?
+///
+/// We implement `Debug` and `Clone` because enclave codes relies on it.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct SealKey {
+    /// Our seal key is represented as a [`SEAL_KEY_LENGTH`] byte array.
+    material: [u8; SEAL_KEY_LENGTH],
+}
+
+impl SealKey {
+    pub const fn new(seal_key: [u8; SEAL_KEY_LENGTH]) -> Self {
+        Self { material: seal_key }
+    }
+
+    /// Convert key into a format that `[aes_gcm]` understands.
+    fn as_aes256gcm(&self) -> &Key<Aes256Gcm> {
+        (&self.material).into()
+    }
+
+    pub fn material(&self) -> &[u8] {
+        &self.material
+    }
+}
+
+impl TryFrom<Vec<u8>> for SealKey {
+    type Error = CryptoError;
+
+    fn try_from(material: Vec<u8>) -> Result<Self, Self::Error> {
+        let length = material.len();
+        let material = <[u8; SEAL_KEY_LENGTH]>::try_from(material)
+            .map_err(|_| CryptoError::IncorrectSealKeySize(length))?;
+        Ok(SealKey::new(material))
+    }
+}
+
+impl SigningPrivateKey {
+    /// Given a [`SigningPrivateKey`] shard and encrypt it using the `seal_key`.
+    ///
+    /// Note: `self` (`SigningPrivateKey`) will be zeroized as this value is
+    /// dropped.
+    pub fn shard_key_and_encrypt(
+        self,
+        seal_key: &SealKey,
+    ) -> Result<Vec<EncryptedShard>, CryptoError> {
+        let shards = shamir::split_secret::<Scalar, u8, Vec<u8>>(
+            SHARD_THRESHOLD,
+            NUM_SHARDS,
+            *self.as_nonzero_scalar().deref(),
+            &mut OsRng,
+        )
+        .map_err(|e| CryptoError::ShardingFailed(e.to_string()))?;
+
+        // Encrypt shards. (`shards` consumed by `UnencryptedShard` constructor, so
+        // no need to zeroize this value.)
+        shards
+            .into_iter()
+            .map(|shard| UnencryptedShard::new(shard).encrypt_shard(seal_key))
+            .collect()
+    }
+}
+
+/// Rebuild the [`SigningPrivateKey`] from its encrypted shards using the
+/// specified `seal_key`.
+///
+/// 1) Decrypt each shard with the given `seal_key`.
+/// 2) Rebuild key from decrypted shards.
+pub fn rebuild_key_from_encrypted_shards(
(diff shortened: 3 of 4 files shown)


lock-keeper/src/crypto/shard.rs
```diff
@@ -0,0 +1,180 @@
...
+            &mut OsRng,
+        )
+        .expect("TODO");
+
+        // Strip off identifiers from keys, we use their place in the array for knowing
+        // their position.
+        let shards: Vec<UnencryptedShard> = vsss_shard
+            .into_iter()
+            .map(|shard| UnencryptedShard::new(shard.value().to_vec()))
+            .collect();
```
gatoWololo, line 34: When you shard a value, the shards have an additional byte at the front representing the ID of the shard. This makes the size of the shards 16 (original shard size) + 1 (ID byte). So have to strip this value when chunking + encrypting, and later add this value back when rebuilding the key.

Which seems pretty ugly. Ideally, we would remove this code and work with 17 bytes shards? Thoughts?

lock-keeper/Cargo.toml
```diff
@@ -34,12 +34,17 @@ zeroize.workspace = true
 
 
 # Other dependencies
+
+# AES Used for encryption/decryption of sharded keys.
+aes = "0.8"
```
marsella, line 39: This [`aes`](https://crates.io/crates/aes) crate is not the correct choice for the dependency. Per the crate documentation:
> This crate implements the low-level AES block function, and is intended for use for implementing higher-level constructions only. It is NOT intended for direct use in applications. 

We should be using the [`aes-gcm`](https://docs.rs/aes-gcm/latest/aes_gcm/) crate instead, which depends on `aes` but provides a much better user interface and has been audited. It also supports `no_std` and can avoid allocation if that is necessary for our use case.

lock-keeper/src/crypto/shard.rs
marsella: Yeah, thanks for flagging this. When we switch to the aes-gcm library, we can circumvent the low-level break-your-message-into-chunks interface altogether and encrypt arbitrary-sized messages, so this should go away.

lock-keeper/src/crypto/shard/chunk.rs
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// impl From<EncryptedShard> for Vec<u8> {
+//     fn from(value: EncryptedShard) -> Self {
+//         value.0
+//     }
+// }
+
+/// Creates an `Aes256` cipher based on our sealing key.
+pub fn get_cipher(seal_key: &[u8]) -> Aes256 {
```
marsella, line 85: Perhaps you plan to do this in the future, but the seal key would ideally have its own type so it's easier to identify.

lock-keeper/src/crypto/shard/chunk.rs
```diff
@@ -0,0 +1,132 @@
...
+pub const CHUNK_SIZE: usize = 16;
+
+pub struct Chunk {
+    chunk: [u8; CHUNK_SIZE],
+}
+
+impl Chunk {
+    fn encrypt_chunk(&mut self, seal_key: &[u8]) {
+        let block = GenericArray::from_mut_slice(&mut self.chunk);
+        get_cipher(seal_key).encrypt_block(block);
```
marsella, line 19: Noting: this library encrypts in-place; you can use the corresponding [`encrypt_in_place` methods](https://docs.rs/aes-gcm/latest/aes_gcm/index.html#in-place-usage-eliminates-alloc-requirement) in the `aes-gcm` library if that's an important feature.

That might be affecting some of the zeroizing behavior that you noted was missing; it might not be possible to zeroize the original input if we use these in-place encryption methods. From a cryptographic perspective I would be fine to allocate memory and explicitly zeroize the original input.

marsella (changes requested):
Thanks for tagging me on this, Omar! The root of the complications you've pointed out here are because we are using the wrong AES library; the one included here provides a low-level interface for cryptographers. There's a better library (linked in comments) that provides a high-level interface for applications. We need to use that one instead.

gatoWololo:
@marsella This is now ready for a review!

lock-keeper/src/crypto/generic.rs
```diff
@@ -15,30 +15,42 @@ use zeroize::{Zeroize, ZeroizeOnDrop};
 /// Errors that arise in the cryptography module.
 #[derive(Debug, Error)]
 pub enum CryptoError {
```
gatoWololo, line 17: Moved around some of the variants to be in alphabetical order.

lock-keeper/src/crypto/sharding.rs
```diff
@@ -0,0 +1,297 @@
...
+use zeroize::Zeroize;
+
+// TODO: Is this comment true? I don't see why NUM_SHARDS must be a constant.
+// The number of shards we want to use must be a constant
+pub const NUM_SHARDS: usize = 3;
+pub const SHARD_THRESHOLD: usize = 3;
+/// Seal key is 32 bytes.
+pub const SEAL_KEY_LENGTH: usize = 32;
+
+/// TODO What exactly is a seal key?
```
gatoWololo, line 17: I should ask someone what a seal key actually is, so we can document it.

lock-keeper/src/crypto/sharding.rs
```diff
@@ -0,0 +1,297 @@
...
+    fn new(material: Vec<u8>) -> Self {
+        UnencryptedShard { material }
+    }
+
+    fn encrypt_shard(mut self, seal_key: &SealKey) -> Result<EncryptedShard, CryptoError> {
+        let nonce: Nonce<_> = Aes256Gcm::generate_nonce(&mut OsRng);
+        let cipher = Aes256Gcm::new(seal_key.as_aes256gcm());
+
+        // Encrypt shard, zeroize unecrypted shard regardless of the results of encrypt.
+        let result = cipher.encrypt(&nonce, self.material.as_slice());
```
gatoWololo, line 139: I ended up NOT encrypting in place, I think this makes it easier to reason about copies and I can explicitly zeroize.

lock-keeper/src/crypto/sharding.rs
```diff
@@ -0,0 +1,297 @@
...
+    pub fn take_shards(self) -> Vec<EncryptedShard> {
+        self.encrypted_shards
+    }
+
+    pub fn into_parts(self) -> (SigningPublicKey, Vec<EncryptedShard>) {
+        (self.public_key, self.encrypted_shards)
+    }
+}
+
+// TODO: Should encrypted shards or nonce be zeroized after use?
```
gatoWololo, line 180: This is a question for you :)

lock-keeper/src/crypto/sharding.rs
```diff
@@ -0,0 +1,297 @@
...
+            .collect()
+    }
+}
+
+/// Rebuild the [`SigningPrivateKey`] from its encrypted shards using the
+/// specified `seal_key`.
+///
+/// 1) Decrypt each shard with the given `seal_key`.
+/// 2) Rebuild key from decrypted shards.
+pub fn rebuild_key_from_encrypted_shards(
```
marsella, line 83: question: Why is this and `rebuild_key_from_shards` not methods on `SigningPrivateKey`?

lock-keeper/src/crypto/sharding.rs
```diff
@@ -0,0 +1,297 @@
...
+    /// Our seal key is represented as a [`SEAL_KEY_LENGTH`] byte array.
+    material: [u8; SEAL_KEY_LENGTH],
+}
+
+impl SealKey {
+    pub const fn new(seal_key: [u8; SEAL_KEY_LENGTH]) -> Self {
+        Self { material: seal_key }
+    }
+
+    /// Convert key into a format that `[aes_gcm]` understands.
```
marsella, line 31: todo: Make this actually be a link to the `aes_gcm` library, or the output type.

lock-keeper/src/crypto/sharding.rs
```diff
@@ -0,0 +1,297 @@
...
+    }
+}
+
+impl TryFrom<Vec<u8>> for SealKey {
+    type Error = CryptoError;
+
+    fn try_from(material: Vec<u8>) -> Result<Self, Self::Error> {
+        let length = material.len();
+        let material = <[u8; SEAL_KEY_LENGTH]>::try_from(material)
+            .map_err(|_| CryptoError::IncorrectSealKeySize(length))?;
```
marsella, line 47: question: Will this fail in both the too-long and too-short cases, or only too-short?

lock-keeper/src/crypto/sharding.rs
```diff
@@ -0,0 +1,297 @@
...
+            public_key,
+            encrypted_shards,
+        }
+    }
+
+    pub fn get_public_key(&self) -> &SigningPublicKey {
+        &self.public_key
+    }
+
+    pub fn get_shards(&self) -> &[EncryptedShard] {
```
marsella, line 167: todo: I would recommend removing `get_` from these names to align with [Rust naming conventions](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter). Is that an annoying API change?

lock-keeper/src/crypto/sharding.rs
```diff
@@ -0,0 +1,297 @@
...
+        &self.encrypted_shards
+    }
+
+    pub fn take_shards(self) -> Vec<EncryptedShard> {
```
marsella, line 171: nit: Similarly, I would call this `into_shards`

lock-keeper/src/crypto/sharding.rs
```diff
@@ -0,0 +1,297 @@
...
+    fn random_shard() -> UnencryptedShard {
+        let rng = &mut OsRng;
+        // Create buffer to fill (length chosen arbitrarily).
+        let mut material: [u8; 128] = [0; 128];
+        rng.fill(&mut material);
+        UnencryptedShard::new(material.to_vec())
+    }
+
+    #[test]
+    fn encrypt_shard_works() {
```
marsella, line 229: thought: This is a kind of nebulous thing to test if we're not also decrypting it. You might check, say, that the encrypted shard is different from the unencrypted one, but the property we really want to test ("an attacker cannot determine any information about the plaintext from the ciphertext") is not exactly falsifiable.

marsella (changes requested):
Thanks, this looks pretty good! I have some cleanup notes and other thoughts throughout.

lock-keeper/src/crypto/sharding.rs
```diff
@@ -0,0 +1,297 @@
...
+use aes_gcm::{aead::Aead, AeadCore, Aes256Gcm, Key, KeyInit, Nonce};
+use k256::{elliptic_curve::PrimeField, NonZeroScalar, Scalar};
+use rand::rngs::OsRng;
+use serde::{Deserialize, Serialize};
+use std::ops::Deref;
+use vsss_rs::{combine_shares, shamir};
+use zeroize::Zeroize;
+
+// TODO: Is this comment true? I don't see why NUM_SHARDS must be a constant.
+// The number of shards we want to use must be a constant
```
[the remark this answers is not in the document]
gatoWololo: Oh neat!

lock-keeper/src/crypto/sharding.rs
```diff
@@ -0,0 +1,297 @@
...
+pub const NUM_SHARDS: usize = 3;
+pub const SHARD_THRESHOLD: usize = 3;
+/// Seal key is 32 bytes.
+pub const SEAL_KEY_LENGTH: usize = 32;
+
+/// TODO What exactly is a seal key?
+///
+/// We implement `Debug` and `Clone` because enclave codes relies on it.
+#[derive(Debug, Clone, Eq, PartialEq)]
```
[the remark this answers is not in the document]
gatoWololo: That's a great idea. I can implement by own Debug instances.

lock-keeper/src/crypto/sharding.rs
```diff
@@ -0,0 +1,297 @@
...
+pub struct SealKey {
+    /// Our seal key is represented as a [`SEAL_KEY_LENGTH`] byte array.
+    material: [u8; SEAL_KEY_LENGTH],
```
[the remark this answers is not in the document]
gatoWololo: I tried adding the Box but it causes issues since the new constructor is a const function right now:
```rust
pub const fn new(seal_key: [u8; SEAL_KEY_LENGTH]) -> Self
```
This makes it possible to use the constructor to create const `SealKey` values for testing in this code and in the enclave:
```rust
const TESTING_SEAL_KEY: SealKey = SealKey::new([
        249, 95, 53, 174, 82, 104, 71, 122, 147, 115, 101, 210, 113, 233, 22, 31, 46, 191, 19, 167,
        45, 199, 77, 215, 193, 41, 153, 197, 10, 193, 255, 145,
    ]);
```
I can't do that with the Box. I did add a `ZeroizeOnDrop` to this type! I don't think this is an issue memory management-wise. This type is slightly large at 32-bytes but I think that's still reasonable.

lock-keeper/src/crypto/sharding.rs
```diff
@@ -0,0 +1,297 @@
...
+    let shards: Vec<Vec<u8>> = shards.into_iter().map(|shard| shard.material).collect();
+    let results = combine_shares(&shards);
+
+    // We no longer need the unencrypted shards. Zeroize them.
+    shards.into_iter().for_each(|mut shard| shard.zeroize());
+
+    let scalar: Scalar = results.map_err(|e| CryptoError::CombineShardsFailed(e.to_string()))?;
+    let non_zero_scalar = NonZeroScalar::from_repr(scalar.to_repr());
+
+    if bool::from(non_zero_scalar.is_none()) {
```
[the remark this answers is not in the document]
gatoWololo: Thanks! We weren't sure how to do this :dizzy_face:

lock-keeper/src/crypto/sharding.rs
```diff
@@ -0,0 +1,297 @@
...
+    #[test]
+    fn shard_roundtrip_works() {
+        let shard = random_shard();
+        // Make clone shard for comparison.
+        let shard_copy = shard.clone();
+
+        let encrypted = shard.encrypt_shard(&TESTING_SEAL_KEY).unwrap();
+        let decrypted = encrypted.decrypt_shard(&TESTING_SEAL_KEY).unwrap();
+
+        // Note: Cannot use `assert_eq!` as `UnencryptedShard` does not impl Debug.
```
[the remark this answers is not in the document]
gatoWololo: Great idea!

lock-keeper/src/crypto/sharding.rs
gatoWololo: Good point.

lock-keeper/src/crypto/sharding.rs
gatoWololo: It will fail both if the vec is both too short or too long.

lock-keeper/src/crypto/sharding.rs
gatoWololo: That makes sense. _deleted_

gatoWololo:
@marsella This is ready for re-review! I marked all the conversations as "Resolved" for things I fixed/changed. I commented on the few that remain.

marsella (approved):
Looks good, thanks!

PR #537, opened 2023-07-20
83 opaque password encryption

### LINKED ISSUE #530: OPAQUE password encryption for key servers

tposavec:
Closes #530 

The branch for this PR should have been named 530-opaque-password-encryption.

NOTE: For reference, there is an older PR https://github.com/boltlabs-inc/key-mgmt/pull/531 that was closed because the code had to be refactored.


lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+};
+
+use chacha20poly1305::{
+    aead::{Aead, Payload},
+    AeadCore, ChaCha20Poly1305, KeyInit,
+};
+
+/// The [`Password`] type.
+/// A password contains password `material` and `context` data.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, ZeroizeOnDrop)]
```
gatoWololo, line 21: I think we talked abount not implementing `Debug` for password to avoid it accidentally being printed.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+        Ok(bytes)
+    }
+}
+
+/// Implementation of the `TryFrom` trait for converting from `Vec<u8>` to
+/// `Password`
+impl TryFrom<Vec<u8>> for Password {
+    type Error = CryptoError;
+
+    #[instrument(skip_all)]
```
gatoWololo, line 91: Ooh, I didn't realize we could use the instrument macro inside trait impl blocks like this!

gatoWololo: I think you wanna add `err(Debug)` like this: `#[instrument(skip_all, err(Debug))]` for this. This makes the error logging a little nicer.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+        let context: Vec<u8> = parse.take_rest()?.to_vec();
+
+        // make sure that the length of the context matches the parsed length
+        if context.len() != context_length as usize {
+            return Err(CryptoError::ConversionError);
+        }
+
+        // create and return a new Password instance from the parsed password and
+        // context
+        Ok(Self {
```
gatoWololo, line 113: Personally, I like avoiding unnecessary comments like this one. This comment is mostly stating the obvious. Our code mostly follows this style. I do appreciate all the other comments throughout!

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+            material: password_material,
+            context: context.try_into()?,
+        })
+    }
+}
+
+impl Password {
+    /// Encrypt a `Password` using the provided encryption key.
```
gatoWololo, line 121: When referring to types that are in scope in doc strings. You can use the syntax [`Password`] to have Rust create a hyperlink in the generated docs. I think you should switch this one here. As well as  `EncryptedPassword` and `CryptoError` below

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+        let payload = Payload {
+            msg: &self.material,
+            aad: &context_bytes,
+        };
+
+        // Generate nonce and encrypt the payload
+        let nonce = ChaCha20Poly1305::generate_nonce(rng);
+        let encrypted_password = cipher
+            .encrypt(&nonce, payload)
+            .map_err(|_| CryptoError::EncryptionFailed)?;
```
gatoWololo, line 154: In this `map_err` you probably wanna print the error via our error log:
```rust
 .map_err(|e| {
     error!("{:?}", e);
     CryptoError::EncryptionFailed
     })?;
```
Otherwise we know that encryption failed, but we don't know why it failed. This will make it easier to debug if it ever happens.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+    #[allow(dead_code)]
+    fn context(&self) -> &PasswordContext {
+        &self.context
+    }
+}
+
+/// The context (a.k.a. associated data) used in [`Password`]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+struct PasswordContext {
+    key_server_name: Vec<u8>,
```
gatoWololo, line 173: How come key server name is a `Vec<u8>` instead of a String? I see below that it seems to be in hex format?

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+        PasswordContext {
+            // use as_ref method to get a byte slice from the parameter,
+            // then convert the byte slice to a Vec<u8> with the to_vec method
+            key_server_name: key_server_name.as_ref().to_vec(),
+        }
+    }
+}
+
+/// Conversion from `Vec<u8>` to PasswordContext
+impl TryFrom<Vec<u8>> for PasswordContext {
```
gatoWololo, line 226: This conversion seems to be infallible. You can just use the `From` trait in that case. This simplifies the code since we don't have to handle a non-existent error path.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+/// Conversion from PasswordContext to `Vec<u8>`
+impl From<PasswordContext> for Vec<u8> {
+    fn from(password_context: PasswordContext) -> Self {
+        password_context.key_server_name
+    }
+}
+
+/// The [`EncryptedPassword`] type.
+/// An encrypted password contains `ciphertext`, `context` and `nonce`.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
```
gatoWololo, line 246: Should this implement `Debug`?

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+pub struct EncryptedPassword {
+    ciphertext: Vec<u8>,
+    context: PasswordContext,
+    nonce: chacha20poly1305::Nonce,
+}
+
+/// An implementation of the `std::fmt::Display` trait for `EncryptedPassword`
+impl std::fmt::Display for EncryptedPassword {
```
gatoWololo, line 254: I'm curious what the use case for the `Display` implementations are? We don't really use them in any of our repos and I haven't found it necessary.

lock-keeper/src/crypto/password.rs
gatoWololo: If there isn't really a use case for using `Vec<u8>` over string. I would suggest using strings as they seem a better fit for this use case.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+    fn password_vec_u8_conversion() -> Result<(), CryptoError> {
+        let mut rng = rand::thread_rng();
+
+        // generate a random 32-byte password
+        let password_plaintext: [u8; 32] = rng.gen();
+
+        // create a Password instance
+        let password = Password {
+            material: password_plaintext.to_vec(),
+            context: PasswordContext {
```
gatoWololo, line 346: You can use your constructor here to simplify:
```rust
context: PasswordContext::new(KEY_SERVER_NAME),
```

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+
+        // generate a random 32-byte password
+        let password_material: [u8; 32] = rng.gen();
+
+        // generate a password encryption key
+        let password_encryption_key = PasswordKey::generate(&mut rng);
+
+        let plaintext_password = Password {
+            material: password_material.to_vec(),
+            context: PasswordContext {
```
gatoWololo, line 397: Same as above, simplify `PasswordContext` creation here.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+        let password = Password {
+            material: password_plaintext.to_vec(),
+            context: PasswordContext {
+                key_server_name: KEY_SERVER_NAME.as_bytes().to_vec(),
+            },
+        };
+
+        // convert Password to Vec<u8>
+        let password_vec: Vec<u8> = password.clone().try_into()?;
+
```
gatoWololo, line 353: I don't think we have solid guidelines about whitespace. But personally, I think we could get rid of some of these newlines. It makes the code a little more compact. I like new lines between logically distinct lines of code. But here they can be compressed a bit.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+    /// encryption process.
+    ///
+    /// # Errors
+    ///
+    /// * Returns a `CryptoError` if there is an error during encryption,
+    ///   decryption, or verification.
+    #[test]
+    fn password_encrypt_decrypt() -> Result<(), CryptoError> {
+        let mut rng = rand::thread_rng();
+
```
gatoWololo, line 388: Similar comment about these new lines here.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+            \nPlaintext password length: {password_len}\
+            \n\nDecrypted password:\n{decrypted_password}",
+            password = password_material,
+            password_len = password_material.len(),
+            decrypted_password = decrypted_password,
+        );
+
+        // make sure the decrypted password matches the original password
+        assert_eq!(original_password, decrypted_password);
+
```
gatoWololo, line 436: Last example I will point out about newlines I consider unnecessary.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+        // generate a random 32-byte password
+        let password_material: [u8; 32] = rng.gen();
+
+        // generate a password encryption key
+        let password_encryption_key = PasswordKey::generate(&mut rng);
+
+        // create a new instance of Password
+        let plaintext_password = Password {
+            material: password_material.to_vec(),
+            context: PasswordContext {
```
gatoWololo, line 460: Simplify `PasswordContext` construction.

gatoWololo, line 460: lol

gatoWololo (changes requested):
Overall the code looks good! Left some comments about style and small changes. I left more comments than I usually would about style since you are still getting used to how we write code :)

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+        println!(
+            "\nEncrypted password:{encrypted_password}",
+            encrypted_password = encrypted_password
+        );
+
+        // modify the context aka associated data
+        encrypted_password
+            .context
+            .key_server_name
+            .extend_from_slice("__baaaaaad__".as_bytes());
```
gatoWololo, line 478: Lol

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+        Ok(EncryptedPassword {
+            ciphertext: encrypted_password,
+            context: self.context.clone(),
+            nonce,
+        })
+    }
+
+    /// Retrieve the context for this password.
+    #[allow(dead_code)]
+    fn context(&self) -> &PasswordContext {
```
marsella, line 165: Do you expect to use this method in this crate in a later refactor? If not, I'd recommend removing it. If it's supposed to be publicly accessible, update the publicity and remove the `#[allow(...)]`.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+    PasswordKey,
+};
+
+use chacha20poly1305::{
+    aead::{Aead, Payload},
+    AeadCore, ChaCha20Poly1305, KeyInit,
+};
+
+/// The [`Password`] type.
+/// A password contains password `material` and `context` data.
```
marsella, line 20: suggestion: I think this would benefit from more specific context about how you expect the password type to be used and what guarantees this type provides.

In particular, this isn't a normal user facing password in our application, and this type isn't really appropriate for that use case: it doesn't check any measure of "strength" or security of the password. Will send you some private notes about our actual application.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+        // create and return a new Password instance from the parsed password and
+        // context
+        Ok(Self {
+            material: password_material,
+            context: context.try_into()?,
+        })
+    }
+}
+
+impl Password {
```
marsella, line 120: todo: I don't see any constructor other than the `TryFrom<Vec<u8>>`, which is only appropriate for serialized passwords.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+    ///     let context = PasswordContext::new("my_key_server");
+    /// ```
+    ///
+    /// From `Vec<u8>`...
+    /// ```ignore
+    ///     let key_server_name = "my_key_server".as_bytes().to_vec();
+    ///     let context = PasswordContext::new(key_server_name);
+    /// ```
+    #[allow(dead_code)]
+    fn new<T: AsRef<[u8]>>(key_server_name: T) -> Self {
```
marsella, line 216: todo: This function seems like it's supposed to be public. I would recommend changing the publicity and removing the allow if so.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+        let cipher = ChaCha20Poly1305::new(&password_key.key_material);
+
+        // Convert PasswordContext to a Vec<u8>
+        let context_bytes: Vec<u8> = self.context.clone().into();
+
+        // Format ciphertext and associated data
+        let payload = Payload {
+            msg: &self.ciphertext,
+            aad: &context_bytes,
+        };
```
marsella, line 306: todo: Same note as above re: removing clone for the context.

lock-keeper/src/crypto/password_key.rs
```diff
@@ -0,0 +1,292 @@
...
+use zeroize::{Zeroize, ZeroizeOnDrop};
+
+/// The [`PasswordKey`] type is a default-length symmetric encryption key
+/// for an AEAD scheme. It can be used to securely encrypt passwords.
+///
+/// Note: we don't implement the Copy trait.
+/// This is because implementing Copy could potentially lead to multiple copies
+/// of the key in memory, which increases the chances of the key being leaked or
+/// exposed.
+#[derive(Debug, Clone, PartialEq, Eq, Zeroize, ZeroizeOnDrop)]
```
marsella, line 19: issue: As above, I'm inclined to prefer a redacted `Debug` implementation and no `Display` implementation because I don't want this secret getting left around.

lock-keeper/src/crypto/password_key.rs
```diff
@@ -0,0 +1,292 @@
...
+            key_len = self.key_material.len(),
+        )
+    }
+}
+
+impl PasswordKey {
+    /// Length of the password encryption key in `bytes`
+    const PASSWORD_KEY_LENGTH: usize = 32;
+
+    /// Generate a new symmetric AEAD encryption key from scratch.
```
marsella, line 45: todo: Instead of "from scratch" I would say:
```rust
/// Generate a new `PasswordKey` uniformly at random.
///
/// A `PasswordKey` is a 32-byte pseudorandom key for use in the 
/// [ChaCha20Poly1305 scheme](https://www.rfc-editor.org/rfc/rfc8439) for
/// authenticated encryption with associated data (AEAD).

File: lock-keeper/src/crypto/password.rs [hunk 24]
```diff
@@ -0,0 +1,496 @@
...
+
+/// Conversion from PasswordContext to `Vec<u8>`
+impl From<PasswordContext> for Vec<u8> {
+    fn from(password_context: PasswordContext) -> Self {
+        password_context.key_server_name
+    }
+}
+
+/// The [`EncryptedPassword`] type.
+/// An encrypted password contains `ciphertext`, `context` and `nonce`.
```
marsella, line 245: todo: would recommend docs like

```rust

/// A ciphertext representing a `Password`, encrypted under the
/// [ChaCha20Poly1305 scheme](https://www.rfc-editor.org/rfc/rfc8439) for
/// authenticated encryption with associated data (AEAD).
///
/// As implied by the scheme name, this uses the recommended 20 rounds and a
/// standard 96-bit nonce. For more details, see the
/// [ChaCha20Poly1305 crate](https://docs.rs/chacha20poly1305/latest/chacha20poly1305/index.html).

File: lock-keeper/src/crypto/password_key.rs [hunk 25]
```diff
@@ -0,0 +1,292 @@
...
+    pub fn generate(rng: &mut (impl CryptoRng + RngCore)) -> Self {
+        Self::new(rng)
+    }
+
+    /// Returns the [`PasswordKey`] stored in a file.
+    pub fn read_from_file(path: impl AsRef<Path>) -> Result<Self, CryptoError> {
+        let bytes = std::fs::read(&path)
+            .map_err(|e| CryptoError::FileIo(e, path.as_ref().to_path_buf()))?;
+
+        Self::from_bytes(&bytes)
```
marsella, line 62: todo: Zeroize `bytes` before returning from this method.

lock-keeper/src/crypto/password_key.rs
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    /// Returns the [`PasswordKey`] key based on key material in the caller's
+    /// byte array.
+    fn from_bytes(key_material: &[u8]) -> Result<Self, CryptoError> {
```
marsella, line 67: thought: I wish chacha20 keys had a more standardized format than just being a random byte string :(

lock-keeper/src/crypto/password_key.rs
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    /// WARNING: this generates a copy of the key,
+    /// and should only be used to derive another key from this key using a KDF.
+    /// This should explicitly stay pub(super) to avoid abuse.
+    #[allow(dead_code)]
+    pub(super) fn into_bytes(self) -> [u8; PasswordKey::PASSWORD_KEY_LENGTH] {
+        (*self.key_material).into()
+    }
+}
```
marsella, line 87: issue: I would rather delete this function and add key derivation to this module when we need it. Is that a functionality currently happening elsewhere in our code base?

lock-keeper/src/crypto/password_key.rs
```diff
@@ -0,0 +1,292 @@
...
+
+/// Implementation of the `TryFrom` trait for converting from `PasswordKey` to
+/// `Vec<u8>`
+impl TryFrom<PasswordKey> for Vec<u8> {
+    type Error = CryptoError;
+
+    fn try_from(key: PasswordKey) -> Result<Self, Self::Error> {
+        // len || key || context
```
marsella, line 95: nit: comment is out of date; there's no context for a `PasswordKey`.

lock-keeper/src/crypto/password_key.rs
```diff
@@ -0,0 +1,292 @@
...
+
+/// Implementation of the `TryFrom` trait for converting from `Vec<u8>` to
+/// `PasswordKey`
+impl TryFrom<Vec<u8>> for PasswordKey {
+    type Error = CryptoError;
+
+    fn try_from(bytes: Vec<u8>) -> Result<Self, Self::Error> {
+        // len (should always be 32) || key (32 bytes) || context (remainder)
+        let len = *bytes.first().ok_or(CryptoError::ConversionError)? as usize;
+        if len != 32 {
```
marsella, line 112: nit: use your constant here instead of the raw number 32

lock-keeper/src/crypto/password_key.rs
```diff
@@ -0,0 +1,292 @@
...
+        }
+
+        let key_offset = len + 1;
+        let key = bytes
+            .get(1..key_offset)
+            .ok_or(CryptoError::ConversionError)?;
+        let _context: Vec<u8> = bytes
+            .get(key_offset..)
+            .ok_or(CryptoError::ConversionError)?
+            .into();
```
marsella, line 123: question: Can you use parsebytes here to avoid the offset code? I hated writing that originally.

lock-keeper/src/crypto/password_key.rs
```diff
@@ -0,0 +1,292 @@
...
+
+        // compare the original key and the newly created key from the original key
+        // bytes
+        assert_eq!(password_encryption_key_clone, new_password_encryption_key);
+
+        Ok(())
+    }
+
+    #[test]
+    fn password_key_from_bytes_wrong_length_fails() -> Result<(), CryptoError> {
```
marsella, line 202: todo: I think this test should try both the too-long and too-short cases.

marsella (changes requested):
Thanks for the updated PR! Looks much better already. Apologies in advance -- I think I have some overlapping comments with Omar.

Can you explain the `pub(super)`s her? I think some of these types should just be straight-up public; in the previous use case, this module was only used by types within the library, but now it's going to be used by external libraries and the primary types (`Password`, `PasswordKey`) likely need to be public. You might have to fight a little with the existing module publicity to make sure those are exposed.

In general for testing, I don't include explicit `println` s unless I'm actively debugging something. Once the test works, if I think there needs to be additional info, I either add context to the asserts to give a hint as to what went wrong in a failure case or use the `tracing` logging infrastructure. I don't think this crate is actually set up with tracing though. This is a stylistic preference and I'll defer to @gatowololo if he disagrees or doesn't care.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+    fn try_from(bytes: Vec<u8>) -> Result<Self, Self::Error> {
+        // Input byte array format:
+        // password len (2 bytes) || password material || context len (2 bytes) ||
+        // context data
+
+        // we'll be parsing the byte array, and creating an instance of Password type
+        let mut parse = ParseBytes::new(bytes);
+
+        let password_length = parse.take_bytes_as_u16()?;
+        let password_material = parse.take_bytes(password_length as usize)?.to_vec();
```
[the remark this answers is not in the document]
gatoWololo: Password length is being "extended" from a `u16` to a `usize` this conversion is always valid and guaranteed to never truncate data (unlike converting from usize -> u16 which can truncate data :grimacing: , I consider this a flaw in Rust). But we should add a comment here to be explicit.

lock-keeper/src/crypto/password.rs
```diff
@@ -0,0 +1,496 @@
...
+
+        // Generate nonce and encrypt the payload
+        let nonce = ChaCha20Poly1305::generate_nonce(rng);
+        let encrypted_password = cipher
+            .encrypt(&nonce, payload)
+            .map_err(|_| CryptoError::EncryptionFailed)?;
+
+        Ok(EncryptedPassword {
+            ciphertext: encrypted_password,
+            context: self.context.clone(),
```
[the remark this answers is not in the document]
gatoWololo: In this case you do actually need the clone:
```rust
error[E0509]: cannot move out of type `password::Password`, which implements the `Drop` trait
   --> lock-keeper/src/crypto/password.rs:158:22
    |
158 |             context: self.context,
    |                      ^^^^^^^^^^^^
    |                      |
    |                      cannot move out of here
    |                      move occurs because `self.context` has type `PasswordContext`, which does not implement the `Copy` trait

```
This is because `Password` implements `ZeroizeOnDrop` so the compiler won't let you move a field out of the struct. Even though this field is never used in the `Drop` impl, the compiler is unable to reason about this.

This is an issue with using `ZeroizeOnDrop`. One workaround if we really wanted to avoid the clone is to create a `InnerPassword` type which just wraps the `Vec<u8>` this way the entire `Password` struct does not need to impl drop. This the reason I propose the approach described in https://github.com/boltlabs-inc/tss-ecdsa/issues/330

gatoWololo:
> I don't think this crate is actually set up with tracing though. This is a stylistic preference and I'll defer to @gatoWololo if he disagrees or doesn't care.

I agree with your stylistic preference and we also don't usually use `println!`. We rely on asserts which can take [an optional message](https://doc.rust-lang.org/std/macro.assert.html#custom-messages)!

The entire crate does have support for tracing, but I don't think we use it in the crypto module. The problem with using `tracing` in tests is that you have to init the logger per thread, which is annoying! So I would not advise to do that.

lock-keeper/src/crypto/password.rs
marsella: I am less concerned about logging encrypted blobs, although it might be best to manually redact the nonce.

tposavec:
I did a major refactor of this PR. I'm closing it, and creating a new PR.

PR #540, opened 2023-07-28
83 opaque password encryption

### LINKED ISSUE #530: OPAQUE password encryption for key servers

tposavec:
Closes #530 

The branch for this PR should have been named 530-opaque-password-encryption


lock-keeper/src/infrastructure/sensitive_info.rs
```diff
@@ -0,0 +1,177 @@
...
+
+use serde::{Deserialize, Serialize};
+use zeroize::{Zeroize, ZeroizeOnDrop};
+
+/// The [`SensitiveInfoConfig`] type is used to configure how we handle
+/// sensitive information such as cryptographic keys, passwords, etc.
+#[derive(Clone, PartialEq, Eq, Zeroize, ZeroizeOnDrop, Serialize, Deserialize)]
+pub struct SensitiveInfoConfig {
+    /// A boolean indicating whether sensitive information should be redacted.
+    ///     If set to true, the sensitive information will be redacted (hidden
```
gatoWololo, line 16: Nit: Weird whitespace on second comment line

lock-keeper/src/infrastructure/sensitive_info.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// or masked).     If set to false, the sensitive information will not
+    /// be redacted.
+    redact_sensitive_info: bool,
+}
+
+/// Conversion from `Vec<u8>` to [`SensitiveInfoConfig`]
+impl From<Vec<u8>> for SensitiveInfoConfig {
+    fn from(bytes: Vec<u8>) -> Self {
+        let config: SensitiveInfoConfig = bincode::deserialize(&bytes).unwrap();
```
gatoWololo, line 25: I'm concerned about this unwrap. We usually have `.unwrap()` only for testing code and not anything that could be used in production.
1) You can use `TryFrom` instead to return an error from this function.

lock-keeper/src/infrastructure/sensitive_info.rs
```diff
@@ -0,0 +1,177 @@
...
+        config
+    }
+}
+
+/// Conversion from [`SensitiveInfoConfig`] to `Vec<u8>`
+impl From<SensitiveInfoConfig> for Vec<u8> {
+    fn from(config: SensitiveInfoConfig) -> Self {
+        let encoded: Vec<u8> = bincode::serialize(&config).unwrap();
```
gatoWololo, line 33: Same concern with unwrap here.

lock-keeper/src/crypto/cryptor.rs
```diff
@@ -0,0 +1,746 @@
...
+}
+
+/// The [`Decryptor`] type represents a ciphertext encrypted under the
+/// [ChaCha20Poly1305 scheme](https://www.rfc-editor.org/rfc/rfc8439) for
+/// authenticated encryption with associated data (AEAD).
+///
+/// As implied by the scheme name, this uses the recommended 20 rounds and a
+/// standard 96-bit nonce. For more details, see the
+/// [ChaCha20Poly1305 crate](https://docs.rs/chacha20poly1305/latest/chacha20poly1305/index.html).
+#[derive(Clone, Eq)]
```
gatoWololo, line 359: Question: Is implementing equality meaningful for Decryptor? Won't the nonce be different in every instantiation due to it being randomly chosen? What is the use case for `Decryptor` comparison?

lock-keeper/src/crypto/cryptor.rs
```diff
@@ -0,0 +1,746 @@
...
+        println!(
+            "\nOriginal raw plaintext data: 0x{data:02x?}",
+            data = data_plaintext
+        );
+        println!(
+            "Original raw plaintext data length: {data_len}",
+            data_len = data_plaintext.len()
+        );
+
+        println!("\nOriginal encryptor: {encryptor}", encryptor = encryptor);
```
gatoWololo, line 595: Any reason you prefer this syntax over `println!("\nOriginal encryptor: {encryptor}");` or `println!("\nOriginal encryptor: {}", encryptor);`?

lock-keeper/src/crypto/cryptor_key.rs
```diff
@@ -0,0 +1,316 @@
...
+use rand::{CryptoRng, RngCore};
+use std::path::Path;
+use zeroize::{Zeroize, ZeroizeOnDrop};
+
+use crate::infrastructure::sensitive_info::SensitiveInfoConfig;
+
+/// The [`CryptorKey`] type is a default-length symmetric encryption key
+/// for an AEAD scheme. It can be used to securely encrypt data.
+///
+/// Note: we don't implement the Copy trait.
```
gatoWololo, line 17: So Rust actually won't let us implement Copy on this type. Rust makes it so only types that "look like bytes" can be copied. The `Box` here implies memory allocation, so no copy allowed.

lock-keeper/src/crypto/cryptor_key.rs
```diff
@@ -0,0 +1,316 @@
...
+        // encountered an error.
+        match fs::remove_file(CRYPTOR_KEY_FILENAME) {
+            Ok(()) => (),
+            Err(e) => println!("Failed to remove a temp key file: {}", e),
+        }
+
+        // Now handle the result of key_from_file_helper()
+        match result {
+            Ok(()) => (),
+            Err(e) => println!("Test failed: {}", e),
```
gatoWololo, line 271: You probably want to panic! on this Err(_) branch. Currently, it just prints "Test failed: ..." but the test still passes from `cargo test`'s POV.

lock-keeper/src/crypto/cryptor_key.rs
```diff
@@ -0,0 +1,316 @@
...
+        }
+
+        Ok(())
+    }
+
+    /// Helper function, so we can handle read from file failures correctly.
+    fn key_from_file_helper() -> Result<(), CryptoError> {
```
gatoWololo, line 278: I see you are mixing `?` and expects in this test function. That makes sense as you are returning a `CryptoError`. This is a perfect use case for the `anyhow` crate. We use it to return all errors on tests. We can talk about how to use this at some point!

gatoWololo: But short version. Change fn signature to: `fn key_from_file_helper() -> Result<(), anyhow::Error> {` and change all uses of `.except(...)` to `?`. It just works :)

gatoWololo (commented):
Overall this looks good. I have minor comments.

lock-keeper/src/infrastructure/sensitive_info.rs
tposavec: @gatoWololo fixed here https://github.com/boltlabs-inc/key-mgmt/commit/0b512026b3145a09a3e3f1332599fdfc94848f04

lock-keeper/src/crypto/cryptor.rs
tposavec: @gatoWololo I think the equality operation on the `Decryptor` is useful.  `Nonce` will/should be different, as you point out.  However, here are some use cases for the `PartialEq` trait:  
- testing & debugging (I use this in my unit tests): this helps verify that the encryption and decryption processes are both functioning correctly.  Also, when testing conversion functionality to/from vec<u8>...
- state verification: it might be necessary to verify that a Decryptor has not changed during its lifecycle, particularly in a multi-threaded/distributed system. This could be done by periodically comparing a Decryptor to a previously stored version.
- caching: If Decryptor objects are cached, for wahtever reason, we could use the PartialEq implementation to identify whether a suitable Decryptor is already available in the cache.

lock-keeper/src/crypto/cryptor_key.rs
tposavec: @gatoWololo changed here https://github.com/boltlabs-inc/key-mgmt/commit/f3e391ea6f154dc34f629d1c9d12f595a97dff27

lock-keeper/src/crypto/cryptor_key.rs
tposavec: @gatoWololo fixed here https://github.com/boltlabs-inc/key-mgmt/commit/85ab86acd7502d29ac9e20973ca416735c265c67

lock-keeper/src/crypto/cryptor_key.rs
tposavec: @gatoWololo fixed here https://github.com/boltlabs-inc/key-mgmt/commit/6f4a7bab573546808e75fb567ad7d40db963cdb2

lock-keeper/src/infrastructure/sensitive_info.rs
tposavec: @gatoWololo fixed here https://github.com/boltlabs-inc/key-mgmt/commit/af136ea6848fc2a26e14b87353b7604040453128

lock-keeper/src/infrastructure/sensitive_info.rs
tposavec: @gatoWololo fixed here https://github.com/boltlabs-inc/key-mgmt/commit/af136ea6848fc2a26e14b87353b7604040453128

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.65
+FROM rust:1.67
```
marsella, line 1: question: Why leave the Docker container at an older version than the library requires?

marsella (changes requested):
I have a few changes requested, but overall this looks much better.

Process note: I'll look at the diff to see whether you addressed my changes, so you don't need to link the commit in which you made the change to each of my comments.

lock-keeper/src/crypto/cryptor.rs
gatoWololo: That makes sense. Thanks for the explanation!

lock-keeper/src/crypto/cryptor.rs
gatoWololo: Fair enough.

gatoWololo (approved):
Looks good Tom. I'll let Marcella handle the security and crypto :)

Please revert the Rust version changes! After a rebase, it should work again.

Dockerfile
tposavec: @marsella those were temp changes, so I could get the CI jobs to build cleanly.  Omar's changes with the .lock file solved the issue, so I reverted my temp changes here https://github.com/boltlabs-inc/key-mgmt/pull/540/commits/0354071c50bc5a2af85a9e5ee57af98d7c63e5e9

lock-keeper/src/crypto/cryptor.rs
```diff
@@ -0,0 +1,746 @@
...
+            write!(f, "Decryptor {}", self.config.clone().redacted_label())
+        }
+    }
+}
+
+/// Implement the `PartialEq` trait for the [`Decryptor`] type.
+impl PartialEq for Decryptor {
+    /// Determines if two [`Decryptor`] instances are equal.
+    ///
+    /// This function compares the `data`, `ciphertext` and `none` fields of two
```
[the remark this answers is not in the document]
tposavec: @marsella fixed here https://github.com/boltlabs-inc/key-mgmt/pull/540/commits/3b146c796c750d7ad5eab7f9b329fd8c8ee09cfa

lock-keeper/src/crypto/cryptor_key.rs
```diff
@@ -0,0 +1,313 @@
...
+        Ok(Self {
+            key_material: Box::new(key_bytes.into()),
+            config: SensitiveInfoConfig::new(true),
+        })
+    }
+
+    /// Converts [`CryptorKey`] to a byte array.
+    pub fn into_bytes(self) -> [u8; CryptorKey::CRYPTOR_KEY_LENGTH] {
+        (*self.key_material).into()
+    }
```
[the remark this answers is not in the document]
tposavec: @marsella my fault, it is not used for key derivation purposes.  This module will not have/use key derivation functionality.  This function is used only for conversion purposes.

lock-keeper/src/crypto/cryptor.rs
```diff
@@ -0,0 +1,746 @@
...
+            SensitiveInfoConfig::new(true),
+        ))
+    }
+}
+
+/// The context (a.k.a. associated data).
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct CryptorContext {
+    key_server_name: Vec<u8>,
+}
```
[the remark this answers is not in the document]
tposavec: @marsella I agree, the context should be more generic.  I also agree that it should stay as is for this PR.

@marsella @gatoWololo for something like this... do we typically put a `TODO` comment in the code, or do we create a `tech debt` ticket?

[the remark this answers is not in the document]
gatoWololo: @tposavec We usually create an issue. Then we add a
`///TODO: Short Message. See Issue #X.  `

lock-keeper/src/crypto/cryptor.rs
```diff
@@ -0,0 +1,746 @@
...
+    ///
+    /// * `nonce` - A `chacha20poly1305::Nonce` instance. This is used in the
+    ///   encryption process to ensure the security of the ciphertext. Every
+    ///   time data is encrypted, a unique nonce should be generated.
+    ///
+    /// # Returns
+    ///
+    /// Returns a new instance of the [`Decryptor`] type, initialized with the
+    /// provided `ciphertext`, `context`, `nonce`, and `config`.
+    pub fn new(
```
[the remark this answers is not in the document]
tposavec: @marsella I agree, my mistake, didn't mean to make that `pub`.  I changed the constructor to be `private` here https://github.com/boltlabs-inc/key-mgmt/pull/540/commits/9e71801f37a0e748ab104920e09304f55594a3e0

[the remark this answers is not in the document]
marsella: Nice, thanks.

lock-keeper/src/crypto/cryptor_key.rs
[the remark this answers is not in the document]
marsella: Hm, okay. In that case I would recommend adding a warning of some kind to the docs that emphasizes that this is giving direct access to the key bytes and that the caller should be careful to manually zeroize them after use.

lock-keeper/src/crypto.rs
```diff
@@ -31,6 +33,7 @@ mod storage_key;
 
 use crate::rpc::Message;
 pub use arbitrary_secret::Secret;
+pub use cryptor_key::CryptorKey;
```
marsella, line 36: question: Do we need to make any other types publicly visible here? I would think you need to `pub use cryptor::{Encryptor, Decryptor, CryptorContext};` or just make the whole `cryptor` module public.

marsella (commented):
Okay, looks good to me! Will approve once the docs are added for the two remaining open threads.

lock-keeper/src/crypto/cryptor_key.rs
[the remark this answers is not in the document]
tposavec: @marsella fixed here https://github.com/boltlabs-inc/key-mgmt/pull/540/commits/224c79b7c45e1b9de61301b62f9c55ca3f718383

lock-keeper/src/crypto/cryptor.rs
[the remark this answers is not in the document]
tposavec: @marsella fixed here https://github.com/boltlabs-inc/key-mgmt/pull/540/commits/9c9bc97c963a9d353cba790249401ae48719f2bd

lock-keeper/src/crypto.rs
tposavec: @marsella fixed here https://github.com/boltlabs-inc/key-mgmt/pull/540/commits/7e200264ee515893f4bfa4e51c5acd1d79d2c758

lock-keeper/src/crypto/cryptor_key.rs
```diff
@@ -145,6 +145,29 @@ impl CryptorKey {
...
+    ///
+    /// # Warning
+    ///
+    /// This method gives direct access to the key material bytes.
+    /// The caller should be careful to manually zeroize them after use to
+    /// prevent unintended exposure of sensitive information. Consider using
+    /// the `zeroize` crate to securely zero the data.
+    ///
+    /// # Example
+    ///
```
marsella, line 157: nice example!

marsella (approved):
Looks good, thanks!

PR #545, opened 2023-08-14
Create new constructor for `ShardedSigningKeyPair`.

### LINKED ISSUE #544: Modify Constructor for ShardedSigningKeyPair
The existing `new` constructor for `ShardedSigningKeyPair` does not guarantee that public key corresponded to shards of the private key. It is up to the user to get this right.

Solution: Make a new constructor which both shards and encrypts.

gatoWololo:
Closes #544 

- New constructor `create_and_encrypt`.
- Removes previous constructor which did not guarantee that public key corresponded to shards of private key.

### DIFF
--- a/lock-keeper/src/crypto/sharding.rs
+++ b/lock-keeper/src/crypto/sharding.rs
@@ -168,11 +168,14 @@ pub struct ShardedSigningKeyPair {
 }
 
 impl ShardedSigningKeyPair {
-    pub fn new(public_key: SigningPublicKey, encrypted_shards: Vec<EncryptedShard>) -> Self {
-        Self {
-            public_key,
-            encrypted_shards,
-        }
+    pub fn create_and_encrypt(seal_key: &SealKey) -> Result<ShardedSigningKeyPair, CryptoError> {
+        let key_pair = SigningPrivateKey::generate(&mut OsRng);
+
+        let key_data = ShardedSigningKeyPair {
+            public_key: key_pair.public_key(),
+            encrypted_shards: key_pair.shard_key_and_encrypt(seal_key)?,
+        };
+        Ok(key_data)
     }
 
     pub fn public_key(&self) -> &SigningPublicKey {


lock-keeper/src/crypto/sharding.rs
```diff
@@ -168,11 +168,14 @@ pub struct ShardedSigningKeyPair {
 }
 
 impl ShardedSigningKeyPair {
-    pub fn new(public_key: SigningPublicKey, encrypted_shards: Vec<EncryptedShard>) -> Self {
-        Self {
-            public_key,
-            encrypted_shards,
-        }
+    pub fn create_and_encrypt(seal_key: &SealKey) -> Result<ShardedSigningKeyPair, CryptoError> {
+        let key_pair = SigningPrivateKey::generate(&mut OsRng);
```
marsella, line 172: I would recommend taking the RNG as a parameter (`impl CryptoRng + RngCore`) so that the calling application can make a consistent decision about their RNG usage instead of hardcoding this here.

marsella (changes requested):
Looks lovely, just one comment.

lock-keeper/src/crypto/sharding.rs
```diff
@@ -168,11 +168,14 @@ pub struct ShardedSigningKeyPair {
 }
 
 impl ShardedSigningKeyPair {
-    pub fn new(public_key: SigningPublicKey, encrypted_shards: Vec<EncryptedShard>) -> Self {
-        Self {
-            public_key,
-            encrypted_shards,
-        }
+    pub fn create_and_encrypt(seal_key: &SealKey) -> Result<ShardedSigningKeyPair, CryptoError> {
```
marsella, line 171: Also, consider writing docs for this method?

gatoWololo: Done!

lock-keeper/src/crypto/sharding.rs
gatoWololo: That makes a lot sense :brain:

marsella (approved):
Looks great, thank you!

PR #550, opened 2023-08-16
549 add serialization trait for decryptor type

### LINKED ISSUE #549: Add the serialization trait for the Decryptor type
We need to add serialization and de-serialization traits for the Decryptor type.

tposavec:
See #549

### DIFF
--- a/lock-keeper-client/src/client.rs
+++ b/lock-keeper-client/src/client.rs
@@ -41,10 +41,9 @@ impl FromStr for Password {
 }
 
 impl Password {
-    // tptodo:
-    // pub fn new(bytes: &[u8]) -> Self {
-    //     Password(bytes.to_vec())
-    // }
+    pub fn new(bytes: &[u8]) -> Self {
+        Password(bytes.to_vec())
+    }
 
     pub fn as_bytes(&self) -> &[u8] {
         &self.0[..]
--- a/lock-keeper-client/src/client.rs
+++ b/lock-keeper-client/src/client.rs
@@ -41,8 +41,8 @@ impl FromStr for Password {
 }
 
 impl Password {
-    pub fn new(bytes: &[u8]) -> Self {
-        Password(bytes.to_vec())
+    pub fn new(bytes: Vec<u8>) -> Self {
+        Password(bytes)
     }
 
     pub fn as_bytes(&self) -> &[u8] {
--- a/lock-keeper-client/src/client.rs
+++ b/lock-keeper-client/src/client.rs
@@ -41,6 +41,11 @@ impl FromStr for Password {
 }
 
 impl Password {
+    // tptodo:
+    // pub fn new(bytes: &[u8]) -> Self {
+    //     Password(bytes.to_vec())
+    // }
+
     pub fn as_bytes(&self) -> &[u8] {
         &self.0[..]
     }
--- a/lock-keeper/src/crypto/cryptor.rs
+++ b/lock-keeper/src/crypto/cryptor.rs
@@ -228,6 +228,28 @@ impl Encryptor {
         }
     }
 
+    /// Returns a reference to the encrypted data bytes stored within the
+    /// [`Encryptor`] instance.
+    ///
+    /// This method provides read-only access to the internal `data` field,
+    /// ensuring that the encrypted data cannot be modified directly by the
+    /// caller.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// // Assuming you've set up an `Encryptor` instance named `encryptor`
+    /// let data_bytes: &[u8] = encryptor.data();
+    /// ```
+    ///
+    /// # Returns
+    ///
+    /// A byte slice (`&[u8]`) that represents the encrypted data.    
+    /// Returns a reference to the data bytes.
+    pub fn data(&self) -> &[u8] {
+        &self.data
+    }
+
     /// Encrypts data using the provided encryption key.
     ///
     /// # Arguments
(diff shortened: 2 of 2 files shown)


lock-keeper/src/crypto/cryptor.rs
```diff
@@ -429,6 +451,113 @@ impl std::fmt::Debug for Decryptor {
...
+
+    fn try_from(decryptor: Decryptor) -> Result<Self, Self::Error> {
+        // Output byte array format:
+        // ciphertext len (2 bytes) || ciphertext || context len (2 bytes) ||
+        // context data || nonce len (2 bytes) || nonce data || config len (2 bytes) ||
+        // config data
+
+        let context: Vec<u8> = decryptor.context.to_owned().into();
+        let config: Vec<u8> = decryptor.config.to_owned().try_into()?;
+        let nonce: Vec<u8> = decryptor.nonce.to_vec();
```
marsella, line 467: todo: Rather than cloning each field, I would destruct (?) this like so:
```rust
let Decryptor {
    ciphertext,
    nonce,
    config,
    context
} = decryptor;
```

Then you can directly access e.g. `ciphertext` and manually convert each field to bytes using the `into` / `try_into` / `to_vec` fields.

marsella: Also, the nonce doesn't need to be converted `to_vec`. You should be able to use the generic array in the same way as the vecs.

lock-keeper/src/crypto/cryptor.rs
```diff
@@ -429,6 +451,113 @@ impl std::fmt::Debug for Decryptor {
...
+        // because a usize is always at least as large as a u16.
+        let ciphertext = parse.take_bytes(ciphertext_length as usize)?.to_vec();
+
+        let context_length = parse.take_bytes_as_u16()?;
+        let context: Vec<u8> = parse.take_bytes(context_length as usize)?.to_vec();
+
+        // make sure that the length of the context matches the parsed length
+        if context.len() != context_length as usize {
+            return Err(CryptoError::ConversionError);
+        }
```
marsella, line 524: note: I think you don't need this check; `take_bytes` will fail if it doesn't get exactly `context_length` bytes. In other places, we use a manual length check when we use the `ParseBytes::take_rest` method, which doesn't take a length as parameter.

lock-keeper/src/crypto/cryptor.rs
```diff
@@ -429,6 +451,113 @@ impl std::fmt::Debug for Decryptor {
...
+
+        // create a default instance of Nonce type, so we can get the expected length
+        let nonce_length_default = chacha20poly1305::Nonce::default().len();
+        // need to clone nonce from the slice that was previously serialized
+        let mut nonce_new = chacha20poly1305::Nonce::default();
+        if nonce_length == nonce_length_default as u16 {
+            nonce_new.clone_from_slice(&nonce);
+        } else {
+            return Err(CryptoError::ConversionError);
+        }
```
marsella, line 537: suggestion: This reads as kind of confusing to me. I would probably hardcode the nonce length (since we've already hardcoded the chacha parameters everywhere by dint of using `chacha20poly1305` explicitly) like:
```rust
let nonce_length = parse.take_bytes_as_u16()?;
let nonce: [u8; 12] = parse.take_bytes(nonce_length as usize)?.try_into()
            .map_err(|_| CryptoError::ConversionError)?;
```
Then just do `nonce.into()` when you create `self` down at the bottom to turn it back to a generic array.

lock-keeper/src/crypto/cryptor.rs
```diff
@@ -429,6 +451,113 @@ impl std::fmt::Debug for Decryptor {
...
+
+        // make sure that the length of the nonce matches the parsed length
+        if nonce.len() != nonce_length as usize {
+            return Err(CryptoError::ConversionError);
+        }
```
marsella, line 542: todo: Again, don't need this length check because it is handled by `take_bytes`.

gatoWololo (commented):
Looks good to me. I will defer to Marcella for the crypto details.

lock-keeper/src/crypto/cryptor.rs
```diff
@@ -607,6 +736,64 @@ mod test {
...
 
+    /// Converts a decryptor between [`Decryptor`] and `Vec<u8>`
+    /// representations and verifies the conversion.
+    ///
+    /// # Errors
+    ///
+    /// Returns a [`CryptoError`] if there is an error during encryption or
+    /// conversion.
+    #[test]
+    fn decryptor_vec_u8_conversion() -> Result<(), CryptoError> {
```
marsella, line 747: suggestion: This is a good happy test. You might consider adding sad tests that break each field / length to ensure that we need every field to be encoded, and that all the lengths need to be correct. Will leave that up to you.

marsella (changes requested):
Thanks! I have a few comments about ways to simplify and reduce cloning throughout.

Are the changes to the `Password` type supposed to be here? They don't seem related to the issue description.

tposavec:
> Thanks! I have a few comments about ways to simplify and reduce cloning throughout.
> 
> Are the changes to the `Password` type supposed to be here? They don't seem related to the issue description.

@marsella sorry, it seemed like a minor change, but you are correct it is not related to the issue description.  Would you prefer I submit a separate PR for the password related change?

lock-keeper/src/crypto/cryptor.rs
tposavec: @marsella fixed here https://github.com/boltlabs-inc/key-mgmt/pull/550/commits/ccb79975ea7c41b72783ba8a7cb843a489a183f5

lock-keeper/src/crypto/cryptor.rs
tposavec: @marsella fixed here https://github.com/boltlabs-inc/key-mgmt/pull/550/commits/c5c8e52ce54e3e06ca1d365bb8b9044021ecd77b

lock-keeper/src/crypto/cryptor.rs
tposavec: @marsella fixed here https://github.com/boltlabs-inc/key-mgmt/pull/550/commits/42864625aec4d500f839738867f411a4d821dc04

lock-keeper/src/crypto/cryptor.rs
tposavec: @marsella fixed here https://github.com/boltlabs-inc/key-mgmt/commit/42864625aec4d500f839738867f411a4d821dc04

lock-keeper-client/src/client.rs
```diff
@@ -41,6 +41,10 @@ impl FromStr for Password {
 }
 
 impl Password {
+    pub fn new(bytes: &[u8]) -> Self {
+        Password(bytes.to_vec())
```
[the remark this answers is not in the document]
tposavec: @marsella fixed here https://github.com/boltlabs-inc/key-mgmt/pull/550/commits/a3bdbfcae7265409e67f9c432f9981bb3bd7e609

lock-keeper/src/crypto/cryptor.rs
tposavec: @marsella fixed here https://github.com/boltlabs-inc/key-mgmt/pull/550/commits/589373dc99e5b0980d41a0a27a7909ba575c97fc

lock-keeper/src/crypto/cryptor.rs
```diff
@@ -607,6 +730,110 @@ mod test {
...
+    }
+
+    /// Converts a decryptor from `Vec<u8>` to [`Decryptor`].
+    /// Removes some bytes to make the conversion fail.
+    ///
+    /// # Errors
+    ///
+    /// Returns a [`CryptoError`] if there is an error
+    #[test]
+    fn decryptor_conversion_test_incomplete_fields() -> Result<(), CryptoError> {
```
marsella, line 772: This is an absolute aesthetic nit.

I like to name tests to describe the invariant or property that they're testing. Especially with Rust where `cargo test` spits out all the names onto the terminal, it can be helpful to get little descriptive sentences rather than sort of keyword collections.

So, for the three tests you have here, I would name them something like:
- ` decryptor_from_bytes_requires_all_fields`
- `decryptor_from_bytes_cannot_have_extra_bytes`
- `decryptor_vec_u8_conversion_works`

marsella (approved):
We discussed the `Password` change offline; it's fine to wrap into this PR. I mainly wanted to make sure it was intentional, and to flag that type for further review to make sure it's as misuse-resistant as possible for the uses cases we need it for.

This looks good, otherwise. I raised one concern about memory management but I would say it's non-blocking; if Omar shares my concern then we should probably make a separate issue and review all the `From<Vec<u8>>` instances in this repo to make sure we're zeroizing appropriately.

The other comments are suggestions.

lock-keeper/src/crypto/cryptor.rs
```diff
@@ -429,6 +451,107 @@ impl std::fmt::Debug for Decryptor {
...
+        let ciphertext = parse.take_bytes(ciphertext_length as usize)?.to_vec();
+
+        let context_length = parse.take_bytes_as_u16()?;
+        let context: Vec<u8> = parse.take_bytes(context_length as usize)?.to_vec();
+
+        let nonce_length = parse.take_bytes_as_u16()?;
+        let nonce: [u8; 12] = parse
+            .take_bytes(nonce_length as usize)?
+            .try_into()
+            .map_err(|_| CryptoError::ConversionError)?;
```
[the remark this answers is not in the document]
tposavec: Addressed here https://github.com/boltlabs-inc/key-mgmt/pull/550/commits/773fd51af64fd24fd73a24c698830143d64dbb64

lock-keeper/src/crypto/cryptor.rs
tposavec: Changed here https://github.com/boltlabs-inc/key-mgmt/pull/550/commits/1421e5494357eb9f4a5f59d32d39ec73f7025a5a

lock-keeper/src/crypto/cryptor.rs
```diff
@@ -429,6 +451,107 @@ impl std::fmt::Debug for Decryptor {
...
+            .chain(config_bytes)
+            .collect();
+
+        Ok(bytes)
+    }
+}
+
+/// Implementation of the `TryFrom` trait for converting from `Vec<u8>` to
+/// [`Decryptor`]
+impl TryFrom<Vec<u8>> for Decryptor {
```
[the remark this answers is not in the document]
gatoWololo: > I can't tell from here whether the try_intos zeroize at the end

Good question! `.try_into` is a generic implementation for anything that implements `TryFrom`. From the standard library:
```rust
// TryFrom implies TryInto
#[stable(feature = "try_from", since = "1.34.0")]
impl<T, U> TryInto<U> for T
where
    U: TryFrom<T>,
{
    type Error = U::Error;

    #[inline]
    fn try_into(self) -> Result<U, U::Error> {
        U::try_from(self)
    }
}
```
So as long as our `TryFrom` instances are correctly zeroizing the bytes, then we are good.

> I'm not sure if Rust memory management semantics will copy the bytes as they get passed to try_into so there's another extra copy around

Out of the 3 conversion:
```rust
Ok(Self {
            ciphertext,
            context: context.try_into()?,
            nonce: nonce.try_into()?,
            config: config.try_into()?,
        })
```
Both `context` and `config` are `Vec<u8>` so no copying is happening when calling `.try_into`. The vectors are being passed "by value". `nonce` is `[u8; 12]`, this is probably being copied onto the stack when passed into `.try_into`. I'm not sure if there is much we can do about this, as the value is technically _moved_ by the call, so we cannot zeroize it.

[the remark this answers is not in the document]
marsella: Okay, thanks. Then yeah, let's just leave this as is for now.

PR #555, opened 2023-08-24
Rust compiler version pin fix

gatoWololo:
Properly pin Rust version. Today's update broke key-mgmt :unamused:


gatoWololo:
Ugh. This is wrong. die.

Issue #563, opened 2023-09-14
Crypto Library: Provide encryption scheme for AES-GCM-SIV


LordQ1:
**Tasks:** 

1. Update the crypto library to provide support for AES-GCM-SIV
2. Question: Do we need to provide backwards compatible support for AES-GCM for prior data that was encrypted with this library?
    - If this is easy enough to do, then let's do it
    - If this is complicated (as presumed), then let's document the issues and our choice
3. Provide any necessary unit tests
4. Provide any necessary developer documentation

PR #565, opened 2023-09-18
update encryption primitive #563

marsella:
Updates encryption library from AES-GCM to AES-GCM-SIV.

I tested locally and did some integration testing and it seems to work fine.

See #563


lock-keeper/src/crypto/sharding.rs
```diff
@@ -33,16 +34,20 @@ impl SealKey {
...
-    }
-
     pub fn material(&self) -> &[u8] {
         &self.material
     }
 }
 
+impl AsRef<Key<Aes256GcmSiv>> for SealKey {
+    fn as_ref(&self) -> &Key<Aes256GcmSiv> {
+        // I think this won't clone the bytes; the `From` impl between slices and
```
gatoWololo, line 44: This is a great change! Thanks. This is a good pattern we could apply elsewhere perhaps.

gatoWololo (approved):
Nice short PR. Love it!

PR #570, opened 2023-10-02
Merge `dev` changes to `main`

gatoWololo:
Version 1.12

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1272,7 +1272,7 @@ dependencies = [
 
 [[package]]
 name = "key-server-cli"
-version = "1.8.0"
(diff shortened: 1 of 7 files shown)


jwantz (approved):
I verified the following before approving the PR:

- There are 3 commits and 7 files changed in this PR, which matches the 3 successful commits in the `develop` branch since Aug 25, which was after the v1.8 tag was applied.
- The 3 commit hashes after Aug 25 match between `develop` and this PR.
- The version number in the cargo workspace has been updated to "1.12.0".
Note: We are not yet capturing version numbers in the SQL DB for key-mgmt.

PR #575, opened 2023-10-16
Fix store_data_blob_identity retrieved flag

picklenerd:
This PR fixes a bug in the tests that causes occasional failures in CI.

This test case was calling the wrong `get` function and checking the `retrieved` flag which is not used for blobs.


gatoWololo (approved):
:pray:

Issue #587, opened 2023-11-15
Add min_connections to DB config


jrdngr:
We've seen big performance gains from enabling this feature in the DB but the code that adds this is in a testing branch. Implement the config change properly.

PR #588, opened 2023-11-15
#587 profiling and performance

jrdngr:
#587 

This PR implements some of the profiling and performance improvements that we've worked on in temp branches.


jrdngr:
Breaking this up into two PRs

PR #594, opened 2023-11-24
Add additional documentation around channels and secure storage.

gatoWololo:
Additional documentation for our code!


lock-keeper-key-server/src/server/channel.rs
```diff
@@ -71,6 +77,61 @@ pub struct Authenticated<RNG: CryptoRng + RngCore> {
...
+                .map_err(LockKeeperError::Crypto)?
+        }
+        .try_into_message()?;
+
+        let payload = Ok(encrypted_message);
+
+        Ok(self.sender.send(payload).await?)
+    }
+}
+
```
gatoWololo, line 134: I just moved this "up" in the file so the impl would be right next to its struct definition :tipping_hand_man:

lock-keeper-key-server/src/channel.md
```diff
@@ -0,0 +1,39 @@
...
+sent over `Channel`s.
+
+## Setting up Channels in Client and Server
+Some ceremony is required to set up these channels. `Channel`s are implemented in terms of two unidirectional channels:
+- Channel 1: Server holds receiving end and client holds sending end.
+- Channel 2: Server holds sending end and client holds receiving end. 
+
+Channel 1 is instantiated by the client when making the initial gRPC call. The receiver end is kept by the client, the
+sending end is sent along with the initial `Request`. When the server receives this request, it saves the receiver end
+that the client sent. The server then creates a new channel: The server keeps the sender end and is expected to send the
```
krs85, line 29: This is kinda confusing... you say the _receiver_ end is kept by the client, and the _sending_ end is sent to the server. But then immediately after you say the server saves the _receiver_ end that the client sent. Do you mean the _sender_ end?

gatoWololo: You are totally right. This is why I wanted a review. Fixed!

krs85 (approved):
Looks great to me. Thanks for updating the documentation, it  is very obvious to me when reading this what the code is trying to accomplish :-)

PR #599, opened 2023-11-27
Merge CI changes into develop

jrdngr:


jrdngr:
Ok git doesn't like this without develop being up to date. I'll make a separate PR cherry picking this.

Issue #605, opened 2023-11-28
Reduce CI Times


gatoWololo:
CI times for the Lock-Keeper repository seem reasonable but we should see if there is some low-hanging fruit we can address for better times.

Task: Reduce CI execution.

jrdngr:
At the moment we're running tests 3 separate times in series for the three different types of environments we have set up. These environments were set up based on some early designs that are no longer relevant. We can get a significant speedup by simply disabling the tests for two of them.

jrdngr:
Results are in, removing the extra 3 environments from testing saves us 1 minute and removing them from Docker saves us an addition 10 seconds!

This brings our total CI time down from around 12 minutes to around 11 minutes!

LordQ1:
Per Omar: Nothing more to do here; we have other Ci improving tickets.

PR #606, opened 2023-11-28
Disable test environments that we don't use #605

jrdngr:
This PR disables integration tests for the environments we don't use. This should make CI testing times around 1/3 of what they were before.

#605


gatoWololo (approved):
:eye:

jrdngr:
This saves a whopping one minute out of the 12 minutes it was taking before. Might as well just keep the extra tests 😆

PR #614, opened 2023-12-13
EncryptedShard derives ZeroizeOnDrop

krs85:
Just adds the `ZeroizeOnDrop` struct to the `EncryptedShard` struct.


gatoWololo (approved):
Approved

PR #616, opened 2023-12-15
Merge `qa` branch into `main`

gatoWololo:
Merge LK 2.3.0 changes into main branch.


jwantz (approved):
This merge contains just 4 commits and 18 file changes to match with the 2.3 release. I also checked the following:

- The 4 commits between Nov 15- 27 match with those in the `qa` branch
- One interesting anomaly is that it appears that 2 extra commits exist in the `qa` branch for a total of 6 (instead of 4)
- This happened during the 2.3 development timeline, as we were introducing the new `qa` branch
- We found that we needed to sync up `qa` and `main` at that time, which accounted for what appears to be 2 extra commits

PR #619, opened 2023-12-28
add SecureBuffer and SecureString

krs85:
This PR adds the `secure_structs` module, which contains `SecureBuffer` and `SecureString`. Both are wrappers around primitive types that help us ensure we zeroize the sensitive data the structs contain.


gatoWololo (approved):
Looks good to me!

Issue #627, opened 2024-01-26
Add functionality to seal a signing private key


krs85:
We need new Rust types and functions that allow us to seal a signing private key.

PR #630, opened 2024-02-05
Add functionality to seal a signing private key #627

krs85:
This PR implements #627.

### DIFF
--- a/lock-keeper/Cargo.toml
+++ b/lock-keeper/Cargo.toml
@@ -9,6 +9,7 @@ edition = "2021"
 # Workspace dependencies
 argon2.workspace = true
 async-trait.workspace = true
+base64.workspace = true
 bincode.workspace = true
 bytes.workspace = true
 futures.workspace = true
@@ -34,7 +35,7 @@ zeroize.workspace = true
 
 
 # Other dependencies
-aes-gcm-siv = "0.11" # Used for encryption/decryption of sharded keys.
+aes-gcm-siv = "0.11"                                                # Used for encryption/decryption of sharded keys.
 anyhow = "1.0"
 chacha20poly1305 = "0.10"
 hkdf = "0.12"
--- a/lock-keeper/src/crypto/seal_signing_private_key.rs
+++ b/lock-keeper/src/crypto/seal_signing_private_key.rs
@@ -1,0 +1,52 @@
+use aes_gcm_siv::{aead::Aead, AeadCore, Aes256GcmSiv, KeyInit, Nonce};
+use base64::{engine::general_purpose, Engine};
+use rand::rngs::OsRng;
+use serde::{Deserialize, Serialize};
+use zeroize::Zeroize;
+
+use super::{sharding::SealKey, CryptoError, SigningPrivateKey};
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct EncryptedAttestationPrivateKey {
+    encrypted: Vec<u8>,
+    nonce: Nonce,
+}
+
+impl EncryptedAttestationPrivateKey {
+    pub fn seal(
+        seal_key: &SealKey,
+        attestation_private_key: &SigningPrivateKey,
+    ) -> Result<Self, CryptoError> {
+        let nonce: Nonce = Aes256GcmSiv::generate_nonce(&mut OsRng);
+        let cipher = Aes256GcmSiv::new(seal_key.as_ref());
+        let encrypted = cipher
+            .encrypt(&nonce, attestation_private_key.as_bytes().as_slice())
+            .map_err(|_err| CryptoError::EncryptionFailed)?;
+        Ok(Self { encrypted, nonce })
+    }
+
+    pub fn decrypt(self, seal_key: &SealKey) -> Result<SigningPrivateKey, CryptoError> {
+        let cipher = Aes256GcmSiv::new(seal_key.as_ref());
+        let mut decrypted = cipher
+            .decrypt(&self.nonce, self.encrypted.as_slice())
+            .map_err(|_err| CryptoError::DecryptionFailed)?;
+        let signing_private_key = SigningPrivateKey::from_bytes(&decrypted)
+            .map_err(|_err| CryptoError::ConversionError)?;
+        decrypted.zeroize();
+        Ok(signing_private_key)
+    }
+
+    pub fn into_base64(self) -> Result<String, CryptoError> {
+        let serialized = bincode::serialize(&self)?;
+        Ok(general_purpose::STANDARD.encode(serialized))
+    }
+
+    pub fn from_base64(attestation_key_string: &str) -> Result<Self, CryptoError> {
+        let decoded: Vec<u8> = general_purpose::STANDARD.decode(attestation_key_string)?;
+        Ok(bincode::deserialize(&decoded)?)
+    }
+
+    pub fn set_nonce(&mut self, new_nonce: Nonce) {
+        self.nonce = new_nonce;
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1336,6 +1336,7 @@ dependencies = [
  "anyhow",
  "argon2",
  "async-trait",
+ "base64 0.21.2",
  "bincode",
  "bytes",
  "chacha20poly1305",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ version = "2.5.0"
 [workspace.dependencies]
 argon2 = "0.4"
 async-trait = "0.1"
+base64 = "0.21.0"
 bincode = "1"
 bytes = { version = "1", features = ["serde"] }
 clap = { version = "3", features = ["derive", "env"] }
--- a/lock-keeper/src/crypto.rs
+++ b/lock-keeper/src/crypto.rs
@@ -26,6 +26,7 @@ mod cryptor;
 mod cryptor_key;
 mod data_blob;
 mod generic;
+pub mod seal_signing_private_key;
 pub mod secure_structs;
 pub mod sharding;
 mod signing_key;
--- a/lock-keeper/src/crypto/cryptor.rs
+++ b/lock-keeper/src/crypto/cryptor.rs
@@ -161,7 +161,7 @@ impl TryFrom<Vec<u8>> for Encryptor {
 
         Ok(Self {
             data,
-            context: context.try_into()?,
+            context: context.into(),
             config: config.try_into()?,
         })
     }
@@ -546,8 +546,8 @@ impl TryFrom<Vec<u8>> for Decryptor {
 
         Ok(Self {
             ciphertext,
-            context: context.try_into()?,
-            nonce: nonce.try_into()?,
+            context: context.into(),
+            nonce: nonce.into(),
             config: config.try_into()?,
         })
     }
--- a/lock-keeper/src/crypto/generic.rs
+++ b/lock-keeper/src/crypto/generic.rs
@@ -28,6 +28,10 @@ pub enum CryptoError {
     DecryptionFailed,
     #[error("Encryption failed")]
     EncryptionFailed,
+    #[error(transparent)]
+    FromBase64(#[from] base64::DecodeError),
+    #[error(transparent)]
+    FromBincode(#[from] bincode::Error),
     #[error("Invalid encryption key")]
     InvalidEncryptionKey,
     #[error("Sensitive info check failed")]
--- a/lock-keeper/src/infrastructure/sensitive_info.rs
+++ b/lock-keeper/src/infrastructure/sensitive_info.rs
@@ -45,7 +45,7 @@ impl TryFrom<SensitiveInfoConfig> for Vec<u8> {
 impl SensitiveInfoConfig {
     /// Tis is the label used to redact sensitive information in the library
     /// output
-    const REDACTED_INFO_LABEL: &str = "***REDACTED***";
+    const REDACTED_INFO_LABEL: &'static str = "***REDACTED***";
 
     /// Constructor for [`SensitiveInfoConfig`] type.
     /// The default is to redact sensitive information in the output.


lock-keeper/src/crypto/seal_signing_private_key.rs
```diff
@@ -0,0 +1,74 @@
...
+use super::{sharding::SealKey, CryptoError, SigningPrivateKey};
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct EncryptedAttestationPrivateKey {
+    encrypted: Vec<u8>,
+    nonce: GenericArray<u8, U12>,
+}
+
+impl EncryptedAttestationPrivateKey {
+    pub fn seal_attestation_private_key(
```
krs85, line 18: The original function was `new()` (a constructor) that was called from a wrapper function called `seal_attestation_key()`. This function does more than construct an `EncryptedAttestationPrivateKey`, but I also found it confusing to call its constructor from another function. So I removed `seal_attestation_key()` and renamed this function.

lock-keeper/src/crypto/seal_signing_private_key.rs
```diff
@@ -0,0 +1,74 @@
+use aes_gcm_siv::{
```
krs85, line 1: I wasn't totally sure _where_ to put `EncrpytedAttestationPrivateKey` and `UnsealedAttestationKey`. After going through the lock keeper code, I couldn't find an existing file that felt like the right place, but I do still feel that they belong under `crypto`. Making this comment because I am open to suggestions if anyone thinks there's a better spot for them!

lock-keeper/Cargo.toml
```diff
@@ -34,7 +35,7 @@ zeroize.workspace = true
 
 
 # Other dependencies
-aes-gcm-siv = "0.11" # Used for encryption/decryption of sharded keys.
+aes-gcm-siv = "0.11"                                                # Used for encryption/decryption of sharded keys.
```
krs85, line 38: I don't know why my IDE keeps formatting this this way. 🥲

lock-keeper/src/crypto/seal_signing_private_key.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+    pub fn from_string(attestation_key_string: String) -> Result<Self, CryptoError> {
+        let decoded: Vec<u8> = general_purpose::STANDARD.decode(attestation_key_string)?;
+        Ok(bincode::deserialize(&decoded)?)
+    }
+
+    pub fn set_nonce(&mut self, new_nonce: Nonce) {
+        self.nonce = new_nonce;
+    }
+}
```
krs85, line 53: I removed the original `serialize()` and `deserialize()` methods for `EncryptedAttestationPrivateKey`, because they were just wrappers around `bincode::serialize()` and `bincode::deserialize()`.

lock-keeper/src/crypto/seal_signing_private_key.rs
```diff
@@ -0,0 +1,74 @@
...
+        Ok(general_purpose::STANDARD.encode(serialized))
+    }
+    pub fn from_string(attestation_key_string: String) -> Result<Self, CryptoError> {
+        let decoded: Vec<u8> = general_purpose::STANDARD.decode(attestation_key_string)?;
+        Ok(bincode::deserialize(&decoded)?)
+    }
+
+    pub fn set_nonce(&mut self, new_nonce: Nonce) {
+        self.nonce = new_nonce;
+    }
```
krs85, line 52: I _could_  make the `nonce` a public field, but the other field isn't public, and it just doesn't seem worth it to me.

lock-keeper/src/crypto/seal_signing_private_key.rs
```diff
@@ -0,0 +1,74 @@
...
+
+use super::{sharding::SealKey, CryptoError, SigningPrivateKey};
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct EncryptedAttestationPrivateKey {
+    encrypted: Vec<u8>,
+    nonce: GenericArray<u8, U12>,
+}
+
+impl EncryptedAttestationPrivateKey {
```
krs85, line 17: I removed the original `serialize()` and `deserialize()` methods for `EncryptedAttestationPrivateKey` because they were just wrappers around `bincode::serialize()` and `bincode::deserialize()`.

lock-keeper/src/crypto/seal_signing_private_key.rs
```diff
@@ -0,0 +1,74 @@
...
+use base64::{engine::general_purpose, Engine};
+use rand::rngs::OsRng;
+use serde::{Deserialize, Serialize};
+
+use super::{sharding::SealKey, CryptoError, SigningPrivateKey};
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct EncryptedAttestationPrivateKey {
+    encrypted: Vec<u8>,
+    nonce: GenericArray<u8, U12>,
```
gatoWololo, line 14: We should use the type `Nonce` here, it can be imported as: `use aes_gcm_siv::Nonce`.

lock-keeper/src/crypto/seal_signing_private_key.rs
gatoWololo: I like this change could we rename this function to just `seal` since it is an associated function we have to call it as `EncryptedAttestationPrivateKey::seal(...)`. I think this cuts down on some verbosity.

lock-keeper/src/crypto/seal_signing_private_key.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+impl EncryptedAttestationPrivateKey {
+    pub fn seal_attestation_private_key(
+        seal_key: &SealKey,
+        attestation_private_key: &SigningPrivateKey,
+    ) -> Result<Self, CryptoError> {
+        let nonce: Nonce = Aes256GcmSiv::generate_nonce(&mut OsRng);
+        let cipher = Aes256GcmSiv::new(seal_key.as_ref());
+        let attestation_private_key_sealed = cipher
```
gatoWololo, line 24: Minor nit: rename `attestation_private_key_sealed` to just `sealed` or `encrypted`.

lock-keeper/src/crypto/seal_signing_private_key.rs
```diff
@@ -0,0 +1,74 @@
...
+            .encrypt(&nonce, attestation_private_key.as_bytes().as_slice())
+            .map_err(|_err| CryptoError::EncryptionFailed)?;
+        Ok(Self {
+            encrypted: attestation_private_key_sealed,
+            nonce,
+        })
+    }
+
+    pub fn decrypt(self, seal_key: &SealKey) -> Result<SigningPrivateKey, CryptoError> {
+        let nonce: GenericArray<u8, U12> = GenericArray::clone_from_slice(&self.nonce);
```
gatoWololo, line 34: Use `Nonce` type here instead of `GenericArray`

lock-keeper/src/crypto/seal_signing_private_key.rs
```diff
@@ -0,0 +1,74 @@
...
+        let cipher = Aes256GcmSiv::new(seal_key.as_ref());
+        let decrypted = cipher
+            .decrypt(&nonce, self.encrypted.as_slice())
+            .map_err(|_err| CryptoError::DecryptionFailed)?;
+        SigningPrivateKey::from_bytes(&decrypted).map_err(|_err| CryptoError::ConversionError)
+    }
+    pub fn into_string(self) -> Result<String, CryptoError> {
```
gatoWololo, line 41: Add newline here.

lock-keeper/src/crypto/seal_signing_private_key.rs
```diff
@@ -0,0 +1,74 @@
...
+        let serialized = bincode::serialize(&self)?;
+        Ok(general_purpose::STANDARD.encode(serialized))
+    }
+    pub fn from_string(attestation_key_string: String) -> Result<Self, CryptoError> {
```
gatoWololo, line 45: Add newline here.

lock-keeper/src/crypto/seal_signing_private_key.rs
```diff
@@ -0,0 +1,74 @@
...
+        let decoded: Vec<u8> = general_purpose::STANDARD.decode(attestation_key_string)?;
+        Ok(bincode::deserialize(&decoded)?)
+    }
+
+    pub fn set_nonce(&mut self, new_nonce: Nonce) {
```
gatoWololo, line 50: Why is this needed? Does the caller need to be able to change the `Nonce` of an already constructed `Self`?

lock-keeper/src/crypto/seal_signing_private_key.rs
gatoWololo: Why is this fn needed? Does the caller need to be able to change the `Nonce` of an already constructed `Self`?

lock-keeper/src/crypto/seal_signing_private_key.rs
gatoWololo, line 41: Would a better name for this be `into_base64`?

lock-keeper/src/crypto/seal_signing_private_key.rs
gatoWololo, line 45: And this could be `from_base64`?

lock-keeper/src/crypto/seal_signing_private_key.rs
```diff
@@ -0,0 +1,74 @@
...
+        let attestation_private_key_sealed = cipher
+            .encrypt(&nonce, attestation_private_key.as_bytes().as_slice())
+            .map_err(|_err| CryptoError::EncryptionFailed)?;
+        Ok(Self {
+            encrypted: attestation_private_key_sealed,
+            nonce,
+        })
+    }
+
+    pub fn decrypt(self, seal_key: &SealKey) -> Result<SigningPrivateKey, CryptoError> {
```
gatoWololo, line 33: We actually don't need the Nonce cloning here:
```rust
let cipher = Aes256GcmSiv::new(seal_key.as_ref());
        let decrypted = cipher
            .decrypt(&self.nonce, self.encrypted.as_slice())
            .map_err(|_err| CryptoError::DecryptionFailed)?;
```
This works :)

lock-keeper/src/crypto/seal_signing_private_key.rs
```diff
@@ -0,0 +1,74 @@
...
+        let nonce: GenericArray<u8, U12> = GenericArray::clone_from_slice(&self.nonce);
+        let cipher = Aes256GcmSiv::new(seal_key.as_ref());
+        let decrypted = cipher
+            .decrypt(&nonce, self.encrypted.as_slice())
+            .map_err(|_err| CryptoError::DecryptionFailed)?;
+        SigningPrivateKey::from_bytes(&decrypted).map_err(|_err| CryptoError::ConversionError)
```
gatoWololo, line 39: Should we call `zeroize` on `decrypted` here? I think so.

lock-keeper/src/crypto/seal_signing_private_key.rs
```diff
@@ -0,0 +1,74 @@
...
+        let decoded: Vec<u8> = general_purpose::STANDARD.decode(attestation_key_string)?;
+        Ok(bincode::deserialize(&decoded)?)
+    }
+
+    pub fn set_nonce(&mut self, new_nonce: Nonce) {
+        self.nonce = new_nonce;
+    }
+}
+
+pub struct UnsealedAttestationKey {
```
gatoWololo, line 55: I will make a follow-up issue to refactor change this type. Based on its usage, it does not seem necessary.

gatoWololo (changes requested):
Looks good! Just some feedback to polish it.

lock-keeper/src/crypto/seal_signing_private_key.rs
gatoWololo: This place seems good to me.

lock-keeper/Cargo.toml
gatoWololo: lol

lock-keeper/src/crypto/seal_signing_private_key.rs
krs85: I love that!

lock-keeper/src/crypto/seal_signing_private_key.rs
```diff
@@ -0,0 +1,65 @@
...
+    sealed_attestation_private_key: &str,
+) -> Result<(SigningPrivateKey, EncryptedAttestationPrivateKey), CryptoError> {
+    let encrypted_attestation_key =
+        EncryptedAttestationPrivateKey::from_base64(sealed_attestation_private_key.to_string())?;
+
+    Ok((
+        encrypted_attestation_key.clone().decrypt(seal_key)?,
+        encrypted_attestation_key,
+    ))
+}
```
krs85, line 65: Just throwing a comment on this because I want to add documentation / comments to everything in this file, and I need a reminder. Just waiting until we finalize some things!

lock-keeper/src/crypto/seal_signing_private_key.rs
krs85: I *think* I did this right but please check my work 😬

lock-keeper/src/crypto/seal_signing_private_key.rs
krs85: Yes, in one spot the caller needs to change the `Nonce` of an existing `Self`. It is only in one place, so I don't know if the function is worth it. On the other hand, neither field in `EncryptedAttestationPrivateKey` is `pub` right now, and I don't know if it's good practice to have one of them `pub` and one not.

lock-keeper/src/crypto/seal_signing_private_key.rs
krs85: I decided to just do the fix because it wasn't too complicated. 😅 Let me know what you think.

lock-keeper/src/crypto/seal_signing_private_key.rs
gatoWololo: Yep!

lock-keeper/src/crypto/seal_signing_private_key.rs
gatoWololo: Okay, I think this is fine then.

gatoWololo: The only usage I found seems to be for a unit test, would it work if we added a `#[cfg(test)]` to this function? IDK if cfg(test) works across dependencies like this.

lock-keeper/src/crypto/seal_signing_private_key.rs
gatoWololo: Can we just get rid of this function all together? This function is just combining two steps:
1) Turn the `sealed_attestation_private_key: &str` into a EncryptedAttestationPrivateKey.
2) And calling `decrypt` on the `EncryptedAttestationPrivateKey`.

Couldn't the calling application just do these two steps in order?

lock-keeper/src/crypto/seal_signing_private_key.rs
```diff
@@ -0,0 +1,65 @@
...
+        decrypted.zeroize();
+        Ok(signing_private_key)
+    }
+
+    pub fn into_base64(self) -> Result<String, CryptoError> {
+        let serialized = bincode::serialize(&self)?;
+        Ok(general_purpose::STANDARD.encode(serialized))
+    }
+
+    pub fn from_base64(attestation_key_string: String) -> Result<Self, CryptoError> {
```
gatoWololo, line 44: This does not need to be of type `String`. `&str` is sufficient.

gatoWololo (changes requested):
Looks pretty good. Some final feedback.

lock-keeper/src/crypto/seal_signing_private_key.rs
krs85: The use cases in the application also almost never need both of these at once so... I have no idea what the original reasoning for this type and function were 🤷‍♀️

lock-keeper/src/crypto/seal_signing_private_key.rs
gatoWololo: Actually, I'm pretty sure my suggestion won't work. good enough IMO

Issue #642, opened 2024-03-19
Provide support for boltlabs-release.toml file


vidyaap:
Provide support for boltlabs-release.toml file

We wish to have the the status and any api that returns the software version, return the version documented above. This information should also be logged to the application's main log file when the maestro/key-mgmt application starts up.

1. Create a new file in the key-mgmt repository _boltlabs-release.toml_
    ```toml
    [release]
    keymgmt=1.2.3+075.f8ada50.GA
    build-date=YYYYMMDDHHMM+TZ
    ```
2. The default values would be _develop_
3. At build creation time the build process will update the toml file with the appropriate values.

## API
1. Add the information from this toml file to the check_session endpoint in key-mgmt

PR #643, opened 2024-03-20
Add release number to check_session #642

vidyaap:
#642 

I removed the `[release]` heading from the example toml because otherwise I think I would have had to make an extra level of nesting in the Rust type for it. I also made the hyphen in "build-date" an underscore instead.

### DIFF
--- a/boltlabs-release.toml
+++ b/boltlabs-release.toml
@@ -1,0 +1,2 @@
+keymgmt="develop"
+build-date="develop"
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -29,7 +29,7 @@ jobs:
       # Install Rust toolchain
       # action repo is here https://github.com/dtolnay/rust-toolchain
       - name: Install Rust toolchain stable
-        uses: dtolnay/rust-toolchain@1.70.0
+        uses: dtolnay/rust-toolchain@1.73.0
         with:
           # Do NOT use `toolchain` input. It will cause the cargo version specified
           # above to be ignored!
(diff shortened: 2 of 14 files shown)


lock-keeper-key-server/src/lib.rs
```diff
@@ -11,3 +11,5 @@ pub mod server;
 
 pub use config::Config;
 pub use error::LockKeeperServerError;
+
+pub const RELEASE_TOML_PATH: &str = "./boltlabs-release.toml";
```
jrdngr, line 15: What do you think about making this a config value instead? That might give us a little more flexibility with how we set up the file system on the servers.

boltlabs-release.toml
```diff
@@ -0,0 +1,2 @@
+keymgmt="2.6.0+001.54b29c6.DV"
+build-date="202403200408+TZ"
```
jrdngr, line 2: Are these supposed to be "develop" by default?

vidyaap: Oh I wasn't sure how literal that sentence was - i.e. whether they meant literally the string "develop" or just the information from develop filled out here. It does look like the devops script will completely overwrite what's here anyway so I think it can be whatever.

jrdngr: I think the ideas was that if it's "develop" or some other abnormal word, then we know the devops script didn't run correctly.

vidyaap: Oh that makes sense!

jrdngr (approved):

vidyaap:
> Don't you need something in the Dockerfile(s) to copy the boltlabs-release.toml into place on the container?

@opike-boltlabs Just added! Sorry, thought that portion was just for the maestro version of the ticket.

